Issue #1, opened 2022-10-03
Boostrap the Project


halprin:
We're starting a new project.  We want to create a good base to start from.

# Tasks

- [x] Gradle Java project.
- [x] Basic "Hello World" ReST endpoint.
- [x] Check existing markdown files to ensure they're correct.
- [x] Renovate to keep dependencies up-to-date.
- [x] Give access to this repository to to Renovatebot
- [x] Create CI GitHub Action pipeline.
  - [x] Unit tests.
  - [x] Code format verification.
- [x] Set-up Snyk to do SCA stuff
- [x] Set-up CodeQL security scanning.
- [x] Precommit hooks
- [x] Github templates - @halprin
  - [x] story
  - [x] devex/opex
  - [x] design/research
  - [x] bug report

PR #34, opened 2022-10-12
1: GitHub Templates

halprin:
Updated the GitHub issue templates we have.

The GitHub PR templates weren't working because one would need to specify a special URL query parameter for them to work.  I decided to reduce this down to one default PR template so that it would "just work" when creating PRs through the normal means.

Associated with #1.


JohnNKing (approved):
These all look great to me. I love the story template.

Issue #13, opened 2022-10-08
Submit Generic Order data for an individual order


JohnNKing:
# Story

## Sentence

As a Hospital,
In order to process newborn screening lab tests,
I need to be able to submit an individual order.

## Preconditions
- One fake healthcare provider supported

## Acceptance Criteria

- [ ] Healthcare providers can request/transmit an order (a REST API is sufficient; no UI required)
- [ ] The order must include: patient identifier & test type
- [ ] The order data and the current system date must be stored for near-term use (e.g. minutes, not days)
- [ ] The submitter should receive an appropriate acknowledgment if the order was stored successfully (e.g. HTTP 200)
- [ ] Orders must be transmitted securely (e.g. via TLS), except in local development environments

## Tasks

- [x] ADRs.
  - [x] Java
  - [x] Gradle - @halprin 
  - [x] Groovy and Spock for testing - @halprin 
  - [x] OeSA - @halprin 
- [ ] Linting (Sonar?).
- [ ] DAST scanning
- [x] End-to-end tests - @halprin
- [ ] CD to at least a Flexion Azure environment
  - [ ] Terraform
  - [ ] Blue/green strategy
- [ ] API documentation (OpenAPI)
- [x] OeSA application context - @halprin @jorg3lopez 
- [ ] OeSA dependency injection - @halprin @jorg3lopez @tjohnson7021 
- [x] Code coverage check for at least 90%. - @halprin 
- [x] Add domain plugin architecture - @halprin
- [x] Add domain endpoint
- [ ] SBOM generation
- [x] Learn more about FHIR - @halprin
- [ ] Logging framework - @jorg3lopez 

## Additional Context

Add any other context or screenshots about the story here.

### Engineering
- [ ] Threat model updated
- [ ] Code refactored for clarity and no design/technical debt
- [ ] Adhere to separation of concerns; code is not tightly coupled, especially to 3rd party dependencies.
- [ ] Source code is merged to the main branch.
- [ ] Unit test coverage of our code >= 90%
- [ ] Code is reviewed or developed by pair; 1 approval is needed but consider requiring an outside-the-pair reviewer.
- [ ] Build process updated
- [ ] Feature toggles created and/or deleted.  Document the feature toggle.
- [ ] Source code documentation created when the code is not self-documenting.
- [ ] API documentation generated
- [ ] Code quality checks passed
- [ ] Security & privacy gates passed
- [ ] Load tests passed
[...]

PR #45, opened 2022-10-19
OESA Foundation - Create Application Context

jorg3lopez:
This PR adds the Application Context class that behaves like a factory and it will be used for injecting implementations via annotations. 

#### Changes
- Added ApplicationContext.java class
- Added ApplicationContext unit test (ApplicationContextTest)


app/src/test/groovy/gov/hhs/cdc/trustedintermediary/ApplicationContextTest.groovy
```diff
@@ -0,0 +1,17 @@
...
+        setup:
+        def contextA = ApplicationContext.getInstance()
+        def contextB = ApplicationContext.getInstance()
+
+        when:
+        def result = contextA == contextB?true:false
+
+        then:
+        result == true
+    }
```
jorg3lopez, line 16: I want to add a test case to test the retrieval of the implementations. Currently, we have no classes to test this with, but I'm open to suggestions. Let me know your thoughts.

halprin: You could test it with a standard Java class. For example,

```java
applicationContext.register(String.class, "DogCow");
```

halprin:
I can tell while looking at this PR's description on my mobile phone whether #13 will be closed upon merge, but this PR will not fulfill all the acceptance criteria of #13. So, I'd say remove the "resolves" thing.

app/src/main/java/gov/hhs/cdc/trustedintermediary/context/ApplicationContext.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public static <T> T getImplementation(Class<T> clazz) {
+        return (T) OBJECT_MAP.get(clazz);
+    }
+
+    public static ApplicationContext getInstance() {
+        if (applicationContext == null) {
+            synchronized (ApplicationContext.class) {
+                if (applicationContext == null) applicationContext = new ApplicationContext();
```
halprin, line 30: Mind putting curly braces on this if statement? I think we should always use them. I'm actually a little surprised that the code formatter allowed this.

app/src/main/java/gov/hhs/cdc/trustedintermediary/context/ApplicationContext.java
```diff
@@ -0,0 +1,35 @@
+/**
+ * ApplicationContext class works similar to a factory. The idea is to use this class in conjunction
+ * with annoations that will be used to inject the implementations. There is the option of using the
```
halprin, line 3: Just a small typo.
```suggestion
 * with annotations that will be used to inject the implementations. There is the option of using the
```

halprin (commented):
Nice threading protection. I just have a few comments, but this looks great!

app/src/test/groovy/gov/hhs/cdc/trustedintermediary/ApplicationContextTest.groovy
jorg3lopez: Yes! We think alike. I was thinking in those parameters. I'll update it when I'm able to, thanks!

app/src/main/java/gov/hhs/cdc/trustedintermediary/context/ApplicationContext.java
jorg3lopez: I'll get on it. I think I had it with curly braces, but the formatter changed it. I'll double check.

jorg3lopez: I was able to add the curly braces!

app/src/main/java/gov/hhs/cdc/trustedintermediary/context/ApplicationContext.java
jorg3lopez: Good catch!

app/src/test/groovy/gov/hhs/cdc/trustedintermediary/ApplicationContextTest.groovy
```diff
@@ -0,0 +1,27 @@
+package gov.hhs.cdc.trustedintermediary.context
```
halprin, line 1: I just realized, this file is in the wrong folder.  It also needs to be added into a `context` sub-folder.

jorg3lopez: Good catch, I imported the class.

halprin: I meant you should go the other way.  Keep the `.context` on the `package` line and move this file into a newly created `context` sub-folder so it mirrors the `src/main`.

PR #46, opened 2022-10-19
13: Initial API Domain Handling

halprin:
# Initial API Domain Handling

This PR adds the concept of a "domain".  We currently only have one domain: ETOR (`etor`).  These domains register themselves with the larger application through a plug-in like architecture.  There could be future domains that the trusted intermediary could have.  It is the intention that domains (including the current ETOR one) could be extracted into a separate JAR, Gradle project, or even repository if we ever wanted to.

The ETOR domain in this PR is stubbed out.  It creates a `/v1/etor/order` endpoint that I expect to change in future PRs.

This plug-in architecture registers endpoints, and function handlers for those endpoints, as specified by the domain.  It also helps convert a Javalin (or whatever we may use in the future) request to a `DomainRequest` and a `DomainResponse` to a Javalin response (again, or anything we use in the future like Azure Functions).

## Issue

#13.


jorg3lopez (approved):
Great job, looks good.

PR #47, opened 2022-10-21
13: No need to construct ApplicationContext through singleton

halprin:
# No need to construct ApplicationContext through singleton

We actually don't need to construct the `ApplicationContext`.  All of its fields and methods are `static`.

## Issue

#13.

### DIFF
--- a/app/src/main/java/gov/hhs/cdc/trustedintermediary/context/ApplicationContext.java
+++ b/app/src/main/java/gov/hhs/cdc/trustedintermediary/context/ApplicationContext.java
@@ -1,8 +1,6 @@
 /**
  * ApplicationContext class works similar to a factory. The idea is to use this class in conjunction
- * with annotations that will be used to inject the implementations. There is the option of using
- * the getImplementation method to get an implementation of a class. This is a double checked
- * locking singleton class.
+ * with annotations that will be used to inject the implementations.
  */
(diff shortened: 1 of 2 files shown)


app/src/main/java/gov/hhs/cdc/trustedintermediary/context/ApplicationContext.java
```diff
@@ -11,7 +9,6 @@
 
 public class ApplicationContext {
 
-    private static volatile ApplicationContext applicationContext = null;
     private static final Map<Class<?>, Object> OBJECT_MAP = new HashMap<>();
```
jorg3lopez, line 12: I was thinking that this HashMap will get accessed throughout many parts of the codebase. Should we make it a concurrent HashMap or maybe a synchronized one? The concurrent HashMap might give us nice optimization when adding or retrieving. What are your thoughts?

jorg3lopez (approved):
Nicely done!

app/src/main/java/gov/hhs/cdc/trustedintermediary/context/ApplicationContext.java
halprin: Oh, good point.  Normally we aren't _writing_ to this `HashMap` once the application is booted up, but that's a good safety mechanism.  I'll make that change in this PR.

PR #55, opened 2022-10-24
13: Groovy and Spock ADR

halprin:
# Groovy and Spock ADR

Groovy and Spock ADR.

## Issue

#13.


jorg3lopez (approved):
Spock is growing into me. I like how we can use the pojo validator with it; great job!

PR #57, opened 2022-10-24
13: Code Coverage

halprin:
# Code Coverage

Add checking for code coverage.  We require 90% or greater.  We allow exclusions, but be judicious about what is added.

## Issue

#13.


jorg3lopez (approved):
Great job!

PR #80, opened 2022-11-10
13: Dependency Injection

halprin:
# Dependency Injection

Added the ability to do dependency injection via a `@Inject` annotation.

## Issue

#13.


jorg3lopez (approved):
Great job, everything looks good.

PR #92, opened 2022-11-29
Logging Dependency

jorg3lopez:
# Logger

This PR adds logging functionality to the code base. The dependency was isolated in a wrapper class in order to align with the OeSA architecture standards.

If changes to the logger need to be made, then we only need to make those changes in the wrapper class instead of multiple areas of the application.

# 
#### Addresses:
Story 13 - logging framework

https://app.zenhub.com/workspaces/cdc-ti-633de70d939f6c00155f3722/issues/cdcgov/trusted-intermediary/76


app/build.gradle
```diff
@@ -28,6 +28,8 @@ dependencies {
     implementation 'org.slf4j:slf4j-simple:2.0.5'
     implementation 'org.reflections:reflections:0.10.2'
     implementation 'javax.inject:javax.inject:1'
+    implementation 'org.slf4j:slf4j-api:2.0.3'
+    implementation 'org.slf4j:slf4j-simple:2.0.3' // Can be replaced with logBack, log4j, log4j2, Java Utility Logger
```
halprin, line 32: We already have this dependency a few lines above.

app/build.gradle.orig
```diff
@@ -0,0 +1,98 @@
+/*
```
halprin, line 1: I presume this file can be deleted?

app/src/main/java/gov/hhs/cdc/trustedintermediary/wrappers/TiLogger.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Humble object interface for logging This class was created in order to take away the 3rd party
+ * dependency of the logger. The idea is to have the logger dependency only in this class. If there
+ * ever is a reason to use a different logger, then we only need to make the changes here.
+ */
+public class TiLogger implements MyLogger {
```
halprin, line 11: I would do slightly different names. I would name this class something Kike `Slf4jLogger`. Basically something that indicates it is specific to SLF4J. I would rename the interface to be something like `TiLogger` or just `Logger`. Basically the interface should have the generic name that transcends any implementation.

app/src/main/java/gov/hhs/cdc/trustedintermediary/wrappers/TiLogger.java
```diff
@@ -0,0 +1,61 @@
...
+    // ANSI escape code
+    public static final String ANSI_RESET = "\u001B[0m";
+    public static final String ANSI_BLACK = "\u001B[30m";
+    public static final String ANSI_RED = "\u001B[31m";
+    public static final String ANSI_GREEN = "\u001B[32m";
+    public static final String ANSI_YELLOW = "\u001B[33m";
+    public static final String ANSI_BLUE = "\u001B[34m";
+    public static final String ANSI_PURPLE = "\u001B[35m";
+    public static final String ANSI_CYAN = "\u001B[36m";
+    public static final String ANSI_WHITE = "\u001B[37m";
```
halprin, line 24: Super cool!

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/javalin/App.java
```diff
@@ -13,6 +14,8 @@ public static void main(String[] args) {
 
         app.get("/health", ctx -> ctx.result("Operational"));
 
+        ApplicationContext.register(TiLogger.class, TiLogger.getLogger());
```
halprin, line 17: We should register the logger under the interface name. So it would become `MyLogger.class`. Or just `Logger.class` if you take my rename suggestion I made in another comment.

halprin (changes requested):
I've got a few suggestions. This is great though! I especially think the coloring is cool.

app/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/javalin/DomainsRegistrationTest.groovy
```diff
@@ -26,6 +28,7 @@ class DomainsRegistrationTest extends Specification {
         javalinContext.body() >> bodyString
         javalinContext.url() >> urlString
         javalinContext.headerMap() >> headerMap
+        ApplicationContext.register(TiLogger.class, TiLogger.getLogger()) // Needed since DomainRegistrationTest resets it
```
jorg3lopez, line 32: Maybe it can go in a def setup(){} / def setupSpec(){} method if more test cases need it.

app/build.gradle
jorg3lopez: Good catch

app/build.gradle.orig
jorg3lopez: Deleted! No more orig. ;)

app/src/main/java/gov/hhs/cdc/trustedintermediary/wrappers/TiLogger.java
jorg3lopez: Yes, I agree. This was the part that I had a bit of trouble with. I didn't go with the Logger at first because of the different options for import. But I like that you pointed it out. I do like the Logger and Slf4jLogger names better.

app/src/main/java/gov/hhs/cdc/trustedintermediary/wrappers/TiLogger.java
jorg3lopez: Thumbs up!

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/javalin/App.java
jorg3lopez: I love the idea.

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/javalin/DomainsRegistration.java
```diff
@@ -15,9 +16,17 @@
...
+
+    private static Slf4jLogger logger = ApplicationContext.getImplementation(Slf4jLogger.class);
+
     public static void registerDomains(
             Javalin app, Set<Class<? extends DomainConnector>> domainConnectors) {
 
+        logger.logInfo("Info");
+        logger.logWarning("Warning");
+        logger.logDebug("Debug");
+        logger.logError("Error");
```
halprin, line 28: Were these here for testing?

app/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/javalin/DomainsRegistrationTest.groovy
halprin: I would be OK to changing this to `TestApplicationContext.register(...)` and in the constructor of the `TestApplicationContext`, we always register the logger.  Also, it should be registered under `Logger.class`, not `Slf4jLogger.class`.

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/javalin/App.java
```diff
@@ -13,6 +14,8 @@ public static void main(String[] args) {
 
         app.get("/health", ctx -> ctx.result("Operational"));
 
+        ApplicationContext.register(Slf4jLogger.class, Slf4jLogger.getLogger());
```
halprin, line 17:
```suggestion
        ApplicationContext.register(Logger.class, Slf4jLogger.getLogger());
```
We want to register the `Slf4jLogger` under just the `Logger` class.

halprin (changes requested):
I like the rename, but I'd appreciate that we register the implementation in the Application Context under its interface.  More details in the inline comments.

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/javalin/DomainsRegistration.java
jorg3lopez: Yeah, those come up when the app is fired up. I can take them down, it's only a reference for the color.

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/javalin/App.java
jorg3lopez: Good catch, having the parent is the way to go. I'm still working on having the implementation in the ApplicationContext

halprin (approved):

Issue #76, opened 2022-11-10
Receiving Orders


JohnNKing:
# Story

As a Hospital, so that my newborn screening order can reach a lab, I need the TI service to be able to receive the order.

## Pre-conditions
- Only a single, fake hospital is supported.
- The order is a mock order.

## Acceptance Criteria
- Orders must be transmitted securely in non-development environments.
- The hospital receives confirmation that the order was successfully received.

## Tasks
- [x] ADRs.
  - [x] Java
  - [x] Gradle - @halprin 
  - [x] Groovy and Spock for testing - @halprin 
  - [x] OeSA - @halprin 
- [x] Basic Linting
- [x] End-to-end tests - @halprin
- [x] CD to at least a Flexion Azure environment - @JohnNKing
  - [x] Terraform - @JohnNKing 
- [x] OeSA application context - @halprin @jorg3lopez 
- [x] OeSA dependency injection - @halprin @jorg3lopez @tjohnson7021 
- [x] Code coverage check for at least 90%. - @halprin 
- [x] Add domain plugin architecture - @halprin
- [x] Add domain endpoint
- [x] SBOM generation - @tjohnson7021 
- [x] Learn more about FHIR - @halprin
- [x] Logging framework - @jorg3lopez 
- [x] Containerize application with docker @jorg3lopez 
  - [x] Docker image @jorg3lopez 
  - [x] Docker-compose @jorg3lopez 
- [x] Ingest an "order" like JSON object for the request. @jorg3lopez 
- [x] Output some kind of acceptance JSON object for the response. @jorg3lopez 

### Engineering
- [x] Threat model updated - @jorg3lopez 
- [x] Code refactored for clarity and no design/technical debt
- [x] Adhere to separation of concerns; code is not tightly coupled, especially to 3rd party dependencies.
- [x] Source code is merged to the main branch.
- [x] Unit test coverage of our code >= 90%
- [x] Code is reviewed or developed by pair; 1 approval is needed but consider requiring an outside-the-pair reviewer.
- [x] Build process updated
- [x] Feature toggles created and/or deleted.  Document the feature toggle.
- [x] Source code documentation created when the code is not self-documenting. - @tjohnson7021 
- [x] API documentation generated
- [x] Code quality checks passed
- [x] Security & privacy gates passed
- [x] Load tests passed - @tjohnson7021 
- [x] Documentation and diagrams created or updated - @jorg3lopez 
- [x] API(s) are versioned
[...]

PR #100, opened 2022-12-09
Docker container and Azure deployment

JohnNKing:
# Docker container and Azure deployment
The application is now containerized and automated deployment to a staging Azure environment is performed via GitHub actions.

The existing GitHub CI workflow must finish successfully before the staging deployment will take place. Also, the container build and deployment workflow has been designed to be reusable (i.e. used for environments other than staging).

For demonstrations of working builds, see https://github.com/CDCgov/trusted-intermediary/actions/workflows/cicd.yml

## Issue
https://github.com/CDCgov/trusted-intermediary/issues/76


.github/workflows/codeql_reusable.yml
```diff
@@ -1,3 +1,5 @@
+name: CodeQL
```
halprin, line 1: I think the reason I didn't have this here originally was because this text doesn't show up anywhere.  So I omitted it.

Dockerfile
```diff
@@ -0,0 +1,15 @@
+# Use Linux-Alpine image
+FROM amazoncorretto:17-alpine
+
+ARG JAR_LIB_FILE=./app/build/libs/app-all.jar
+# Create directory and switch to it
+WORKDIR /app
+
+# Add project to created folder
```
halprin, line 8: Can we change this comment?  We aren't addint the entire project anymore.

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/javalin/App.java
```diff
@@ -13,6 +13,7 @@ public class App {
     public static void main(String[] args) {
         var app = Javalin.create().start(8080);
 
+        app.get("/", ctx -> ctx.result("Welcome to the CDC TI service"));
```
halprin, line 16: Should this get removed?  I don't personally see a reason for this, but I'm curious what others think?

halprin (commented):
Super awesome!  I'm not approving just yet due to some of my outstanding thoughts and questions on specific lines.  Also, when you look at the files in this PR, many of them don't have newlines at the end of the file.  I know it's kind of dumb, but I'd like to have all our files have and ending newline.

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/javalin/App.java
jorg3lopez: I'm ok with removing it.

.github/workflows/codeql_reusable.yml
JohnNKing: I'm not 100% certain, but what I could find indicates the name would only replace the path in the GitHub Actions UI once merged to main. (I believe deploy.yml was behaving the same until I gave it a name. Well... in this case it just disappeared, but I suspect as a workflow with no runs it will show up again once it exists on main.)

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/javalin/App.java
JohnNKing: I'm cool with removing it. However, without something to respond to a 'GET /' against the app's Azure deployment url, the result is a 404. In the future I expect we may want to direct the guest to our documentation or other information about how to use the API.

Dockerfile
JohnNKing: Updated

halprin (approved):

PR #108, opened 2022-12-21
Terraform

JohnNKing:
# Terraform

Utilizing Terraform within the GitHub workflows to provision Azure resources as part of the deployment process.

## Issue
https://github.com/cdcgov/trusted-intermediary/issues/76


halprin (approved):
I think this is the new hotness, but I've been heavily involved with this.  Does anyone else want to review?

PR #115, opened 2022-12-27
Load Tests - modified for latency

ghost:
# Load Tests - modified for latency 

- modified load tests so that tests consider response times, not just returned http status codes.

## Issue 76

[Definition of Done](https://app.zenhub.com/workspaces/cdc-ti-633de70d939f6c00155f3722/issues/gh/cdcgov/trusted-intermediary/76)


sfradkin (changes requested):
This seems reasonable as a starting point, but we need to be aware that PR #110 will need to add to the code for the load tests.

Let's add some constants for the re-used strings and the max successful response time.
Also, please check the latency functions as the code in each should be identical, but it isn't.  Typically, anywhere with duplicated code we would refactor to avoid copy/paste errors.

We also need a way to invoke the locust test.  We need to decide when it should be run and add it to one of the GHA workflows.

tjohnson7021:
Thank you for the feedback @sfradkin . I started adding some constants, and fixed the inconsistency in the latency tests. I will pair with the team to access the remainder of the concerns.

halprin (changes requested):
Great change to verify the response times!

1. I would collapse the different tasks into one for each endpoint.  E.g. one for the `/health` and one for `/v1/etor/order`.
2. There's no need to verify the response body.  This is a load test that only cares about the response timings.  Our e2e tests are the ones that verify the response bodies.
3. Does the response time checking code verify on each individual response or is it an average of some sort?  It _seems_ like it is each individual response.  I would change it to be a failure only if it is over a second given the 95% percentile of all response times in the session.

halprin (approved):
🎸

sfradkin (approved):

PR #116, opened 2023-01-02
fixing typos

sinapis-amicus:
fixing a few typos which should not affect functionality


halprin (approved):
Thanks for finding these typos.

Issue #77, opened 2022-11-10
Receiving Newborn Demographic Data in a Well-Known Format


JohnNKing:
# Story

## Sentence

As a hospital, so that newborn demographic data can be processed, I need the TI service to be able to receive the data in a well known format.

## Pre-conditions
- Only a single, fake hospital is supported.
- A well known format of our choice is supported.

## Acceptance Criteria
- Demographic data must be transmitted securely in non-development environments.
- The hospital receives confirmation that the newborn demographic data was successfully received **by the TI service**. (Notification from the lab is not yet expected)

### Research
- [x] Review HL7 LOI IG and map data elements on NBS cards
- [x] Map ADT-01 IG to HL7 LOI to compare overlap
- [x] Decide on the protocol and format for receiving newborn demographic data
- [x] Decide on the internal format to support
- [x] Decide if next of kin info will come from the FHIR patient resource - it will!
- [x] Determine a list of demographic data fields and types (from FHIR patient profile & NBS cards; review LOI; cross reference FHIR patient profile with per-state spreadsheet)
- [x] Get an example(s) of ADT HL7v2 messages for a newborn: https://confluence.hl7.org/display/OO/v2+Sample+Messages

###  Engineering
- [x] Parse FHIR format message (via a FHIR library?) - @halprin, @tjohnson7021 
- [x] Decide if next of kin is in scope
- [x] Parse out next of kin - @halprin 
- [x] Decide on acceptance response (also FHIR? custom?)
- [x] Return acceptance response
- [x] Foundational: DAST scanning -@Jorge
- [x] Foundational: API documentation (OpenAPI) - @tjohnson7021, @halprin
- [x] Foundational: Improved Linting (Sonar)  - @halprin 
- [x] Change e2e tests (`e2e-execute.sh`) to run through the JAR file instead of Gradle. @Jorge
- [x] Add support for the request body to be a FHIR bundle instead of _just_ a Patient resource. - @halprin
- [x] Add support for the [US Core Profile](https://hl7.org/fhir/us/core/) for the [Patient resource](https://hl7.org/fhir/us/core/StructureDefinition-us-core-patient.html).  This will probably add support for some of the missing fields that we didn't find in the default [Patient resource](https://www.hl7.org/fhir/patient.html). - @halprin 
- [x] Foundational: Arrange meeting with Security Engineer and/or ISSO to talk ATO - @JohnNKing
- [x] Looking into how to store private keys in Azure. - @halprin 
[...]

PR #122, opened 2023-01-09
77: Sonar Linting

halprin:
# Sonar Linting

Added Sonar linting from SonarCloud.  We don't have a check that will fail the PR yet, but I'll need to get this in so it can start scanning `main` first I think.

## Issue

#77


jorg3lopez (approved):
Great job, everything looks good.

PR #132, opened 2023-01-13
77: Reference OpenAPI YAML that works in a JAR file

halprin:
# Reference OpenAPI YAML that works in a JAR file

Reference OpenAPI YAML that works in a JAR file

## Issue

#77.


jorg3lopez (approved):
Great job!

PR #135, opened 2023-01-16
Updating story template

JohnNKing:
# Story Template Update

Updating story template to break out research and engineering tasks. Also adding and removing sections based upon recent stories.

## Issue

n/a

### DIFF
--- a/.github/ISSUE_TEMPLATE/story.md
+++ b/.github/ISSUE_TEMPLATE/story.md
@@ -10,10 +10,10 @@ assignees: ''
 # Story
 As a _, so that _, I need _.
 
-### Pre-conditions
+## Pre-conditions
 - [ ] *Assumptions of prior or future work that's out of scope for this story*
 
-### Acceptance Criteria
+## Acceptance Criteria
 - [ ] *Required outcomes of the story*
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/story.md
```diff
@@ -8,32 +8,25 @@ assignees: ''
 ---
 
 # Story
+As a _, so that _, I need _.
 
-## Sentence
+### Pre-conditions
```
halprin, line 13: We should only decrease by 1 heading size.  So, instead of `###`, this should be `##`.  This is an accessibility thing.

halprin (changes requested):

.github/ISSUE_TEMPLATE/story.md
JohnNKing: Thanks for calling that out @halprin -- fixed. (I didn't care for the extraneous dividing lines GH was adding, but I didn't know for certain there were accessibility implications.)

halprin (approved):
Ship it!

sfradkin (approved):
I like these additions to the template.  These will be very helpful.

PR #141, opened 2023-01-26
DAST Scanning

jorg3lopez:
# Dynamic Application Security Test 

This PR adds DAST scanning to our application. The scanning is done as part of a workflow. The event gets triggered when pushing to the main branch. 
Currently, the workflow fires up when pushing to the remote branch, this was done on purpose in order to help with testing. Once the PR gets approved, the DAST scanning will get triggered every time something gets pushed to the main branch.

## Changes
New Files
- health-check.sh
- dast.yml

health-check.sh - Checks to see if the api is functional by accessing the health endpoint.

dast.yml - Workflow that uses OWASP ZAP for DAST scanning

## Testing
You want to test the script and also the workflow.

### Testing the script:
- run shadowJar: `./gradlew shadowJar`
- build docker container: `docker compose up --build -d`
- run the script: `./health-check.sh`

### Testing the workflow:
The easiest way to do this would be to add a minor change to the branch and then push it.
Check the Actions on github and make sure everything passes.
There should be an artifact that contains warnings such as 404 status codes. 

## Issue
#77.


app/src/main/java/gov/hhs/cdc/trustedintermediary/OpenApi.java
```diff
@@ -21,6 +21,7 @@ public static OpenApi getInstance() {
 
     public String generateApiDocumentation(final Set<String> openApiSpecifications) {
         openApiSpecifications.add(getBaselineDocumentation());
+        openApiSpecifications.add(getEtorDocumentation());
```
jorg3lopez, line 25: I think we can refactor this better. Maybe we can create a method that adds all of the existing documentation. It feels like `getBaselineDocumentation()` and `getEtorDocumentation()` do a similar job. What are your thoughts?

app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
```diff
@@ -23,16 +23,19 @@ public class EtorDomainRegistration implements DomainConnector {
...
     @Inject Logger logger;
 
+    private Map<HttpEndpoint, Function<DomainRequest, DomainResponse>> domains =
+            Map.of(new HttpEndpoint("POST", "/v1/etor/order"), this::handleOrder);
+
     @Override
     public Map<HttpEndpoint, Function<DomainRequest, DomainResponse>> domainRegistration() {
         ApplicationContext.register(OrderController.class, OrderController.getInstance());
-        return Map.of(new HttpEndpoint("POST", "/v1/etor/order"), this::handleOrder);
+        return domains;
```
jorg3lopez, line 32: Better naming suggestions are welcome.

app/src/main/java/gov/hhs/cdc/trustedintermediary/OpenApi.java
```diff
@@ -42,4 +43,13 @@ String getBaselineDocumentation() {
             throw new RuntimeException(e);
         }
     }
+
+    String getEtorDocumentation() {
+        try (InputStream openApiStream =
+                getClass().getClassLoader().getResourceAsStream("openapi_etor.yaml")) {
+            return new String(openApiStream.readAllBytes(), StandardCharsets.UTF_8);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
jorg3lopez, line 52: This is giving me a code smell for the Sonar scan

.github/workflows/dast.yml
```diff
@@ -0,0 +1,46 @@
+name: ZAP
+
+on:
+  push:
+    branches: [ story/77/dast/jorg3 ]
```
jorg3lopez, line 5: This will change once it gets approved, it should be triggered when pushed to master. Other option can be having it run on every PR; thoughts?

app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
jorg3lopez: endpoints/ endpointsHandlers?

health-check.sh
```diff
@@ -0,0 +1,59 @@
...
+API_HEALTH_CHECK_FAILED="FAIL: API Health Check"
+
+is_container_running() {
+  if docker ps --format '{{.Names}}' | grep -Eq "^${CONTAINER_NAME}\$"; then
+    echo "$CONTAINER_PASSED"
+  else
+    echo "$CONTAINER_FAILED"
+    echo "$API_HEALTH_CHECK_FAILED"
+    exit 1
+  fi
```
jorg3lopez, line 24: Remove "if container is running"

app/src/main/java/gov/hhs/cdc/trustedintermediary/OpenApi.java
jorg3lopez: Did not need this line of code as the Etor documentation comes in via the parameter `openApiSpecifications`

app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
jorg3lopez: renamed to `endpoints`

app/src/main/java/gov/hhs/cdc/trustedintermediary/OpenApi.java
jorg3lopez: We don't need this code as the etor documentation comes in through the method parameter

health-check.sh
jorg3lopez: Done. This will make the script more general and it can be use with other events

health-check.sh
```diff
@@ -0,0 +1,46 @@
...
+            exit 1
+        fi
+        ((attempts=attempts+1))
+        echo 'Waiting for API to respond...'
+        wait
+    done
+
+      if [[ "$HTTP_CODE" -ne 200 ]]; then
+        echo "$API_HEALTH_CHECK_FAILED"
+        echo "status Code: $HTTP_CODE"
```
halprin, line 37:
```suggestion
        echo "Status Code: $HTTP_CODE"
```

health-check.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# Script to check health of API. It sends a curl command to the health endpoint and checks
+# that the status code is 200.
+# Note: This script can only be used when the application is running inside a container.
+
+set -e
+
+echo "API Health Check..."
+
+CONTAINER_NAME="trusted-intermediary-router-1"
```
halprin, line 11: I don't think this variable is being used.

.github/workflows/dast.yml
halprin: I think it should run on every PR and every push to `main`.

jorg3lopez: Good call.

health-check.sh
jorg3lopez: Good catch, thanks!

health-check.sh
jorg3lopez: You are right, this was part of the whole checking for the container; great catch!

.github/workflows/dast.yml
```diff
@@ -0,0 +1,47 @@
+name: ZAP
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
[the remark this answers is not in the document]
jorg3lopez: Great idea!

halprin (approved):

PR #142, opened 2023-01-27
End-To-End Runs Using JAR File

jorg3lopez:
# E2E Test Runs Using JAR

This PR addresses the issue where the app would work correctly while performing data retrieval (File Stream), running in gradle, but it would fail when running in a container or from a JAR file. This was due to the data having different formats (zip) at time of access. 

Gradle come with tools that decode these formats so it would automatically use them, that's why it was not failing when using gradle to run the app. On the other hand, changes were made such as using output streams instead of file stream, in order to prevent the data from getting serialized with a different format.

In order to prevent this from happening in the future, it is a good choice to run the application using the JAR file. Running the application using the JAR file will ensure that the same bug doesn't happen again.

## Testing
1. run `e2e-execute.sh`
2. make sure the test passes

You can go into the e2e-execute.sh file and add `-x` for debugging, at the top, in the `set` command. 
It should look like this: `set -e -x`

## Issue
#77.

### DIFF
--- a/e2e-execute.sh
+++ b/e2e-execute.sh
@@ -2,15 +2,18 @@
 set -e
 
 start_api() {
+  DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+  SUB_DIR="app/build/libs"
+  JAR_NAME="app-all.jar"
     echo 'Starting API'
-    ./gradlew --no-daemon app:clean app:run &
+    java -jar "${DIR}"/"${SUB_DIR}"/"${JAR_NAME}" > /dev/null &
     export API_PID="${!}"
     echo "API starting at PID ${API_PID}"
 }
 
 wait_for_api() {
     attempt_counter=0
-    max_attempts=36
+    max_attempts=5
 
     until curl --output /dev/null --silent --head --fail http://localhost:8080/health; do
         if [ "${attempt_counter}" -eq "${max_attempts}" ];then
--- a/e2e-execute.sh
+++ b/e2e-execute.sh
@@ -1,10 +1,15 @@
 #!/usr/bin/env bash
 set -e
 
+shadowJar() {
+  echo "Running shadowJar..."
+  ./gradlew shadowJar
+}
+
 start_api() {
-  DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-  SUB_DIR="app/build/libs"
-  JAR_NAME="app-all.jar"
+    DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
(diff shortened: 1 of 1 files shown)


e2e-execute.sh
```diff
@@ -1,16 +1,24 @@
 #!/usr/bin/env bash
 set -e
 
+shadowJar() {
+  echo "Running shadowJar..."
+  ./gradlew shadowJar
+}
```
jorg3lopez, line 7: We can live without the shadowJar being ran first, but I think it ensures that the shadowJar gets built thus avoiding compile issues.

e2e-execute.sh
```diff
@@ -1,16 +1,24 @@
...
-    ./gradlew --no-daemon app:clean app:run &
+    java -jar "${DIR}"/"${SUB_DIR}"/"${JAR_NAME}" > /dev/null &
     export API_PID="${!}"
     echo "API starting at PID ${API_PID}"
 }
 
 wait_for_api() {
     attempt_counter=0
-    max_attempts=36
+    max_attempts=20
```
jorg3lopez, line 21: I brought down the `max_attempts` to 20. I think it is more than enough attempts in order to check if the api is running. There is a 5 milliseconds interval per attempt. Let me know your thoughts if I should keep it at 36 attempts.

e2e-execute.sh
halprin: Good idea to have this here.  I would only add a `clean` to the front.

e2e-execute.sh
```diff
@@ -1,16 +1,24 @@
 #!/usr/bin/env bash
 set -e
 
+shadowJar() {
+  echo "Running shadowJar..."
+  ./gradlew shadowJar
```
halprin, line 6:
```suggestion
  ./gradlew clean shadowJar
```

e2e-execute.sh
halprin: It's actually 5 seconds.  But 1.6 minutes instead of 3 minutes should be fine.

halprin (commented):
Thanks dude!

e2e-execute.sh
jorg3lopez: We are on the same wavelength, I just added that.

e2e-execute.sh
jorg3lopez: I tried the `./gradlew clean shadowJar` but it didn't work. It seems to work with `./gradlew :app:clean shadowJar`

jorg3lopez: Nevermind, it was a typo; `./gradlew clean shadowJar` works!

halprin (approved):

tjohnson7021 (approved):

PR #155, opened 2023-02-01
77: ADT A01 Example

halprin:
# ADT A01 Example

Canonical ADT A01 example that we can use.

TODO:
- [x] Add birth order.
- [x] Ensure next of kin is good.
- [x] Ensure first name is good.
- [x] Ensure last name is good.
- [x] Ensure patient ID is good.
- [x] Ensure sex field is good.
- [x] Figure out if race can be free form text or not.
- [x] Make the birth time interesting instead of all 0s.

## Issue

#77.


jorg3lopez (approved):
Perfect!

PR #170, opened 2023-02-09
77: Next of Kin

halprin:
# Next of Kin

We now capture the next of kin's first name, last name, and phone number out of the `Patient` FHIR resource.  We do this in a tiered fashion.  We first look for the official "next of kin" contact.  If we don't find that contact, we look for the mother.  Lastly if mother isn't found, we look for the father.  If none of those three are there, we don't pick anything.

## Issue

#77.

### DIFF
--- a/app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/demographics/NextOfKin.java
+++ b/app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/demographics/NextOfKin.java
@@ -2,6 +2,7 @@ package gov.hhs.cdc.trustedintermediary.etor.demographics;
 
 import java.util.StringJoiner;
 
+/** Contains information about a next of kin relationship. */
 public class NextOfKin {
     private String firstName;
     private String lastName;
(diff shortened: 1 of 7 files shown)


tjohnson7021 (approved):

app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/demographics/NextOfKin.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.StringJoiner;
+
+/** Contains information about a next of kin relationship. */
+public class NextOfKin {
+    private String firstName;
+    private String lastName;
+    private String phoneNumber;
+
+    public NextOfKin(final String firstName, final String lastName, final String phoneNumber) {
```
sfradkin, line 11: Do we need to capture the type of next of kin in the `NextOfKin` object?

halprin: I don't know.  No one has mentioned it, so I wasn't going to capture any additional information until we have a more concrete understanding of what's needed.

jorg3lopez (approved):
Great job, everyting passes and looks good

Issue #79, opened 2022-11-10
Converting Newborn Demographic Data


JohnNKing:
# Story

## Sentence

As a public health lab, so that newborn demographic data can be utilized, I need the TI service to convert the data to a format the lab supports.

## Pre-conditions
- Only a single, fake public health lab is supported.

## Acceptance Criteria
- The newborn demographic data is converted to a well known format of our choosing.

### Research
- [x] Refine list of questions for PHLs to answer about data formats (questions below)
- [x] Decide on a format
- [x] Is loss of extraneous patient IDs (and other fields) acceptable when sending order details to ReportStream and subsequently the PHLs?

###  Engineering

- [x] Generate an example FHIR blob that would be sent to ReportStream.  Put it in the `/examples` directory.  Use the similar information that would be gleaned from our example newborn patient FHIR resource used in our e2e test. - @halprin 
- [x]  Convert demographics to the lab order - @halprin 
- [x] Send data to ReportStream - @jorg3lopez 
  - [x] Base level code to authenticate and send the FHIR resource - @jorg3lopez 
  - [x] (to complete) Fix-up the authentication flow so that it works - @jorg3lopez, @halprin 
- [x] Dynamically determine which ReportStream environment to send to - @halprin
- [x] Local version of sending data to the hard drive (`LocalFileLabOrderSender`) @tjohnson7021 
- [x] e2e tests to test the file created from "sending" the order to the local hard drive - @halprin, @tjohnson7021 
- [x] Correctly pick between the local hard drive "sender" implementation versus the ReportStream sender implementation when we are running locally versus running in a deployed environment. - @tjohnson7021
- [x] terraform to support key/auth infrastructure - @halprin 
- [x] Tackle [code smells](https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&id=trusted-intermediary) in SonarCloud (ignore generic exception ones for now?) - @halprin 
- [x] Modify SonarCloud's quality gate for our own use (in PRs and for the repo as a whole) - @halprin 
[...]

PR #184, opened 2023-02-14
79: Skeleton

halprin:
# Skeleton

We created the usecase that handles the basics that this story is about: converting and sending.

For converting, we have an interface, `LabOrderConverter`, that can have different concrete implementations.  This is abstracted because there can be different third-party FHIR classes we could use.  We don't want the Hapi, Linux4Health, or whatever else dependencies spreading through all of our code.  To this end, we also have a `LabOrder` interface that "wraps" a concrete FHIR dependency lab order.  This allows our usecase to not know the concrete implementation being used behind the `LabOrder`.  So, one of our concrete convert implementations, `HapiLabOrderConverter`, creates a `LabOrder` for the Hapi FHIR `Bundle` class.

For sending, we have an interface, `LabOrderSender`, that can have different concrete implementations.  We created two implementations that will be fleshed out in future PRs.  One is for ReportStream, `ReportStreamLabOrderSender`, which will handle all the work associated with communicating with RS (JWT handling, ReST calls, etc.).  This implementation will be used when we're deployed in an actual environment.  Another implementation, `LocalFileLabOrderSender`, will be used when we run our application locally.  This implementation will write out the order to the local hard drive.

## Issue

#79.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -118,7 +118,7 @@ tasks.named('test') {
             '**/javalin/App*',
             '**/jackson/Jackson*',
             '**/slf4j/Slf4jLogger*',
-            '**/hapi/HapiFhirImplementation'
+            '**/hapi/HapiFhirImplementation*'
         ]
     }
 }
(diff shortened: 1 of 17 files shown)


jorg3lopez (approved):
Looks good, great job!

PR #207, opened 2023-02-23
79: Example FHIR Lab Order

halprin:
# Example FHIR Lab Order

Created an example FHIR lab order that comes out of TI, as of today, given the example input in [newborn_patient.json](https://github.com/CDCgov/trusted-intermediary/blob/main/e2e/src/test/resources/newborn_patient.json).

## Issue

#79.


examples/fhir/lab_order.json
```diff
@@ -0,0 +1,118 @@
...
+                "system": "phone",
+                "value": "+31201234567"
+              }
+            ]
+          }
+        ]
+      }
+    },
+    {
+      "resource": {
```
sfradkin, line 97: We may also need to include the `category`.  There is a SNOMED code for `laboratory procedure`.

Consider adding a `requester`

sfradkin (changes requested):
Some changes and comments for considering optional fields to include.

examples/fhir/lab_order.json
```diff
@@ -0,0 +1,118 @@
+{
+  "resourceType": "Bundle",
+  "id": "278637ea-386c-4e24-8aad-d3b2ec088332",
+  "type": "message",
+  "timestamp": "2023-02-23T15:42:22.877-06:00",
+  "entry": [
+    {
+      "resource": {
+        "resourceType": "MessageHeader",
```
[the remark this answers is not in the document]
halprin: Good point.  I can add the `source`.

As for the `destination` and `focus`, I have no idea what I would put for them.  Yes, I could make something up.  Since they have an optional cardinality, I'm inclined to not put them in until we have stronger evidence that indicates we need them.

examples/fhir/lab_order.json
halprin: Both the `category` and `requester` have a cardinality of optional.  Is there something thats making you feel we need to add these?  I basically completely cooked this up from the required fields (if you couldn't tell).

examples/fhir/lab_order.json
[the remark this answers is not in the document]
halprin: Added the `source`.

examples/fhir/lab_order.json
```diff
@@ -0,0 +1,122 @@
...
+        "multipleBirthInteger": 1,
+        "contact": [
+          {
+            "relationship": [
+              {
+                "coding": [
+                  {
+                    "system": "http://terminology.hl7.org/CodeSystem/v2-0131",
+                    "code": "N",
+                    "display": "Next of kin"
```
jorg3lopez, line 79: Would this be optional to include the next of kin of choice, or should the mother be included in the contact info? I'm wondering as I was looking at the example that FHIR.org has:

http://build.fhir.org/patient-example-infant-twin-1.json.html

jorg3lopez (approved):
I think is a good start for a sample message. I know things will change as we dive deeper in research and as we get to know what the clients will want.  Good job!

examples/fhir/lab_order.json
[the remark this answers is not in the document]
sfradkin: The `destination` we can probably leave off for now or just put a fake lab... it would be the lab that the order is going to.

The `focus` should be the Patient.

examples/fhir/lab_order.json
sfradkin: This is the correct way to add a next of kin, so we should put it here.

examples/fhir/lab_order.json
sfradkin: Just what I've gleaned from other examples I've found.  But yes, those are not required, and they don't appear to map to an OML, but I think they add context.

halprin: Added the `category`.

sfradkin (approved):

PR #212, opened 2023-03-01
Removing FHIR Utilities as a Dependency

halprin:
# Removing FHIR Utilities as a Dependency

The explicit FHIR Utilities dependency was added to remove a vulnerability.  It seems the latest version of the other FHIR dependencies depend on a newer version of FHIR Utilities that doesn't have the vulnerability.


jorg3lopez (approved):
Looks good

PR #213, opened 2023-03-01
79: Terraform for Key Vault

halprin:
# Terraform for Key Vault

We now create an Azure key vault.  I decided not to use the existing ReportStream key vault Terraform module because it has way more required options than we need for now.  I'm sure a lot of them are needed for good security and such, but was going to be overkill for now.  I expect PRs in the future for us to make the RS key vault Terraform module more generic and for our application to make use of it.

This key vault is pre-filled in with a secret that represents our ReportStream sender private key.  Terraform is instructed to ignore the value of that secret so it doesn't try to overwrite it when we customize it with an actual private key.

This key vault has two sets of permissions.  One is for our GitHub deployer the ability to create, read, and delete secrets.  The delete is needed so that GitHub can teardown the PR environment automatically.  The other is for our application the ability to read the secret so it can do the appropriate JWT signing.  To do this, I had to tell Terraform that the application needed a system generated ID.

## Issue

#79.

### DIFF
--- a/.github/workflows/deploy_reusable-skip.yml
+++ b/.github/workflows/deploy_reusable-skip.yml
@@ -1,0 +1,20 @@
+# Used when a PR doesn't have any changes that require the PR deploy.
+# The jobs' keys/names must align with the `deploy_reusable.yml` jobs' keys/names.
+
+name: 'Container Build & Deploy'
+
+on:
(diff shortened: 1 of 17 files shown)


jorg3lopez (approved):
Good Job!

PR #224, opened 2023-03-10
Task/79 setting environmental context

tjohnson7021:
# Setting Environmental Context for App Context

 - Adding environmental context "labels" to help determine if the instance is in LOCAL, STAGING, PROD, etc


halprin (approved):
Boom!

PR #228, opened 2023-03-13
79: Make the Key Vault Name Globally Unique

halprin:
# Make the Key Vault Name Globally Unique

The Azure Key Vault name needs to be globally unique.  Unsurprisingly, someone else already has `key-vault-staging`.  So we are making the name slightly more unique.

## Issue

#79.


jorg3lopez (approved):
Good idea!

PR #231, opened 2023-03-14
79: Inject the ReportStream Domain

halprin:
# Inject the ReportStream Domain

Pass the ReportStream domain we should be using to our application through the `REPORT_STREAM_URL_PREFIX` environment variable.

Construct this environment variable in Terraform so we use a real version when running in our deployed environment.

## Issue

#79.


jorg3lopez (approved):
Great work, this will help us a lot!

PR #232, opened 2023-03-14
79: Fix Some SonarCloud Smells

halprin:
# Fix Some SonarCloud Smells

- Move some JavaDocs to be above the class.
- Add some JavaDocs to some classes where they were missing.
- Make the logging string concatenation be "lazy" so the concatenation only happens if the logging level is enabled.

## Issue

#79.


jorg3lopez (approved):
Looks good, great job!

Issue #84, opened 2022-11-18
Sending Newborn Demographic Data


JohnNKing:
# Story

## Sentence
As a public health lab, so that the newborn demographic data will be received, I need the TI service to send the data to the lab.

## Pre-conditions
- The TI service can receive newborn demographic data (mock or in a specific format) from a mock hospital
- A mock public health lab destination is supported

## Acceptance Criteria
- The mock public health lab has received the newborn demographic data

###  Engineering
- [x] Get our ReportStream Client ID
- [x] Send data to ReportStream - @jorg3lopez 
-   [x] Get the private key from Azure or locally - @jorg3lopez 
  - [x] Fix-up the sending of FHIR resources flow so RS accepts it (if needed, maybe it just works) - @halprin 
- [x] This should work in the staging environment
- [x] Ensure that the secret key we get from Azure is only gotten once and then cached. - @tjohnson7021 
- [x] Ensure that the RS API key is cached so we don't ask for the RS API key for every TI API call. - @jorg3lopez 
- [x] Experiment with not loosing any additional information from the original FHIR input. - @halprin 
- [x] Upload our staging private key to the Azure key vault
- [x] Update our Demographics -> Lab Order logic to support [Arnej's additions](https://github.com/CDCgov/trusted-intermediary/pull/246). - @tjohnson7021 
- [x] Add better custom exceptions when sending stuff to RS and Azure keys.
  - [x] `AuthEngine` and implementations. - @halprin 
  - [x] `HttpClient` and implementations. - @jorg3lopez 
  - [x] `Secrets` and implementations. - @halprin 
  - [x]  Stitch it together in the `LabOrderSender` and implementations. - @jorg3lopez @halprin 
- [x] Get our logging in Azure into one spot. - @halprin 
- [x] Create our dev environment.  Meant to be a wild-west environment where any developer can deploy to at any time. - @halprin 

### Engineering
- [x] Threat model updated
- [x] Code refactored for clarity and no design/technical debt @jorg3lopez  @basiliskus 
- [x] Adhere to separation of concerns; code is not tightly coupled, especially to 3rd party dependencies. @jorg3lopez 
- [x] Source code is merged to the main branch. @jorg3lopez @basiliskus 
- [x] Unit test coverage of our code >= 90% @jorg3lopez @basiliskus 
- [x] Code is reviewed or developed by pair; 1 approval is needed but consider requiring an outside-the-pair reviewer. - @tjohnson7021 
[...]

PR #237, opened 2023-03-16
84: Lossless Patient

halprin:
# Lossless Patient

As I mentioned previously, we only capture very specific fields out of the `Patient` resource.  All other things are stripped away.  This PR changes the approach by making the `Patient` (and everything in the passed in `Bundle` for that matter) be passed through unchanged and therefore lossless.  This lossless is accomplished by using another "wrapper" class similar to how we did the `LabOrder`.  A side benefit of this is _a lot_ of code was able to be deleted!

## Issue

#84.


jorg3lopez (approved):
This is some good stuff, great work!

PR #241, opened 2023-03-17
RS Connection - Response Body

jorg3lopez:
# RS Connection 

This PR adds the logic to the ApacheClient humble object, to return the content (body) when sending a POST request. This is the way we acquire the token from RS.

## Issue

#79

### DIFF
--- a/app/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamLabOrderSender.java
+++ b/app/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamLabOrderSender.java
@@ -42,14 +42,13 @@ public class ReportStreamLabOrderSender implements LabOrderSender {
         sendRequestBody(json, bearerToken);
     }
 
-    protected String sendRequestBody(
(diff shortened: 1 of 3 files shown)


app/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamLabOrderSender.java
```diff
@@ -42,14 +42,13 @@ public void sendOrder(final LabOrder<?> order) {
...
     }
 
-    protected String sendRequestBody(
-            @NotNull String json, @NotNull String bearerToken) { // url param?
+    protected String sendRequestBody(@NotNull String json, @NotNull String bearerToken) {
         String res = "";
         Map<String, String> headers =
                 Map.of(
                         "Authorization", "Bearer " + bearerToken,
                         "client", "flexion",
```
halprin, line 50: The `client` should be `flexion.etor-service-sender`.

jorg3lopez: Good catch!

halprin (approved):

PR #277, opened 2023-04-14
84: More Unit Tests for the Controller

halprin:
# More Unit Tests for the Controller

- Made the controller return an internal server error when we can't marshal the response into JSON.
- I added more unit tests for the controller.

## Issue

#84.


jorg3lopez (approved):
Great addition, good job!

PR #280, opened 2023-04-17
Task/84: Private Key Caching in RS Sender

ghost:
# Private Key Caching for RS Sender

We don't want to repeatedly ask for the key from Azure every time there is a request, so we cache it. 

- added private key cache (thread-safe)
- added unit tests for new methods


app/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamLabOrderSender.java
```diff
@@ -33,6 +34,16 @@ public class ReportStreamLabOrderSender implements LabOrderSender {
 
     private static final String CLIENT_NAME = "flexion.etor-service-sender";
 
+    private String cachedAzureKey;
```
tjohnson7021, line 37: Does this really need to evolve into a map if the key is only saved in the session memory? Wouldn't the necessary key change once the environment changes?

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamLabOrderSender.java
```diff
@@ -105,7 +116,19 @@ protected String requestToken() throws UnableToSendLabOrderException {
         return token;
     }
 
+    protected String retrieveAzureKey(String senderPrivateKey) throws SecretRetrievalException {
```
halprin, line 119: I would remove the need for this variable to be passed in and instead construct that string in this method just like it is outside this method.

`var senderPrivateKey = "report-stream-sender-private-key-" + ApplicationContext.getEnvironment();`

halprin (dismissed):
This is looking good, but I think there are some important changes we could make.

app/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamLabOrderSenderTest.groovy
```diff
@@ -171,4 +171,68 @@ class ReportStreamLabOrderSenderTest extends Specification {
...
+    def "retrieveAzureKey works when cache is empty" () {
+        given:
+
+        def mockSecret = Mock(Secrets)
+        def expected = "New Fake Azure Key"
+        mockSecret.getKey(_ as String) >> expected
+        TestApplicationContext.register(Secrets, mockSecret)
+        TestApplicationContext.injectRegisteredImplementations()
+        def rsLabOrderSender = ReportStreamLabOrderSender.getInstance()
+        rsLabOrderSender.cachedAzureKey = null // TODO - azureKeyCache needs to be emptied from prior test
```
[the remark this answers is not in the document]
tjohnson7021: That was supposed to be just a comment, not a TODO, but I removed it either way.

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamLabOrderSender.java
```diff
@@ -34,6 +35,16 @@ public class ReportStreamLabOrderSender implements LabOrderSender {
 
     private static final String CLIENT_NAME = "flexion.etor-service-sender";
 
+    private String cachedAzureKey;
```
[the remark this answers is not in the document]
tjohnson7021: I changed this to `cachedPrivateKey` ... thoughts?

app/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamLabOrderSenderTest.groovy
```diff
@@ -171,4 +171,68 @@ class ReportStreamLabOrderSenderTest extends Specification {
...
+    def "cachedAzureKey synchronization lock is working" () {
+        given:
+        def rsLabOrderSender = ReportStreamLabOrderSender.getInstance()
+        def threadCount = 15
+        def threads = (1..threadCount).collect { index ->
+            new Thread( {
+                rsLabOrderSender.setCachedAzureKey("${index}")
+                def result = rsLabOrderSender.getCachedAzureKey()
+                assert result =="${index}"
+            })
```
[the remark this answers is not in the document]
tjohnson7021: Would it be better to use a third-party framework for this ?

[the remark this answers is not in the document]
basiliskus: Based on our discussion, you could try this. The test passes
```
       def lock = new Object();
        def threads = (1..threadCount).collect { index ->
            new Thread( {
                synchronized (lock) {
                    rsLabOrderSender.setCachedPrivateKey("${index}")
                    def result = rsLabOrderSender.getCachedPrivateKey()
                    assert result =="${index}"
                }
            })
        }
```

basiliskus (commented):
Looks good to me

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamLabOrderSender.java
```diff
@@ -105,7 +110,21 @@ protected String requestToken() throws UnableToSendLabOrderException {
...
+    protected String retrievePrivateKey() throws SecretRetrievalException {
+        var senderPrivateKey =
+                "report-stream-sender-private-key-" + ApplicationContext.getEnvironment();
+        String key = getCachedPrivateKey();
+        if (key != null) {
+            return key;
+        }
+
+        key = secrets.getKey(senderPrivateKey);
+        setCachedPrivateKey(key);
```
JohnNKing, line 122: Initially I was confused why setCachedPrivateKey was being called for each execution. It's not of course -- I just missed the earlier return.

Would it be helpful to consolidate this into a single return statement? For example:

```
String key = getCachedPrivateKey();

if (key == null) {
    key = secrets.getKey(senderPrivateKey);
    setCachedPrivateKey(key);
}

return key;
```

app/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamLabOrderSenderTest.groovy
```diff
@@ -171,4 +171,71 @@ class ReportStreamLabOrderSenderTest extends Specification {
...
+    }
+
+    def "cachedPrivateKey thread synchronization is working" () {
+        given:
+        def rsLabOrderSender = ReportStreamLabOrderSender.getInstance()
+        def threadCount = 15
+        def lock = new Object()
+        def threads = (1..threadCount).collect { index ->
+            new Thread( {
+                synchronized (lock) {
```
JohnNKing, line 195: What is this scenario really intended to test? Won't the key normally be the same value per execution environment? (which may mitigate the impact of race conditions)

Also, shouldn't synchronization be done within rsLabOrderSender to represent a realistic use case?

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamLabOrderSender.java
```diff
@@ -34,6 +35,16 @@ public class ReportStreamLabOrderSender implements LabOrderSender {
 
     private static final String CLIENT_NAME = "flexion.etor-service-sender";
 
+    private String cachedPrivateKey;
+
+    protected synchronized String getCachedPrivateKey() {
+        return cachedPrivateKey;
+    }
+
+    protected synchronized void setCachedPrivateKey(String cachedPrivateKey) {
```
JohnNKing, line 44: Is the intent that setCachedPrivateKey is only ever called once per execution? If so, instead of synchronizing on get and set, we may need a way to synchronize between and including both calls.

jorg3lopez: @JohnNKing you mean to have a lock for both methods?

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamLabOrderSender.java
tjohnson7021: We discussed this syntax and we decided to keep the original because it returns earlier.

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamLabOrderSender.java
JohnNKing: Potentially -- or a single synchronized method that calls set & get (e.g. retrievePrivateKey()). Let's see if I can map out a scenario, as that may help:

Without anything being synchronized:
1. Thread 1: getCachedPrivateKey() called => null returned
2. Thread 2: getCachedPrivateKey() called => null returned
3. Thread 1: setCachedPrivateKey() called => (method still executing...)
4. Thread 2: setCachedPrivateKey() called => this.cachedPrivateKey set to key value and returns
5. Thread 1: this.cachedPrivateKey set to key value and returns

With get and set methods synced, the above can't happen. However, this can:
1. Thread 1: getCachedPrivateKey() called => null returned
2. Thread 2: getCachedPrivateKey() called => null returned
3. Thread 1: setCachedPrivateKey() called => this.cachedPrivateKey set to key value and returns
4. Thread 2: setCachedPrivateKey() called => this.cachedPrivateKey set to key value and returns

Thread 2 is prevented from calling set or get if Thread 1 is doing the same, but what about when Thread 1 is between those calls? (e.g. what if line 121: key = secrets.getKey(senderPrivateKey) takes a long time to execute?)

jorg3lopez (approved):
Great Job, looks good.

PR #283, opened 2023-04-19
Cache - RS Token

jorg3lopez:
# RS Token Caching 

This PR adds the logic for caching the RS token in order to reuse it multiple times when the token has not expired. The Http requests to get an access token will be minimized, optimizing our lab order sending process.

## Issue
#84


app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/demographics/ConvertAndSendLabOrderUsecase.java
```diff
@@ -21,7 +24,9 @@ public static ConvertAndSendLabOrderUsecase getInstance() {
 
     private ConvertAndSendLabOrderUsecase() {}
 
-    public void convertAndSend(Demographics<?> demographics) throws UnableToSendLabOrderException {
+    public void convertAndSend(Demographics<?> demographics)
+            throws UnableToSendLabOrderException, SecretRetrievalException, InvalidKeySpecException,
+                    NoSuchAlgorithmException {
```
jorg3lopez, line 29: Maybe leave only the "UnableToSendLabOrderException" at the method level, then do a try/catch inside the method and throw an UnableToSendLabOrderException whenever any of the other three exceptions get caught, thoughts?

app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/demographics/LabOrderSender.java
```diff
@@ -1,6 +1,12 @@
...
+import gov.hhs.cdc.trustedintermediary.wrappers.SecretRetrievalException;
+import java.security.NoSuchAlgorithmException;
+import java.security.spec.InvalidKeySpecException;
+
 /** Interface for sending a lab order. */
 public interface LabOrderSender {
-    void sendOrder(LabOrder<?> order) throws UnableToSendLabOrderException;
+    void sendOrder(LabOrder<?> order)
+            throws UnableToSendLabOrderException, SecretRetrievalException, InvalidKeySpecException,
+                    NoSuchAlgorithmException;
```
jorg3lopez, line 11: Same thing as the above idea with the exception.

tjohnson7021 (approved):

app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/demographics/ConvertAndSendLabOrderUsecase.java
basiliskus: I'm not familiar with Java best practices but this sounds right to me at a conceptual level. It makes sense to capture a high level UnableToSendLabOrderException in the convertAndSend method. Unless there's any debugging data that would get lost from the other exceptions. In that case, we could make sure to log the information before throwing UnableToSendLabOrderException

app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/demographics/LabOrderSender.java
basiliskus: Same as above

basiliskus (commented):
Responded to your comments, otherwise looks good to me!

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamLabOrderSender.java
```diff
@@ -63,10 +75,33 @@ public void sendOrder(final LabOrder<?> order) throws UnableToSendLabOrderExcept
         logger.logInfo("Sending the order to ReportStream at {}", RS_DOMAIN_NAME);
 
         String json = fhir.encodeResourceToJson(order.getUnderlyingOrder());
-        String bearerToken = requestToken();
+        String bearerToken;
+        bearerToken = getRsToken();
```
basiliskus, line 79: Is there a reason to have the instantiation in a separate line?

basiliskus (approved):
Looks good! Just left one comment

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamLabOrderSender.java
jorg3lopez: changed.

PR #299, opened 2023-05-04
Hot Fix for Staging

jorg3lopez:
# Hot Fix for getExpirationDate

This PR adds a try catch to the getExpirationDate method. This was necessary as the JWTS library throws an exception when trying to read a token that has expired. Because this method returns a date instance, we needed to return an expiration date of now() - 10 minutes. This way, it fails the validation process of the token.


app/src/main/java/gov/hhs/cdc/trustedintermediary/external/jjwt/JjwtEngine.java
```diff
@@ -91,9 +93,16 @@ protected RSAPrivateKey readPrivateKey(@Nonnull String pemKey)
     public LocalDateTime getExpirationDate(String jwt) {
 
         var tokenOnly = jwt.substring(0, jwt.lastIndexOf('.') + 1);
-        Claims claims = Jwts.parserBuilder().build().parseClaimsJwt(tokenOnly).getBody();
-        Date expirationDate = claims.getExpiration();
-
-        return LocalDateTime.ofInstant(expirationDate.toInstant(), ZoneId.systemDefault());
+        try {
```
basiliskus, line 96: Is the exception only thrown once when the token expires or is there a chance the exception is thrown multiple times consecutively until we get a new token?

jorg3lopez: Once the token expires, the method isValidToken returns false. Then getRsToken calls requestToken and a new token is saved in the cache.

Issue #148, opened 2023-01-31
Receiving FHIR Demographics via REST


JohnNKing:
# Story
As a hospital, so that my NBS order demographic data can be transmitted to a public health lab, I need the receiver to be able to accept FHIR data via REST.

## Pre-conditions
- Only a single, fake hospital is supported.
- The received data is assumed to be correctly formatted and valid in terms of required fields and data types.

## Acceptance Criteria

- [x] Demographic data must be transmitted securely in non-development environments.
- [x] The hospital receives confirmation that the newborn demographic data was successfully received (by the CDC; not the lab).
- [x] If the hospital user provides valid login information, they are allowed to submit newborn demographic data.
- [x] If the hospital user does not provide valid login information, any attempt to submit newborn demographic data results in an access denied error.

### Research
~~- [ ] Investigate with the ReportStream team possibilities for creating a shared Auth service~~

###  Engineering
- [x] Get ReportStream public key
- [x] Caching
    - [x] Create caching class @jorg3lopez @tjohnson7021 
    - [x] Refactor ReportStreamLabOrderSender to use caching class @jorg3lopez 
- [x] Onboard RS client using YAML file @basiliskus @jorg3lopez @JohnNKing 
   - [x] Add implementation to read YAML file into object @basiliskus 
   - [x] Create Organization and OrganizationsLoader class @basiliskus 
   - [x] Validate sender id exists in organizations.yaml @basiliskus
- [x] Add the authentication endpoint the ReportStream will call [ POST /v1/auth ]
   - [x] Determine our bearer token format (JWT?), and whether ReportStream supports short term JWTs. @jorg3lopez 
   - [x] Create endpoint @JohnNKing 
   - [x] Check if user and credentials are valid @JohnNKing 
   - [x] Generate session token @JohnNKing 
   - [x] Get keys (e.g. RS public, TI private) from Azure Key Vault and/or cache @jorg3lopez, @halprin 
   - [x] Change how we load organizations. - @halprin 
   - [x] Harden the AuthController's parsing of the request body. - @halprin
   - [x] Enhance `LocalSecrets` to also read from the Java resources if the local file doesn't work. - @halprin 
   - [x] AuthCotroller needs to return the JWT in a JSON blob similar to how RS does it. @jorg3lopez 
   - [x] Unit tests for exceptions, POJOs, and the usecase - @halprin, @JohnNKing 
   - [x] JavaDocs - @halprin 
[...]

PR #328, opened 2023-05-18
Add initial Report Stream client entry

basiliskus:
# Add organizations YAML

- Add initial Report Stream client entry in YAML file

## Issue

#148

### DIFF
--- a/app/src/main/resources/organizations.yaml
+++ b/app/src/main/resources/organizations.yaml
@@ -1,0 +1,8 @@
+- name: report-stream
+  description: Report Stream
+  senders:
+    - name: default
+      topic: etor
+      customerStatus: active
+      format: FHIR
+  receivers:
--- a/app/src/main/resources/organizations.yaml
+++ b/app/src/main/resources/organizations.yaml
@@ -1,8 +1,2 @@
 - name: report-stream
   description: Report Stream
-  senders:
-    - name: default
-      topic: etor
-      customerStatus: active
-      format: FHIR
-  receivers:
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -32,6 +32,7 @@ jobs:
     with:
       ENVIRONMENT: staging
       REPO: trusted-intermediary-router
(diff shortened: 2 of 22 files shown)


app/src/main/resources/organizations.yaml
```diff
@@ -0,0 +1,8 @@
+- name: report-stream
+  description: Report Stream
+  senders:
+    - name: default
+      topic: etor
+      customerStatus: active
+      format: FHIR
+  receivers:
```
halprin, line 8: Will we have a concept of receivers that are tied to a sender?  Maybe in the future?  I'm out of the loop behind what was the thinking in this PR, so... ?

app/src/main/resources/organizations.yaml
```diff
@@ -0,0 +1,8 @@
+- name: report-stream
+  description: Report Stream
+  senders:
+    - name: default
+      topic: etor
+      customerStatus: active
+      format: FHIR
```
halprin, line 7: I don't think we need to specify format because this TI application will only ever accept FHIR.  I'm not saying we wouldn't ever accept another format, but if we were, some other service (like ReportStream, for example) would convert it first to FHIR and then send it to this TI application.

app/src/main/resources/organizations.yaml
basiliskus: Thanks for reviewing. I'm not sure if we'll need receivers. We should discuss it with @jorg3lopez. We're mimicking the structure of the organizations.yaml file used by Report Stream, so this is open to discussion

app/src/main/resources/organizations.yaml
basiliskus: Fair. I'm ok with removing it

app/src/main/resources/organizations.yaml
halprin: Sure, we can discuss.  I think my overall thought is we should make the setting file our own.  If there's a concept that we don't need for our application, let's not have it.  We can always add concepts back if we ever actually support something.

PR #330, opened 2023-05-22
Auth Endpoint

JohnNKing:
# Auth Endpoint

Adding an auth service to validate the ReportStream JWT and then issue session JWTs.

## Issue

#148


app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthDomainRegistration.java
```diff
@@ -0,0 +1,78 @@
...
+            return authController.constructResponse(400);
+        }
+
+        var token = "";
+        var payload = "";
+
+        try {
+            token = requestSessionTokenUsecase.getToken(authRequest);
+        } catch (InvalidTokenException | UnknownOrganizationException e) {
+            logger.logInfo("Authentication failed", e);
```
basiliskus, line 61: I think it'd be better to use logError or logDebug when logging exceptions since it's not expected behavior

app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthDomainRegistration.java
```diff
@@ -0,0 +1,78 @@
...
+            return authController.constructResponse(401);
+
+        } catch (Exception e) {
+            logger.logFatal("Bad authentication service configuration: Authentication failed", e);
```
basiliskus, line 65: I think logError is more appropriate here since the control flow is not interrupted

app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthDomainRegistration.java
```diff
@@ -0,0 +1,78 @@
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    DomainResponse handleAuth(DomainRequest request) {
+        AuthRequest authRequest;
+        try {
+            authRequest = authController.parseAuthRequest(request);
+        } catch (Exception e) {
+            return authController.constructResponse(400);
```
basiliskus, line 52: Could be useful to have logging here

app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthDomainRegistration.java
```diff
@@ -0,0 +1,78 @@
...
+
+        } catch (Exception e) {
+            logger.logFatal("Bad authentication service configuration: Authentication failed", e);
+            return authController.constructResponse(500);
+        }
+
+        try {
+            payload = authController.constructPayload(authRequest, token);
+        } catch (Exception e) {
+            logger.logFatal("Internal formatter exception occurred", e);
```
basiliskus, line 72: Same as above

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/localfile/LocalSecrets.java
```diff
@@ -27,17 +29,37 @@ public static LocalSecrets getInstance() {
...
 
         try {
-            key = Files.readString(Path.of("..", "mock_credentials", secretName + ".pem"));
-        } catch (IOException e) {
-            var message = "Error getting local key";
-            logger.logError(message, e);
-            throw new SecretRetrievalException(message, e);
+            key = readSecretFromFileSystem(secretName);
+        } catch (SecretRetrievalException exception) {
+            logger.logInfo(
```
basiliskus, line 38: I think logError would be more appropiate here

basiliskus (commented):
Left some comments regarding logging

app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthDomainRegistration.java
halprin: Great catch!  I added it.

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/localfile/LocalSecrets.java
halprin: I decided to do the info level because this may not be an error and completely OK because the secret may be in the resources instead on the filesystem.  Maybe this should be at the warn level?

app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthDomainRegistration.java
halprin: You think so?  This is indicative that we, the engineering team, incorrectly configured or coded something in our application.  When do you think a fatal level should be used?

app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthDomainRegistration.java
halprin: Good catch.  Fixed.

app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthDomainRegistration.java
basiliskus: I think it should be used in exceptional cases when the application becomes unstable due to a critical issue that doesn't allow it to continue working properly. For context, the logging library we're using `slf4j` doesn't even include a FATAL level so I had to add it. I thought it could be useful to add in cases where the app needs urgent attention (e.g. in BEDAP, the person on-call would get paged if a fatal error was logged). Here's some more info: https://betterstack.com/community/guides/logging/log-levels-explained/#fatal

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/localfile/LocalSecrets.java
basiliskus: Ok, that's fair. I'd vote to log it as warning or debug as I see INFO more as log for normal/expected flow, but I don't mind if left as is

app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthDomainRegistration.java
basiliskus: On second thought, you're right. This would be a critical error that needs our immediate attention. Sorry,  I didn't pay close enough attention

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/localfile/LocalSecrets.java
halprin: I changed it to a warning.

app/src/main/java/gov/hhs/cdc/trustedintermediary/organizations/OrganizationConfigException.java
```diff
@@ -0,0 +1,8 @@
+package gov.hhs.cdc.trustedintermediary.organizations;
+
+/** Exception thrown when the organizations settings configuration cannot be loaded. */
+public class OrganizationConfigException extends Exception {
```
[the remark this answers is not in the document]
halprin: That's a good idea!  Which way do you prefer?  And if you want, feel free to make the change yourself.  If not, I'm happy to make it too.

app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthDomainRegistration.java
basiliskus: Sounds good!

app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthController.java
```diff
@@ -0,0 +1,99 @@
...
+
+        String payloadJson;
+        if (token == null || token.isBlank()) {
+            return token;
+        }
+        String scope = authRequest.scope();
+        String token_type = "bearer";
+        Map<String, String> payload = new HashMap<>();
+
+        payload.put("token_type", token_type);
```
halprin, line 87: We don't really need this variable.  We can just hardcode `"bearer"` into the `put` statement.
```suggestion
        Map<String, String> payload = new HashMap<>();

        payload.put("token_type", "bearer");
```

app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthController.java
```diff
@@ -0,0 +1,99 @@
...
+                                    }
+                                }));
+    }
+
+    protected String constructPayload(AuthRequest authRequest, String token) {
+
+        String payloadJson;
+        if (token == null || token.isBlank()) {
+            return token;
+        }
```
halprin, line 82: I don't think we should return a blank response or a `null` response if the `token` is blank or `null`.  Instead, execution should continue and this blank or `null` token should be put into the larger payload regardless.

app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthController.java
```diff
@@ -0,0 +1,99 @@
...
+
+        payload.put("token_type", token_type);
+        payload.put("access_token", token);
+        payload.put("scope", scope);
+
+        try {
+            payloadJson = formatter.convertToJsonString(payload);
+        } catch (FormatterProcessingException e) {
+            throw new RuntimeException(e);
+        }
```
halprin, line 95: I think we can just let this method `throws FormatterProcessingException` so we don't need to surround the `formatter` call with a `try`/`catch`

app/src/main/java/gov/hhs/cdc/trustedintermediary/organizations/OrganizationConfigException.java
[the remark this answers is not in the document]
halprin: There isn't really a naming convention for this in our codebase or Java (that I know of).

app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthController.java
jorg3lopez: Added it to the method signature.

app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthController.java
jorg3lopez: Agree.

app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthController.java
jorg3lopez: done.

app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthDomainRegistration.java
jorg3lopez: added throws to the method signature so the try catch is not there anymore.

app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthController.java
```diff
@@ -0,0 +1,98 @@
...
+        return response;
+    }
+
+    public String constructPayload(AuthRequest authRequest, String token)
+            throws FormatterProcessingException {
+
+        String payloadJson;
+        String bearerToken = "";
+
+        if (token != null && !token.isBlank()) {
```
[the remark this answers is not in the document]
halprin: We could.  I'm not the biggest fan of ternary operators, but whatever.  I think we could also just do away with the entire if statement and put `token` directly into the `payload` map.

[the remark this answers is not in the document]
jorg3lopez: There are two possible conditions to check for. 
1. Is the string null
2. If is is not null, is the the string blanked with whitespaces

If we want to use a ternary, then it would be something like:

`String bearerToken = (token != null) ? (!token.isBlank() ? token: "") : ""` 
or
`String bearerToken = (token == null) ? "" : (token.isBlank() ? "" : token)`

One thing to consider is, should we be checking for blank tokens? thoughts?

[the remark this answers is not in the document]
jorg3lopez: We decided on the nice ternary. Good call Basilio!

JohnNKing (commented):
Things are looking great to me! I'd approve, but I own the PR. 🙃

halprin (approved):
Let 'er rip!  What do others say?

basiliskus (approved):
Great work!

PR #332, opened 2023-05-22
Caching Class

jorg3lopez:
# Caching Class

This PR adds a caching class that is responsible for caching keys from Azure. We were face with the issue that we needed to create an Auth endpoint. This Auth endpoint will retrieve keys from Azure to sign and validate JWTs. Adding caching will optimize the performance in general.

The caching class was created in order to combine the logic into a single class that will be used when sending/receiving a payload, and when generating/validating JWTs. 

## Issue
#148

### DIFF
--- a/app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/KeyCache.java
+++ b/app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/KeyCache.java
@@ -1,14 +1,20 @@
 package gov.hhs.cdc.trustedintermediary.etor;
 
-import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
+/** Stores keys from Azure to optimize performance. */
 public class KeyCache {
 
     private Map<String, String> keys;
 
     private KeyCache() {
-        keys = new HashMap<>();
+        // ConcurrentHashMap<>() over Collections.synchronizedMap() due to performance. Synchronized
+        // map, locks the
+        // whole object when reading or writing. Concurrent hashmap locks happen at the bucket
+        // level, leaving the read
+        // function unlocked when writing.
+        keys = new ConcurrentHashMap<>();
     }
 
     public static KeyCache getInstance() {
--- a/app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/KeyCache.java
+++ b/app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/KeyCache.java
@@ -9,10 +9,9 @@ public class KeyCache {
     private Map<String, String> keys;
 
     private KeyCache() {
-        // ConcurrentHashMap<>() over Collections.synchronizedMap() due to performance. Synchronized
-        // map, locks the
-        // whole object when reading or writing. Concurrent hashmap locks happen at the bucket
-        // level, leaving the read
+        // ConcurrentHashMap<>() over Collections.synchronizedMap() due to performance.
+        // Synchronized map locks the whole object when reading or writing.
+        // Concurrent hashmap locks happen at the bucket level, leaving the read
         // function unlocked when writing.
         keys = new ConcurrentHashMap<>();
     }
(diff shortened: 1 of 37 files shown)


app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/KeyCache.java
```diff
@@ -0,0 +1,31 @@
...
+public class KeyCache {
+
+    private Map<String, String> keys;
+
+    private KeyCache() {
+        // ConcurrentHashMap<>() over Collections.synchronizedMap() due to performance. Synchronized
+        // map, locks the
+        // whole object when reading or writing. Concurrent hashmap locks happen at the bucket
+        // level, leaving the read
+        // function unlocked when writing.
```
jorg3lopez, line 16: I think the linter formatted the comment in this way. I can see if it will accept making it more symmetrical, thoughts?

jorg3lopez: Looks better now

halprin (changes requested):
Let's make this more generic.  Not just for keys.  Because of this, I would make an interface and "promote" this to be usable across all plugins (which means moving these files outside the `etor` package and register them in `App.java`).

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/in_memory/KeyCache.java
```diff
@@ -0,0 +1,33 @@
+package gov.hhs.cdc.trustedintermediary.external.in_memory;
```
halprin, line 1: I'd make the package one word `inmemory`.  `_` isn't really used for package names.

app/src/main/java/gov/hhs/cdc/trustedintermediary/wrappers/Cache.java
```diff
@@ -0,0 +1,8 @@
+package gov.hhs.cdc.trustedintermediary.wrappers;
+
+public interface Cache {
```
halprin, line 3: Let's add some JavaDocs.

app/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/in_memory/keyCacheTest.groovy
```diff
@@ -0,0 +1,46 @@
+package gov.hhs.cdc.trustedintermediary.external.in_memory
+
+import gov.hhs.cdc.trustedintermediary.external.in_memory.KeyCache
+import spock.lang.Specification
+
+class keyCacheTest extends Specification {
```
halprin, line 6: Rename this file (and this class) to have a beginning capital letter.

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/in_memory/KeyCache.java
```diff
@@ -0,0 +1,33 @@
...
+
+import gov.hhs.cdc.trustedintermediary.wrappers.Cache;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/** Stores keys from Azure to optimize performance. */
```
halprin, line 7: Consider rewriting this JavaDocs to be more generic.

app/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/in_memory/keyCacheTest.groovy
jorg3lopez: Good catch!

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/in_memory/KeyCache.java
jorg3lopez: Agree.

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/in_memory/KeyCache.java
jorg3lopez: done

basiliskus (commented):
LGTM

halprin (approved):

PR #333, opened 2023-05-22
Add implementation to read YAML file into object

basiliskus:
# Add implementation to read YAML file into object

- Refactored to rename Formatter.convertToObject to convertJsonToObject, added implementation for convertYamlToObject, and abstracted convertToObject

## Issue

#148

### DIFF
--- a/app/src/main/java/gov/hhs/cdc/trustedintermediary/wrappers/Formatter.java
+++ b/app/src/main/java/gov/hhs/cdc/trustedintermediary/wrappers/Formatter.java
@@ -6,7 +6,9 @@ package gov.hhs.cdc.trustedintermediary.wrappers;
  */
 public interface Formatter {
 
-    <T> T convertToObject(String input, Class<T> clazz) throws FormatterProcessingException;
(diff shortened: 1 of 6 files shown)


app/src/main/java/gov/hhs/cdc/trustedintermediary/wrappers/Formatter.java
```diff
@@ -6,7 +6,9 @@
  */
 public interface Formatter {
 
-    <T> T convertToObject(String input, Class<T> clazz) throws FormatterProcessingException;
+    <T> T convertJsonToObject(String input, Class<T> clazz) throws FormatterProcessingException;
+
+    <T> T convertYamlToObject(String input, Class<T> clazz) throws FormatterProcessingException;
 
     String convertToString(Object obj) throws FormatterProcessingException;
```
halprin, line 13: Would you mind renaming this method to `convertToJsonString` since we differentiate between JSON and YAML now.

halprin (commented):
Just one quick comment, but great change!

halprin (approved):

PR #347, opened 2023-05-25
Clean-up JWT Expiration Logic

halprin:
# Clean-up JWT Expiration Logic

If an expired JWT was passed in to our JWT expiration time logic, we just returned a time 10 minutes ago regardless of when the JWT actually expired.  We now return the actual expiration time of the JWT.  The tests were also modified.

## Issue

_None._


jorg3lopez (approved):
Looks good, great job!

Issue #315, opened 2023-05-16
Documentation Automation Improvements


sfradkin:
# Backlog Task

Additional improvements for working on and automating documentation

## Completion Criteria

- [ ] A consistent URL is available to access the documentation on the Flexion staging environment
- [ ] A solution exists to allow for any developer to easily update and test documentation generation without having to install more tooling on their computer

## Tasks

- [ ] What tasks need to be done to fulfill the completion criteria?

## Other Notes

- Any other notes to help clarify this task for the team

PR #351, opened 2023-05-30
Partial implementation for issue #315

sfradkin:
adding a new dockerfile that can be built and run with all the tooling to generate documentation, update readme file to explain how to use new dockerfile to do two-way development and generation of documentation, added a small bash script that fixes the execute permissions of scripts after running the publisher update

## Issue

Partial implementation for issue #315


ig/README.md
```diff
@@ -20,10 +20,11 @@ To be written
 1. run `./_updatePublisher.sh` in order to ensure you are using the most recent publisher release
 1. run `./_genonce.sh` to generate the Implementation Guide
 1. In a browser you can open up the `ig/output/index.html` file that was generated as this is the home page for the Implementation Guide
+2. Alternately, you may opt to use the local development Docker approach outlined below to avoid having to install the toolchain listed above.
```
halprin, line 23: Mind making this number a `1` to match?  Either that, or make them all ascend.
```suggestion
1. Alternately, you may opt to use the local development Docker approach outlined below to avoid having to install the toolchain listed above.
```

ig/README.md
```diff
@@ -32,3 +33,24 @@ A Dockerfile is provided that can be used to generate the Implementation Guide.
...
+
+## Using Docker for local development
+
+Provided is a Dockerfile suitable for running the implementation guide documentation generation without the need to install all the tooling locally.
+The Docker image needs to be built and then run with an appropriate `mount` flag in order to mount your local `ig` directory into the running Docker container.
+This will allow bi-directional writes so that the Docker container can generate the documentation, and as a developer you can view the generated documentation easily.
+A developer will also be able to update any input files for the documentation and re-run the documentation generation without any intermediate steps.
+
+1. In a terminal, navigate to this `ig` directory
+2. run `docker buildx build -t ig-local-dev -f Dockerfile-local-dev .` to build the Docker container
```
halprin, line 45: Is `buildx` really needed?  You can just do `build`?

ig/README.md
```diff
@@ -32,3 +33,24 @@ A Dockerfile is provided that can be used to generate the Implementation Guide.
...
+3. run `docker run -it --mount type=bind,source="$(pwd)",target=/trusted-intermediary ig-local-dev bash` to run the container and use an interactive `bash` shell inside the container
```
halprin, line 46: Will the shorthand form work too?

`docker run -it -v "$(pwd)":/trusted-intermediary ig-local-dev bash`

ig/README.md
```diff
@@ -32,3 +33,24 @@ A Dockerfile is provided that can be used to generate the Implementation Guide.
...
+4. Inside the container environment it should default to a working directory of `/trusted-intermediary`; this is bound to the `ig` directory in your local git repo
+5. Periodically update the IG Publisher by running `./_updatePublisher.sh`
```
halprin, line 48: Periodically?  Is there a better set of rules for when and why we would need to run this?

halprin (commented):
Nice!  I've got a few questions, though.

ig/README.md
sfradkin: Ah, I'll fix this.  I believe the markdown validator wants this to be fixed as well

ig/README.md
sfradkin: From my read of the Docker documentation, `buildx` is needed for the copy with permissions that's used in `Dockerfile`.  In theory `build` should work fine for `Dockerfile-local-dev`.  I can test it out.

ig/README.md
sfradkin: According to the documentation in order to do a two way bind type of volume mount you need to use the `--mount` flag and not the `-v` flag.

ig/README.md
sfradkin: It seems like the publisher is updated multiple times per week.  Maybe a better solution is to include a trigger that runs it automatically after the container is started and then again once per day?

ig/README.md
halprin: I did a quick test, and `-v` seems to reflect any changes in the Docker container's filesystem in the hosts' file system.  I looked at some [documentation](https://docs.docker.com/storage/bind-mounts/), and Docker recommends using the new `--mount` argument, so I'm fine with whatever.

ig/README.md
sfradkin: README will be updated with more details, automating this will be deferred to a different backlog task

PR #359, opened 2023-06-02
Cache Synchronization Fails Occasionally

jorg3lopez:
# rsTokenCache Synchronization Fails Occasionally

The synchronization test for the rsTokenCache tends to fail occasionally. Since the cache is only a string variable, we needed a way to verify that each thread was firing to the last iteration. A hashmap was created to house the thread number as the key and the iteration as the value. The issue with this was that the hashmap is not thread-safe. 

In order to address this issue, the hashmap was changed to a concurrent hashmap.

## Issue
Bug hot-fix

## Checklist


basiliskus (approved):
LGTM

PR #360, opened 2023-06-02
OpenAPI updates for demographics endpoint

ghost:
#OpenAPI updates for demographics endpoint
- added _Unauthorized Request_ response to schema.
- add specifications to indicate demographics endpoint requires an API key

## Checklist


tjohnson7021:
Looking for any and all input before I fully complete this story - see changes.

app/src/main/resources/openapi_etor.yaml
```diff
@@ -40,3 +55,15 @@ components:
         patientId:
           type: string
           example: MRN7465737865
+    UnauthorizedError:
+      description: API key is missing or invalid
```
basiliskus, line 59: I'm thinking it's probably better to have a generic description here for the error schema, and a more specific one in under the responses. The description here could be `Unauthorized Request`, and the one above could be `Authentication failed due to invalid token or unknown organization`?

app/src/main/resources/openapi_etor.yaml
```diff
@@ -40,3 +52,14 @@ components:
...
+      description: Unauthorized Request
+    InternalError:
+      description: Internal Server Error
+  securitySchemes:
+    ApiKeyAuth:
+      type: apiKey
+      in: header
+      name: OUR-API-KEY
+security:
+  - ApiKeyAuth: []
```
halprin, line 65: I wouldn't apply this auth to everything.  Let's apply it on the operation level.  I say this because while everything in this file, thus far, will need to be protected, this file will be combined with other plugin's OpenAPI specification which may have endpoints that don't use this API key auth.

halprin (commented):
This is great!  Just some small tweaks.

app/src/main/resources/openapi_etor.yaml
```diff
@@ -40,3 +49,10 @@ components:
         patientId:
           type: string
           example: MRN7465737865
+    UnauthorizedError:
+      description: Unauthorized Request
+  securitySchemes:
+    ApiKeyRequired:
+      type: apiKey
+      in: header
+      name: Authorization
```
halprin, line 58: I read some of the other documentation for OpenAPI.  I would use a [bearer authentication](https://swagger.io/docs/specification/authentication/bearer-authentication/) instead.

halprin (approved):
Woot woot!

PR #366, opened 2023-06-08
adding some hl7v2 examples

sfradkin:
# Adding HL7v2 examples from MN

adding some examples from the state of minnesota for ADT, OML, and ORU


halprin (approved):
Most excellent, dude!

PR #371, opened 2023-06-09
/v1/auth endpoint load tests (DoD for #148)

basiliskus:
# /v1/auth endpoint load tests (DoD for #148)

## Issue

#148 

## Checklist

- [x] I have added code documentation where needed
- [x] I have updated the documentation accordingly (@basiliskus: I checked the main readme and it seems we don't need to update anything there specific to new API endpoints)

### DIFF
--- a/operations/locustfile.py
+++ b/operations/locustfile.py
@@ -2,18 +2,28 @@ from locust import HttpUser, task, events
 from locust.runners import MasterRunner
 
 
-request_body = None
+demographics_request_body = None
+auth_request_body = None
 
 
 @events.test_start.add_listener
 def test_start(environment):
-    global request_body
+    global demographics_request_body
+    global auth_request_body
     if isinstance(environment.runner, MasterRunner):
         # in a distributed run, the master does not typically need any test data
         return
+
     # read the sample request body for the demographics endpoint
-    with open("e2e/src/test/resources/newborn_patient.json", "r") as request_body_file:
-        request_body = request_body_file.read()
+    with open("e2e/src/test/resources/newborn_patient.json", "r") as f:
+        demographics_request_body = f.read()
+
+    # set up the sample request body for the auth endpoint
+    # using a valid test token found in the mock_credentials directory
+    auth_scope = "report-stream"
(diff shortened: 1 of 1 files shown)


operations/locustfile.py
```diff
@@ -27,8 +37,11 @@ def get_health(self):
 
     @task(5)  # this task will get called 5x more than the other
     def post_v1_etor_demographics(self):
-        self.client.post("/v1/etor/demographics", data=request_body)
+        self.client.post("/v1/etor/demographics", data=demographics_request_body)
 
+    @task(5)  # this task will get called 5x more than the other
```
basiliskus, line 42: Not sure if we need to call this test 5x. Realistically, I'm guessing it would be called a lot less than the demographics endpoint?

tjohnson7021: I updated this to 1 (removed the 5) based on what @basiliskus and I discussed.

operations/locustfile.py
```diff
@@ -2,18 +2,28 @@
...
-        request_body = request_body_file.read()
+    with open("e2e/src/test/resources/newborn_patient.json", "r") as f:
+        demographics_request_body = f.read()
+
+    # set up the sample request body for the auth endpoint
+    # using a valid test token found in the mock_credentials directory
+    auth_scope = "report-stream"
+    with open("mock_credentials/report-stream-valid-token.jwt") as f:
+        auth_token = f.read()
+    auth_request_body = {"scope": auth_scope, "client_assertion": auth_token}
```
halprin, line 26: This is in relation to @tjohnson7021 [question](https://github.com/CDCgov/trusted-intermediary/pull/371#issuecomment-1584951995) as to whether we should still be seeing the `ERROR Authentication failed` log.  We're seeing that still because the auth endpoint doesn't take JSON.  It takes a `x-www-form-urlencoded` content type instead.  E.g....

```suggestion
    auth_request_body = f"scope={auth_scope}&client_assertion={auth_token}"
```

There may be a more native way with Locust to send `x-www-form-urlencoded` content than my suggestion, though.

halprin (changes requested):

halprin (approved):
Looks awesome!!!

PR #374, opened 2023-06-09
Rename TI Private Key

halprin:
# Rename TI Private Key

Renamed our private key to be `trusted-intermediary-private-key-*` instead of `report-stream-sender-private-key-*`.

In addition to this, I also set for keys to only be soft-deleted by Terraform when it destroys the keys.

## Issue

_None_.

## Checklist

- [x] Look into `purge_soft_deleted_secrets_on_destroy` and setting it to `false` in the Terraform.


tjohnson7021 (approved):
I reviewed for the name change and everything looks good to me. I will save my follow up questions for later.

basiliskus (approved):
Looks good to me!

PR #385, opened 2023-06-15
Updated python URL encoding

ghost:
# Updated python URL encoding

- implementing a python method for url encoding

### DIFF
--- a/operations/locustfile.py
+++ b/operations/locustfile.py
@@ -1,5 +1,6 @@
 from locust import HttpUser, task, events
 from locust.runners import MasterRunner
+import urllib.parse
 
 
 demographics_request_body = None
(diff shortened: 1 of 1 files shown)


operations/locustfile.py
```diff
@@ -1,5 +1,6 @@
 from locust import HttpUser, task, events
 from locust.runners import MasterRunner
+import urllib.parse
```
basiliskus, line 3: This is nitpicking, and no need to change it in this PR as we don't have much python code in our repo that we're maintaining and I don't think is necessary. So just as FYI. According to PEP8 (Python's style guide):
> Imports should be grouped in the following order:
>     Standard library imports.
>     Related third party imports.
>     Local application/library specific imports.

So, `import urllib.parse` should come first as it's a standard library; and `locust` is a third-party

basiliskus (approved):
Good job!

halprin (commented):
Sorry I'm late to the party, but great job, @tjohnson7021!

PR #387, opened 2023-06-15
Update locust demographics endpoint test to authenticate with valid token

basiliskus:
- Changed demographics test implementation to authenticate with a valid token
- Extracted endpoint variables
- Ran python's `black` for formatting

## Issue

#148 

## Checklist

- [x] I have added docs where required


jorg3lopez (approved):
LGTM, nice job.

halprin (approved):
Great changes!

basiliskus:
> I started thinking on how we could handle load tests that last longer than 5 minutes (that's how long until our JWT expires). So, in a future PR, we could make it so that we redo the `urllib.request.Request` call every so often. _Or_, and I haven't entirely thought this through, but we could perhaps make a long lived JWT signed by our **local** TI private key so we never need to call the auth endpoint to renew the JWT. _But_ that may limit our options since we would be stuck only load testing against the local environment.

Good point. I'll add this to our tech debt channel as a reminder. I also think it'd be a good idea to have the load test run in staging as well

PR #392, opened 2023-06-19
Block Storage Account Public Access

JohnNKing:
# Block Storage Account Public Access

Prevent public access to API docs. This is likely a good pattern to follow for future storage accounts. This does not appear to limit access via the static website setting.

## Issue

#148

## Checklist

n/a


sfradkin (approved):
This seems like the right way to do this.  Alternatively we could use network rules, but this does the trick.

halprin (approved):

basiliskus (approved):

PR #393, opened 2023-06-19
Require HTTPS for API Endpoints

JohnNKing:
# Require HTTPS for API Endpoints

Force redirects for anyone attempting to make an unencrypted HTTP call against our web service.

This currently results in an HTTP 301 redirect to the HTTPS version of the site for any request. However, a more ideal solution might be to fail on requests (e.g. via 400, at least for POSTs), to ensure clients that trying send data unencrypted and which automatically follow links to not work. Unfortunately, Azure does not seem to provide that out of the box. In practice, however, POSTs resulting in a 301 are often converted to GET lacking body data, which would result in a failure.

## Issue

#148.


sfradkin (approved):
This is fine.  We should create a backlog item to investigate the issues pointed out in the PR comment.

halprin (approved):
Even if we send a `400` back, the damage has been done: the information has been sent unencrypted across the Internet.  In fact, it wouldn't matter if we somehow made the web service not respond to HTTP.  The data is still sent unencrypted to the remote server.

basiliskus (approved):

JohnNKing:
Created #395 and documented in our threat model.

PR #397, opened 2023-06-20
Ignore output for api server to reduce noise in load test output

basiliskus:
Muting output for api server so we can read the load test in output in console when running `load-execute.sh`


JohnNKing (approved):
I just tried the same modification and it works for me!

halprin (approved):

jorg3lopez (approved):
Nice!

PR #398, opened 2023-06-20
Updating DAST openAPI path

JohnNKing:
# Updating DAST openAPI path

Correcting the OpenAPI path used by ZAP to perform DAST scans.

## Checklist

n/a

### DIFF
--- a/.github/workflows/dast.yml
+++ b/.github/workflows/dast.yml
@@ -52,6 +52,7 @@ jobs:
 
       - name: Upload Report
         uses: actions/upload-artifact@v3
+        if: always()
         with:
           name: zap-scan-report
           path: api-scan-report.html
--- a/.github/workflows/dast.yml
+++ b/.github/workflows/dast.yml
@@ -40,15 +40,14 @@ jobs:
       - name: Check API Health
         run: ./health-check.sh
 
-      - name: Pull ZAP OWASP Image
-        run: docker pull owasp/zap2docker-weekly
-
-      - name: Create docker network
-        run: docker network create zapnet
+#      - name: Pull ZAP OWASP Image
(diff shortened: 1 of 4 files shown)


app/src/main/java/gov/hhs/cdc/trustedintermediary/external/javalin/App.java
```diff
@@ -31,6 +31,10 @@ public class App {
     public static void main(String[] args) {
         var app = Javalin.create().start(8080);
 
+        // apply this security header to all responses, but allow it to be overwritten by a specific
+        // endpoint by using `before` if needed
+        app.before(ctx -> ctx.header("X-Content-Type-Options", "nosniff"));
```
JohnNKing, line 36: @halprin Ah! That's much cleaner.

jorg3lopez (approved):
Great work! I left a comment as well.

basiliskus (approved):

.github/workflows/dast.yml
```diff
@@ -40,18 +40,12 @@ jobs:
       - name: Check API Health
         run: ./health-check.sh
 
-      - name: Pull ZAP OWASP Image
-        run: docker pull owasp/zap2docker-weekly
-
-      - name: Create docker network
-        run: docker network create zapnet
```
jorg3lopez, line 47: Just wondering why the net was taken out.

halprin: Ah, yeah.  If the network was going to be used, the TI service also needed to run inside the network.  It wasn't.  So the network wasn't buying us anything.  Instead the TI service is being exposed to the host network which is why we do the entire `ip -f inet -o addr show docker0 | awk '{print $4}' | cut -d '/' -f 1` thing to get the host IP address.

PR #400, opened 2023-06-21
Extract API endpoint paths

basiliskus:
Extracted API endpoint paths to public constants to avoid hardcoding


halprin (changes requested):
I like how you're thinking here, but we should keep our plugins separate from each other and separate from the main application.  By putting all the endpoints in a single properties file, we are violating that concept.

A couple of thoughts...

1. We could make multiple properties files, one for each plugin.  Does that still provide value as you see it?
2. We could change to a constant class.  I.e. a class that has nothing but `public static final String` fields.  We would still want to have one per plugin though.  E.g. `EtorEndpoints.DEMOGRAPHICS`.

What do you think about all this?

app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthDomainRegistration.java
```diff
@@ -20,12 +20,14 @@
  */
 public class AuthDomainRegistration implements DomainConnector {
 
+    public static final String AUTH_API_ENDPOINT = "/v1/auth";
```
halprin, line 23: You can make these "package private" which will allow them to be used in tests but not outside the tests.

```suggestion
    static final String AUTH_API_ENDPOINT = "/v1/auth";
```

app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
```diff
@@ -30,15 +30,18 @@
  */
 public class EtorDomainRegistration implements DomainConnector {
 
+    public static final String DEMOGRAPHICS_API_ENDPOINT = "/v1/etor/demographics";
+    public static final String ORDERS_API_ENDPOINT = "/v1/etor/orders";
```
halprin, line 34: You can make these "package private" which will allow them to be used in tests but not outside the tests.

```suggestion
    static final String DEMOGRAPHICS_API_ENDPOINT = "/v1/etor/demographics";
    static final String ORDERS_API_ENDPOINT = "/v1/etor/orders";
```

halprin (approved):
Sorry it took me this long to review your PR.  I have some feedback on the visibility of your constants, but it's not enough to not approve the PR.

jorg3lopez (approved):
Looks good.

Issue #382, opened 2023-06-14
Receiving FHIR Orders via REST


JohnNKing:
# Story
As a public health lab, to be able to receive the NBS order such that it can be ingested by my LIMS, I need to receive a single order via REST that includes demographic data combined with additional order fields.

## Pre-conditions
- [x] All demographic data is submitted with the additional order fields in a single REST call. (Combining previously submitted demographic data with a partial order is a later story.)

## Acceptance Criteria
- [x] We can ingest a FHIR order that includes demographics.
- [x] An outgoing FHIR order is sent soon after receiving the incoming FHIR order
- [x] The outgoing FHIR order includes currently-supported demographic information.
- [x] The service request ID of the incoming order is unchanged and no additional FHIR orders are created.
- [x] The ability to receive only demographic data and then send out a FHIR order continues to function.
- [x] The incoming FHIR order is only accepted if and only if it is sent from a trusted organization.

### Research
- [x] Better understand how billing information can be captured as part of an order @MikeC-A6 [https://miro.com/app/board/uXjVM8df8xY=/]
  - [x] Determine how billing works for MN
- [x] Review HL7v2 to FHIR conversion spreadsheets and see how this could map to FHIR elements (e.g. service request & patient) - @cjterdi 
  - [x] Apply to MN orders - @cjterdi
- [x] Develop a knowledge base for how HL7 and/or FHIR messages are constructed, mapped, and converted (e.g. provenance)
  - [x] Evaluate use of RS terminal-based conversion for creating sample files (Answer: not currently possible)
  - [x] Review sample ADT & OML message from MN
  - [x] Create sample NBS ORM message, based on OML sample and known standards - @JohnNKing 
  - [x] Create sample NBS FHIR order, based on OML sample and known standards - @JohnNKing 
  - [x] Follow-up on ReportStream to define product needs - @JohnNKing 
  - [x] Discuss provenance on 7/7 with ReportStream - @JohnNKing 

###  Engineering
- [x] Decide on architectural design for API endpoint
- [x] Create OrdersController @basiliskus 
- [x] Create OrdersResponse @basiliskus @jorg3lopez 
- [x] Orders endpoint
   - [x] Create v1/etor/orders endpoint skeleton @jorg3lopez 
   - [x] Enforce a valid bearer token on our orders endpoint @jorg3lopez
   - [x] Process request using OrdersController @basiliskus 
   - [x] Send order to ReportStream (new usecase) - @halprin 
[...]

PR #407, opened 2023-06-23
JSON Logging

halprin:
# JSON Logging

For our logging to be better understood by Azure Log Workspace (and Splunk in the future), we need to transition to a structured logging format.  The defacto format is JSON.  I also added a shared query to our Azure Log Workspace so we can easily view our application's log.

Here's what the Azure Log Workspace looks like now.

Screenshot 2023-06-23 at 11 46 55 AM

## Issue

#382

## Checklist

_N/A_.


app/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/slf4j/Slf4jLoggerTest.groovy
```diff
@@ -1,53 +1,36 @@
...
         when:
-        def logEventBuilder = logger.getLoggingEventBuilder(Logger.Level.TRACE, expectedMessage)
-        def actualLevel = logEventBuilder["loggingEvent"]["level"]
-        def actualMessage = logEventBuilder["loggingEvent"]["message"]
-
+        def logEventBuilder = logger.getLoggingEventBuilder(Logger.Level.TRACE, "Trace message")
         then:
-        actualLevel == expectedLevel
-        actualMessage.toString().contains(expectedMessage.toString())
+        logEventBuilder.getClass() == NOPLoggingEventBuilder
```
halprin, line 18: @basiliskus, I had to radically change the tests for trace and debug.  SLF4J apparently changes what logging event builder it returns depending on what SLF4J provider we use.  It makes sense that a NOP logging event builder is returned for trace and debug since we ask for only the info level and higher.  It is is weird that that an actual logging event builder was returned previously when we used the SLF4J-simple provider and had trace and debug disabled there too.

I also question the need for these tests.  Previously, we put `Slf4jLogger` into the code coverage ignore section.  See `app/build.gradle:134`.

What are your thoughts?

app/build.gradle
```diff
@@ -54,7 +54,8 @@ dependencies {
 
     //logging
     implementation 'org.slf4j:slf4j-api:2.0.7'
-    implementation 'org.slf4j:slf4j-simple:2.0.7' // Can be replaced with logBack, log4j, log4j2, Java Utility Logger
+    implementation 'ch.qos.logback:logback-classic:1.4.8'
+    implementation 'net.logstash.logback:logstash-logback-encoder:7.4'
```
jorg3lopez, line 58: Having the simple objects made this a breeze.

jorg3lopez (approved):
Great work.

app/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/slf4j/Slf4jLoggerTest.groovy
basiliskus: I'm ok with removing these tests. When I added them I think I was not clear that it wasn't needed as it is a humble object. I also like the change for SLF4J provider to get a JSON response. It also seems we'd have more control over the logger, which is something that I remember I struggled with when I was trying to configure the simple logger

basiliskus (approved):
Thanks for improving the logger. I'm ok with removing the tests but otherwise I'm approving the PR

app/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/slf4j/Slf4jLoggerTest.groovy
halprin: @basiliskus, are you OK removing just the TRACE and DEBUG levels tests, or are you OK with removing this entire test file?  Thoughts?

basiliskus: I'm Ok with removing the entire test file unless you think there is any value to keeping it. The only logic that we added is in `getLoggingEventBuilder` which sets the console output color. And adds the FATAL level. Probably that's the only thing worth testing, if anything

PR #412, opened 2023-06-29
Process request using OrdersController

basiliskus:
- Refactored endpoint auth: moved authentication logic from `EtorDomainRegistration` to `DomainsRegistration.createHandler` to make it independent from endpoint
- Removed `constructResponse` methods from `PatientDemographicsController` and `OrderController` because of redundancy (need to validate with the team that this is alright)
- Added `isProtected` parameter to `HttpEndpoint`
- Implemented return for `OrderResponse`

## Issue

#382 
## Checklist

- [N/A?] I have updated the documentation accordingly


app/src/main/java/gov/hhs/cdc/trustedintermediary/external/javalin/DomainsRegistration.java
```diff
@@ -111,20 +113,52 @@ static DomainConnector constructNewDomainConnector(Class<? extends DomainConnect
...
+                var errorMessage = "The request failed the authentication check";
+                LOGGER.logError(errorMessage);
+                return domainResponseHelper.constructErrorResponse(401, errorMessage);
+            }
+        } catch (SecretRetrievalException | IllegalArgumentException e) {
+            LOGGER.logFatal("Unable to validate whether the request is authenticated", e);
+            return domainResponseHelper.constructErrorResponse(500, e);
+        }
+
+        return null;
```
jorg3lopez, line 159: I'm wondering if we can return a domain request with 200 status code here, and then we check for the status code instead of null in  processRequest; thoughts?

basiliskus: I think returning a 200 response at that point may be misleading because we don't know if that's going to be the status code yet. Once we call `handler.apply(request)`, we may need to return a 400 response

app/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/javalin/DomainsRegistrationTest.groovy
```diff
@@ -97,14 +103,166 @@ class DomainsRegistrationTest extends Specification {
...
+        TestApplicationContext.register(Formatter, Jackson.getInstance())
+
+        TestApplicationContext.injectRegisteredImplementations()
+
+        when:
+        def res = DomainsRegistration.authenticateRequest(new DomainRequest())
+        def actualStatusCode = res.getStatusCode()
+
+        then:
+        noExceptionThrown()
```
halprin, line 147: This line can be removed since you have other checks.  Be default, Spock checks that no exception was thrown.  The only time you need `noExceptionThrown()` is if it is the only check needed because Spock requires at least one check in the `then:` block.

app/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/javalin/DomainsRegistrationTest.groovy
```diff
@@ -97,14 +103,166 @@ class DomainsRegistrationTest extends Specification {
...
+        TestApplicationContext.injectRegisteredImplementations()
+
+        when:
+        def actualStatusCode = DomainsRegistration.processRequest(request, handler, isProtected).statusCode
+
+        then:
+        actualStatusCode == expectedStatusCode
+    }
+
+    def "processRequest unhappy path works"() {
```
halprin, line 213: Let's provide a better name for this test, if you don't mind.  I can help brainstorm too, if you want.

app/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/javalin/DomainsRegistrationTest.groovy
```diff
@@ -97,14 +103,166 @@ class DomainsRegistrationTest extends Specification {
...
+
+        when:
+        def res = DomainsRegistration.authenticateRequest(new DomainRequest())
+        def actualStatusCode = res.getStatusCode()
+
+        then:
+        actualStatusCode == expectedStatusCode
+    }
+
+    def "processRequest happy path works"() {
```
halprin, line 191: Let's provide a better name for this test, if you don't mind.  I can help brainstorm too, if you want.

app/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/javalin/DomainsRegistrationTest.groovy
```diff
@@ -97,14 +103,166 @@ class DomainsRegistrationTest extends Specification {
...
+        given:
+        def isProtected = true
+        def request = new DomainRequest()
+        def handler =  { DomainRequest req ->
+            return new DomainResponse(200)
+        }
+        def expectedStatusCode = (handler as Function<DomainRequest, DomainResponse>).apply(request).statusCode
```
halprin, line 198: That's one way to do it.  You can also do this...
```suggestion
        def expectedStatusCode = 200
        def handler =  { DomainRequest req ->
            return new DomainResponse(expectedStatusCode)
        }
```

Feel free to take this or leave this.

app/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/javalin/DomainsRegistrationTest.groovy
```diff
@@ -97,14 +103,166 @@ class DomainsRegistrationTest extends Specification {
...
+        then:
+        actualStatusCode == expectedStatusCode
+    }
+
+    def "processRequest unhappy path works"() {
+        given:
+        def isProtected = true
+        def request = new DomainRequest()
+        def handler =  { DomainRequest req ->
+            return new DomainResponse(401)
```
halprin, line 218: I recommend changing the handler's response status code to something that isn't the same as the response we expect to come from the failed authentication check.  Else, we can't differentiate between us failing the authentication check or calling the handler.

app/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/javalin/DomainsRegistrationTest.groovy
```diff
@@ -97,14 +103,166 @@ class DomainsRegistrationTest extends Specification {
...
+        when:
+        def actualStatusCode = DomainsRegistration.processRequest(request, handler, isProtected).statusCode
+
+        then:
+        actualStatusCode == expectedStatusCode
+    }
+
+    def "processRequest unhappy path works"() {
+        given:
+        def isProtected = true
```
halprin, line 215: Feel free to not specify this as a variable since we only ever use it in once below.  You can take this or leave this comment.

halprin (changes requested):
I've got a few things that should be changed.  But besides that, this is looking pretty nice!

I like where the top-level authentication check went.  I know we were sort of trying to figure out how to handle how to check authentication and potentially not call the handler gracefully.  I think it is well done.

halprin (approved):
Sweeeeeeeeet!

PR #416, opened 2023-07-03
Renamed LabOrder => Order and Orders => Order

basiliskus:
Renamed for consistency all instances of:
- `LabOrder` => `Order`
- `Orders` => `Order`


halprin (approved):
🎸

PR #418, opened 2023-07-03
382: Deploy IG Correctly

halprin:
# Deploy IG Correctly

- Deploy the IG to the correct environment instead of always the staging environment.
- Overwrite the existing files so updates actually happen.

## Issue

#382.

## Checklist

### DIFF
--- a/.github/workflows/deploy_reusable.yml
+++ b/.github/workflows/deploy_reusable.yml
@@ -117,4 +117,4 @@ jobs:
       uses: azure/CLI@v1
       with:
         inlineScript: |
-            az storage blob upload-batch --account-name cdctistagingdocs --auth-mode key -d '$web' -s ${{ steps.extract.outputs.destination }}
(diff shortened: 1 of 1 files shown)


basiliskus (approved):

.github/workflows/deploy_reusable.yml
```diff
@@ -117,4 +117,4 @@ jobs:
       uses: azure/CLI@v1
       with:
         inlineScript: |
-            az storage blob upload-batch --account-name cdctistagingdocs --auth-mode key -d '$web' -s ${{ steps.extract.outputs.destination }}
+            az storage blob upload-batch --account-name cdcti${{ inputs.ENVIRONMENT }}docs --overwrite --auth-mode key -d '$web' -s ${{ steps.extract.outputs.destination }}
```
sfradkin, line 120: we can also try using `az storage blob sync...` if `--overwrite` doesn't do what we expect.  It wasn't clear if that would delete files that should be deleted.

PR #420, opened 2023-07-03
382: Logging Locally

halprin:
# Logging Locally

In general, we no longer have the raw ANSI color codes showing up in our deployed environment's logs, and we now have easy to read logs while running locally.

Specifically...

- Added a logger named `local` to `logback.xml` that logs in a simple console format (contrast this to the JSON format).
- Renamed `SLF4JLogger` to `LocalLogger`.
- Changed `LocalLogger` to use the new `local` logger from `logback.xml`.
- Created a new `DeployedLogger` that uses a logger that continues to use the JSON format.
- Removed the `Level` enum.
- Created a helper, `LoggerHelper`, to centralize some of the logging logic.  I'm not entirely sure I'm getting much out of this helper since it replaces 1 line anyway.

## Issue

#382.

## Checklist

- [x] Rename `Slf4JLogger` to something about deployment or structure.

### DIFF
--- a/app/src/main/java/gov/hhs/cdc/trustedintermediary/external/javalin/App.java
+++ b/app/src/main/java/gov/hhs/cdc/trustedintermediary/external/javalin/App.java
@@ -67,8 +67,8 @@ public class App {
         ApplicationContext.register(
                 Logger.class,
                 ApplicationContext.getEnvironment().equalsIgnoreCase("local")
-                        ? LocalLogger.getLogger()
-                        : DeployedLogger.getLogger());
+                        ? LocalLogger.getInstance()
+                        : DeployedLogger.getInstance());
         ApplicationContext.register(Formatter.class, Jackson.getInstance());
         ApplicationContext.register(HapiFhir.class, HapiFhirImplementation.getInstance());
         ApplicationContext.register(YamlCombiner.class, Jackson.getInstance());
--- a/app/src/main/java/gov/hhs/cdc/trustedintermediary/external/javalin/App.java
+++ b/app/src/main/java/gov/hhs/cdc/trustedintermediary/external/javalin/App.java
@@ -13,8 +13,8 @@ import gov.hhs.cdc.trustedintermediary.external.inmemory.KeyCache;
 import gov.hhs.cdc.trustedintermediary.external.jackson.Jackson;
 import gov.hhs.cdc.trustedintermediary.external.jjwt.JjwtEngine;
 import gov.hhs.cdc.trustedintermediary.external.localfile.LocalSecrets;
+import gov.hhs.cdc.trustedintermediary.external.slf4j.DeployedLogger;
 import gov.hhs.cdc.trustedintermediary.external.slf4j.LocalLogger;
-import gov.hhs.cdc.trustedintermediary.external.slf4j.Slf4jLogger;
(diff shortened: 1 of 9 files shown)


app/src/main/java/gov/hhs/cdc/trustedintermediary/external/javalin/App.java
```diff
@@ -63,7 +64,11 @@ private static void registerDomains(Javalin app)
     }
 
     private static void registerClasses() {
-        ApplicationContext.register(Logger.class, Slf4jLogger.getLogger());
+        ApplicationContext.register(
+                Logger.class,
+                ApplicationContext.getEnvironment().equalsIgnoreCase("local")
+                        ? LocalLogger.getInstance()
+                        : DeployedLogger.getInstance());
```
jorg3lopez, line 71: Nice logic.

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/slf4j/DeployedLogger.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public void logDebug(String debugMessage) {
+        Level level = Level.DEBUG;
+        LoggerHelper.logMessageAtLevel(LOGGER, level, debugMessage).log();
+    }
+
+    @Override
+    public void logDebug(String debugMessage, Throwable e) {
```
basiliskus, line 36: Are there use cases for logging an exception at debug level? I can only think of uses at error/fatal level

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/slf4j/DeployedLogger.java
```diff
@@ -0,0 +1,83 @@
...
+
+    private DeployedLogger() {}
+
+    public static DeployedLogger getInstance() {
+        return INSTANCE;
+    }
+
+    @Override
+    public void logTrace(String traceMessage) {
+        Level level = Level.TRACE;
```
basiliskus, line 25: Why do we need the level variable here and other places where it's called once? Is it for consistency?

basiliskus (approved):
Looks good, thanks for the improvement. I just left a couple of comments/questions

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/slf4j/DeployedLogger.java
halprin: Yeah, this method is not even used.  I think I'll remove all the unused methods.  If we need them later, we can add them then.  Thoughts?

basiliskus: That sounds good to me. Same for the one in LocalLogger.java if it's not used. If there's an exception, I think either error or fatal level is more appropriate, unless there is some use case I'm not taking into consideration

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/slf4j/DeployedLogger.java
halprin: Uh, good point.  We don't need a separate variable.  I can pass it in directly.  Let me do that.

basiliskus (approved):
Looks good!

PR #421, opened 2023-07-04
e2e tests for Order use case implementation

ghost:
# e2e tests for Orders use case

- added e2e tests for orders endpoint implementation
- added OrderClient class for OrderTest specification

## Checklist


tjohnson7021:
I am assuming there is some refactoring that could be done since the OrderClient is basically the same as the DemographicsClient, but I wanted to make sure that is something I should try first? I am thinking that there could be some possible changes that would make the OrderClient different from the DemographicsClient in the future.

basiliskus:
> I am assuming there is some refactoring that could be done since the OrderClient is basically the same as the DemographicsClient, but I wanted to make sure that is something I should try first? I am thinking that there could be some possible changes that would make the OrderClient different from the DemographicsClient in the future.

You are right. I think this is a good opportunity to do some refactoring since there is considerable code shared between clients. Another reason to extract and create more generic methods is so we can then use them for future endpoints

e2e/src/test/groovy/gov/hhs/cdc/trustedintermediary/e2e/OrderTest.groovy
```diff
@@ -0,0 +1,65 @@
...
+
+        when:
+        def responseBody = orderClient.submit(labOrderJsonFileString.substring(1))
+        //removed beginning '{' to make this JSON invalid
+
+        then:
+        responseBody == "Server Error"
+    }
+
+    def "payload file check"() {
```
halprin, line 37: I think a better test of the sent payload is only check that it is completely the same as the input FHIR.

halprin (commented):
Great work!

e2e/src/test/groovy/gov/hhs/cdc/trustedintermediary/e2e/OrderTest.groovy
tjohnson7021: Do you mean something like this? 
Screenshot 2023-07-05 at 11 50 52 AM

halprin: Yes, exactly.

halprin (approved):

PR #422, opened 2023-07-04
Request to update gitignore

tjohnson7021:
# Update gitignore to remove a localfilelaborder.json

- remove a localfilelaborder.json that appears under app directory

## Issue #382

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ ig/template
 
 __pycache__
 localfileorder.json
+app/localfilelaborder.json


halprin (changes requested):
Just delete the file.  It won't come back.  It is an older file that was created by our older `LocalFileLabOrderSender`.  With the rename of `LabOrder` -> `Order`, the name of the file we write out change too, hence the `localfileorder.json` line above.

PR #423, opened 2023-07-04
382: IG Onboarding and Tooling Updates

halprin:
# IG Onboarding and Tooling Updates

- Updated the onboarding documentation with a proposed process.
- Created `local-generate.sh` that automates the generation of the IG instead of you having to run the Docker and IG commands manually.
- Updated the documentation on `local-generate.sh`.
- Update `fix-execute.sh` to also update the permissions for `_gencontinuous.sh`.
- Give the local IG Dockerfile a default run set of arguments (`CMD` are overwritten when providing a command in the `docker run` command).
- Remove `buildx` from the local IG creation because it isn't needed.

## Issue

#382.

## Checklist

### DIFF
--- a/ig/local-generate.sh
+++ b/ig/local-generate.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+docker build -t ig-local-dev -f Dockerfile-local-dev .
+docker run -it --rm --mount type=bind,source="$(pwd)",target=/trusted-intermediary ig-local-dev
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  open ./output/index.html
+fi
--- a/ig/Dockerfile-local-dev
+++ b/ig/Dockerfile-local-dev
@@ -5,3 +5,5 @@ RUN cp `find /usr/lib/ruby/gems/*/gems/jekyll-*/exe/jekyll` /usr/bin
 RUN npm install -g fsh-sushi
 
 WORKDIR /trusted-intermediary
+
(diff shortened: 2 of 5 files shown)


ig/local-generate.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+docker build -t ig-local-dev -f Dockerfile-local-dev .
+docker run -it --rm --mount type=bind,source="$(pwd)",target=/trusted-intermediary ig-local-dev
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
```
jorg3lopez, line 6:
```suggestion
if [[ "$OSTYPE" == "darwin"* ]] || [[ "$OSTYPE" == "linux-gnu" ]]; then
```
This will open the browser for linux users

jorg3lopez (changes requested):

jorg3lopez (approved):
I left a comment but looks good, good job.

ig/local-generate.sh
halprin: Does Linux have the `open` command?  I thought it didn't.  According to the man page for `open`, it came from NextStep which is the precursor to Mac OS X (10).

jorg3lopez: It does. I think the problem I had was that the conditional statement, once I added the the rest of the code, it opened a browser for me.

halprin: Oh, super cool!

PR #424, opened 2023-07-04
Refactor e2e helper classes

basiliskus:
- Refactored helper classes to simplify and clarify
- Renamed `Client` => `HttpClient`
- Added test assertions for HTTP status code responses
- Removed redundant test
- Some cleanup


e2e/src/main/java/gov/hhs/cdc/trustedintermediary/e2e/HttpClient.java
```diff
@@ -13,7 +13,7 @@
 import org.apache.hc.core5.http.io.entity.StringEntity;
 
 /** Mocks a client sending a request to the API * */
-public class Client {
+public class HttpClient {
 
     private static final String DEFAULT_PROTOCOL_DOMAIN = "http://localhost:8080";
     private static String protocolDomain = DEFAULT_PROTOCOL_DOMAIN;
```
jorg3lopez, line 19: I know this doesn't include any of the changes, but is there a reason to use `protocolDomain` instead of `DEFAULT_PROTOCOL_DOMAIN` ?

halprin: I wrote that a while ago.  I did that because of the off chance that we can change `protocolDomain` to something different than the default.  There isn't a way to do that right now with how the code is written, but a setter could be created that overwrites `protocolDomain`.

halprin (approved):

PR #428, opened 2023-07-06
ResponseBody Error Message

jorg3lopez:
# ResponseBody Error Message

This PR adds adds code to handle an internal exception when the payload is empty or has a bad format.

## Checklist

### DIFF
--- a/app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
+++ b/app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
@@ -8,8 +8,10 @@ import gov.hhs.cdc.trustedintermediary.domainconnector.DomainResponseHelper;
 import gov.hhs.cdc.trustedintermediary.domainconnector.HttpEndpoint;
 import gov.hhs.cdc.trustedintermediary.domainconnector.UnableToReadOpenApiSpecificationException;
 import gov.hhs.cdc.trustedintermediary.etor.demographics.ConvertAndSendDemographicsUsecase;
+import gov.hhs.cdc.trustedintermediary.etor.demographics.Demographics;
 import gov.hhs.cdc.trustedintermediary.etor.demographics.PatientDemographicsController;
 import gov.hhs.cdc.trustedintermediary.etor.demographics.PatientDemographicsResponse;
+import gov.hhs.cdc.trustedintermediary.etor.orders.Order;
 import gov.hhs.cdc.trustedintermediary.etor.orders.OrderController;
 import gov.hhs.cdc.trustedintermediary.etor.orders.OrderConverter;
 import gov.hhs.cdc.trustedintermediary.etor.orders.OrderResponse;
@@ -82,9 +84,10 @@ public class EtorDomainRegistration implements DomainConnector {
     }
 
     DomainResponse handleDemographics(DomainRequest request) {
-        var demographics = patientDemographicsController.parseDemographics(request);
+        Demographics<?> demographics;
 
         try {
+            demographics = patientDemographicsController.parseDemographics(request);
             convertAndSendDemographicsUsecase.convertAndSend(demographics);
         } catch (UnableToSendOrderException e) {
             logger.logError("Unable to convert and send demographics", e);
@@ -98,9 +101,10 @@ public class EtorDomainRegistration implements DomainConnector {
     }
 
     DomainResponse handleOrders(DomainRequest request) {
-        var orders = orderController.parseOrders(request);
 
+        Order<?> orders;
         try {
+            orders = orderController.parseOrders(request);
             sendOrderUseCase.send(orders);
         } catch (UnableToSendOrderException e) {
             logger.logError("Unable to send lab order", e);
--- a/app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
+++ b/app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
@@ -101,8 +101,8 @@ public class EtorDomainRegistration implements DomainConnector {
     }
 
     DomainResponse handleOrders(DomainRequest request) {
-
         Order<?> orders;
+
         try {
             orders = orderController.parseOrders(request);
             sendOrderUseCase.send(orders);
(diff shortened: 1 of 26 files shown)


app/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiFhirImplementation.java
```diff
@@ -20,9 +21,16 @@ public static HapiFhirImplementation getInstance() {
...
     public <T extends IBaseResource> T parseResource(
-            final String fhirResource, final Class<T> clazz) {
+            final String fhirResource, final Class<T> clazz) throws UnableToSendOrderException {
         IParser resourceParser = CONTEXT.newJsonParser();
-        return resourceParser.parseResource(clazz, fhirResource);
+
+        try {
+            return resourceParser.parseResource(clazz, fhirResource);
+        } catch (Exception e) {
+            throw new UnableToSendOrderException(
```
basiliskus, line 30: We should discuss if we need to create a new exception (e.g. `UnableToParseRequestException`). If not, we should update the error messages in `EtorDomainRegistration.handleDemographics` and `EtorDomainRegistration.handleOrders` (e.g. from `Unable to send lab order` to `Unable to parse request or send lab order`) (CC: @halprin)

basiliskus:
Another question. Does it make sense to add the `400` response to `openapi_etor.yaml` docs to document the case when the caller sends a badly formatted request?

app/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiFhirImplementation.java
halprin: I would recommend a new exception instead of `UnableToSendOrderException`.  `FhirParseException`?

halprin:
> Another question. Does it make sense to add the `400` response to `openapi_etor.yaml` docs to document the case when the caller sends a badly formatted request?

Yes, it does make sense add `400` to the OpenAPI documentation.

app/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistrationTest.groovy
```diff
@@ -116,6 +117,32 @@ class EtorDomainRegistrationTest extends Specification {
...
+        mockHelper.constructErrorResponse(expectedStatusCode, fhirParseException) >> { new DomainResponse(expectedStatusCode)}
+        TestApplicationContext.register(DomainResponseHelper, mockHelper)
+        TestApplicationContext.injectRegisteredImplementations()
+
+        when:
+        def res = domainRegistration.handleDemographics(mockRequest)
+        def actualStatusCode = res.getStatusCode()
+
+        then:
+        actualStatusCode == 400
```
jorg3lopez, line 143: `actualStatusCode == expectedStatusCode`

app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
```diff
@@ -98,12 +104,16 @@ DomainResponse handleDemographics(DomainRequest request) {
...
 
         try {
+            orders = orderController.parseOrders(request);
             sendOrderUseCase.send(orders);
+        } catch (FhirParseException e) {
+            logger.logError("Unable to parse order request", e);
+            return domainResponseHelper.constructErrorResponse(400, e);
         } catch (UnableToSendOrderException e) {
-            logger.logError("Unable to send lab order", e);
+            logger.logError("Unable to convert and send order", e);
```
halprin, line 116: There isn't really any conversion we're doing yet.

app/src/main/resources/openapi_etor.yaml
```diff
@@ -83,6 +95,8 @@ components:
         patientId:
           type: string
           example: MRN7465737865
+    BadRequestError:
+      description: Bad Request
```
halprin, line 99: We return a JSON response body, right?  With an `error` field, yes?  If so, we should flesh this out with those details.  See the above `OrdersResponse` on how to do this.

app/src/main/java/gov/hhs/cdc/trustedintermediary/etor/FhirParseException.java
```diff
@@ -0,0 +1,12 @@
+package gov.hhs.cdc.trustedintermediary.etor;
+
+/**
+ * This exception class gets triggered when an exception is thrown during parsing of a fhir bundle.
+ * Some causes that will trigger it are: an empty payload, wrong payload format as it expects a fhir
+ * bundle.
+ */
+public class FhirParseException extends Exception {
```
halprin, line 8: This should live in the `wrapper` package.  Next to the `HapiFhir` interface.

e2e/src/test/groovy/gov/hhs/cdc/trustedintermediary/e2e/DemographicsTest.groovy
```diff
@@ -51,17 +51,20 @@ class DemographicsTest extends Specification {
 
         when:
         def response = demographicsClient.submit(invalidJsonRequest, true)
-        def responseBody = EntityUtils.toString(response.getEntity())
+        def parsedJsonBody = JsonParsing.parseContent(response)
 
         then:
         response.getCode() == 400
+        !StringUtils.isEmpty(parsedJsonBody.error as String)
```
halprin, line 58: This seems like a weird helper: `org.apache.groovy.parser.antlr4.util.StringUtils`.  Especially since this seems to be some internal-ish helper associated with ANTLR.  You can do the same natively in Groovy by doing the following.
```suggestion
        !(parsedJsonBody.error as String).isEmpty()
```

halprin (approved):
This is great!  I'm a little iffy on the `StringUtils` helper though.

e2e/src/test/groovy/gov/hhs/cdc/trustedintermediary/e2e/DemographicsTest.groovy
basiliskus: Good point, I didn't notice it was coming from antlr, I thought it was internal. I'll use your suggestion, that's what I was looking for

PR #432, opened 2023-07-10
382: Internals in the Implementation Guide

halprin:
# Internals in the Implementation Guide

Add a starting point for our internal section of the IG.  I think this could be fleshed out in the future.

@jorg3lopez, you mentioned about adding additional documentation around interacting with our API.  This file could be where that documentation resides.

## Issue

#382.

## Checklist

### DIFF
--- a/ig/input/pagecontent/internals.md
+++ b/ig/input/pagecontent/internals.md
@@ -1,3 +1,12 @@
 ### Overview
 
-This is where we talk about GitHub and our internal interfaces.
+The ReportStream Intermediary is microservice behind the larger ReportStream system.  This page documents some of the internals behind the microservice.
+
+### GitHub
+
+We develop the ReportStream Intermediary on GitHub in an [opensource repository](https://github.com/CDCgov/trusted-intermediary).
+Please feel free to peruse our code and contribute.
+
+### Internal API
+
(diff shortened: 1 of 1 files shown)


ig/input/pagecontent/internals.md
```diff
@@ -1,3 +1,16 @@
 ### Overview
 
-This is where we talk about GitHub and our internal interfaces.
+The ReportStream Intermediary is microservice behind the larger ReportStream system.  This page documents some of the internals behind the microservice.
```
basiliskus, line 3: Is an `a` missing?: `is microservice` => `is **a** microservice`

halprin: Good catch.  English hard.

halprin: Fixed.

basiliskus (approved):

sfradkin:
I know this is already merged, but is the OpenAPI page accessible on the deployed application?  If so, we should add that link to the documentation as well. @halprin

halprin:
> I know this is already merged, but is the OpenAPI page accessible on the deployed application? If so, we should add that link to the documentation as well. @halprin

The Swagger UI?  No, it is not.  We only have the raw OpenAPI YAML exists.

PR #439, opened 2023-07-11
Add e2e test assertions for auth endpoint failures

basiliskus:
# Add e2e test assertions for auth endpoint failures

Added e2e test assertions to make sure auth endpoint returns error message for failure responses (400 & 401)

## Issue

None


halprin (approved):
Oops, I forgot.  Good catch!

Issue #441, opened 2023-07-14
Log the ReportStream `submissionId` when we send data to RS


halprin:
# DevEx/OpEx

We send data to RS, and they send us back a `submissionId`.  We should log it.

## Tasks

- [x] Add logging.

## Additional Context

https://flexion.slack.com/archives/C055XTF22B0/p1689087031085659

PR #455, opened 2023-08-07
Log the ReportStream submissionId

basiliskus:
# Log the ReportStream submissionId

Log the ReportStream `submissionId` when we send data to RS

## Issue

#441 

## Checklist


tjohnson7021 (approved):

jorg3lopez (approved):
Great work. The end2end test is failing for me but it seems that it is also failing in main. It might just be my environment as it doesn't fail in here, so I will approve it.

PR #459, opened 2023-08-09
Bug Fix

jorg3lopez:
#End to End Bug Fix

This PR fixes the e2e test. Two tests were failing due to formatting issues. The tests were comparing two different files. Instead, I changed the path of the SentPayloadReader to read from the `examples` folder instead from a local file that was created.


e2e/src/test/groovy/gov/hhs/cdc/trustedintermediary/e2e/OrderTest.groovy
```diff
@@ -25,18 +25,6 @@ class OrderTest extends Specification {
...
-    def "payload file check"() {
-        when:
-        def response = orderClient.submit(labOrderJsonFileString, true)
-        def sentPayload = SentPayloadReader.read()
-        def parsedSentPayload = JsonParsing.parse(sentPayload)
-        def parsedLabOrderJsonFile = JsonParsing.parse(labOrderJsonFileString)
-
-        then:
-        response.getCode() == 200
-        parsedSentPayload == parsedLabOrderJsonFile
```
jorg3lopez, line 37: File was deleted because we only check for the status of 200, which is being done in another test.

e2e/src/main/java/gov/hhs/cdc/trustedintermediary/e2e/SentPayloadReader.java
```diff
@@ -14,13 +14,7 @@ public static String read() throws IOException {
 
     private static Path findFilePayload() {
```
basiliskus, line 15: If we're not using this method anywhere else, we probably don't need the `SentPayloadReader` class. We could just move the file opening line to the test where it's called. That way we don't have unnecessary complexity and it would make the test more readable

basiliskus (changes requested):
Great job, I just added some feedback on the findFilePayload class

basiliskus (approved):
Looks good! Thanks for addressing my feedback

Issue #458, opened 2023-08-09
Refactor tests to stop using `localfileorder.json` and deduplicate `lab_order.json`


basiliskus:
# DevEx/OpEx

The use and creation of `localfileorder.json` in multiple tests means it has unreliable contents depending of the test order of execution. Tests should be isolated to avoid this. Each test should create and destroy the files needed for the test

## Tasks

- [x] ~~Find all uses of `localfileorder.json` and refactor so each test creates and destroys the file~~ (Looks like this was already completely addressed with [this PR](https://github.com/CDCgov/trusted-intermediary/pull/459))
- [x] Try to create and read `localfileorder.json` from only one location 
- [x] Check if possible to change `app/bin/test/gov/hhs/cdc/trustedintermediary/external/localfile/LocalFileOrderSenderTest.groovy:"send order works"` test to delete `localfileorder.json` within the test
- [x] Remove `e2e/src/test/resources/lab_order.json` and replace its calls to point to `examples/fhir/lab_order.json` instead, in order to avoid duplication of `lab_order.json` file 
- [x] Check if it's possible to remove `e2e/src/test/resources/newborn_patient.json` and use `examples/fhir/MN NBS FHIR Order Message.json` instead
- [x] Check that end to end tests are running correctly and failing when appropriate in github CI/CD

## Additional Context

[Relevant guidelines to consider when writing tests](https://kentbeck.github.io/TestDesiderata/)

PR #473, opened 2023-08-16
Reverted changes to bring back SentPayloadReader and newborn_patient.json

basiliskus:
- Bringed back `newborn_patient.json` and moved to `examples/fhir` folder
- Bringed back `SentPayloadReader` class
- Still pending: make sure we use only one copy of `localfileorder.json`

## Issue

#458


halprin (approved):
Success!

Issue #290, opened 2023-04-25
Order HL7 ORM => FHIR


JohnNKing:
# Story
As a hospital, to be able to transmit an NBS order via my EHR’s native order interface, I need the CDC ReportStream ecosystem to be able to receive an HL7 ORM order message and then send out an HL7 OML order message to a synthetic lab.

## Pre-conditions
- [x] The EHR is able to create an order that matches the HL7 v2 ORM standard.
- [x] The ability to convert outgoing FHIR to OML messages is already supported

## Acceptance Criteria
- [x] Inbound ORM messages can be received by ReportStream.
- [x] The order is transmitted to the synthetic lab soon after it has been received. Note: Not all data elements in the order need yet be included in the OML message.

## Related Stories
https://github.com/CDCgov/prime-reportstream/issues/10370

### Research
- [x] Contrast CA, MN, & LA sample messages and/or IGs
- [x] Ecosystem Mapping
  - [x] Identify how ORM data is collected for each state
- [x] Review CA's IG (https://drive.google.com/drive/folders/1JxbGZY9kRG8HwFaepSCJIntgORe9edUT)
- [x] Review Epic's Orders Tech Spec (https://open.epic.com/Tech/TechSpec?spec=staged%2FOutgoing%20Results%20and%20Orders%20Interface%20Technical%20Specification.zip or https://drive.google.com/drive/folders/1teOpcHrF74PwJB9FmssEOTfyNhlY6j4Z)
- [x] Review feedback on Cerner IGs (see https://aphlinformatics.atlassian.net/wiki/spaces/IR/pages/1981383536/6+28+2023+CDC+APHL+Flexion+Oracle+Cerner+-+Biweekly+ETOR+Mtg+Notes comments and attachment from Alexa; also reply via email forwarded 7/12 titled "RE: CDC/APHL/Flexion/Oracle Cerner - Biweekly ETOR Mtg")
- [x] Develop a [matrix](https://docs.google.com/spreadsheets/d/1yUZ5SnL5et3QYvu6p1Luk3AUSuOziCL-lRRI8uyYfn4/edit#gid=471449702) (contrasting info with EHR) of how EHRs may be able to provide mother's name and specimen ID. (e.g. mother's name via PID.6, PID.21, GT1, NK1, OBX; specimen ID via OBR, SPM, OBX; practitioner info via OBX, PD1) 
  ~~- [ ] Then request feedback from Epic and Cerner~~
[...]

PR #476, opened 2023-08-18
290: ORM to OML

halprin:
# ORM to OML

Added the ability for our order converter to also convert the order metadata from an ORM to an OML.  This is then used in our sending order usecase.

As part of the `HapiOrderConverter` changes, I also refactored how we set the OML `Coding`.  It used to be created on each call to the convert-demographics-to-order call, but I decided to promote the OML `Coding` object to a static class field now that it is also used identically in the new order-metadata-conversion call.

Lastly, I updated our e2e order test to use a different input file.  This input file is an ORM, which reflects how we will be called, instead of an OML file.

## Issue

#290

## Checklist


basiliskus (approved):
Looks good!

jorg3lopez (approved):
LGTM, good job!

PR #477, opened 2023-08-18
Better Sent Payload File Handling

halprin:
# Better Sent Payload File Handling

Our e2e tests can be brittle and create confusing errors depending on how you've run these tests in the past.

This PR updates how our e2e tests run such that the sent payload file is generated in the same folder no matter how you may run TI.  This simplifies our `SentPayloadReader` in our e2e tests.  In addition to this simplification, I also added an ability to delete the sent payload file, and I call this in our e2e tests so they clean-up.

## Issue

_None_.

## Checklist

### DIFF
--- a/e2e/src/main/java/gov/hhs/cdc/trustedintermediary/e2e/SentPayloadReader.java
+++ b/e2e/src/main/java/gov/hhs/cdc/trustedintermediary/e2e/SentPayloadReader.java
@@ -6,22 +6,13 @@ import java.nio.file.Path;
 
 public class SentPayloadReader {
 
-    public static String read() throws IOException {
-        Path payloadFile = findFilePayload();
+    private static final Path SENT_PAYLOAD_PATH = Path.of("..", "app", "localfileorder.json");
 
-        return Files.readString(payloadFile);
+    public static String read() throws IOException {
+        return Files.readString(SENT_PAYLOAD_PATH);
     }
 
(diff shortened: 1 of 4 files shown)


basiliskus (approved):
Looks good. I just have a question to potentially change the handling of the localfileorder.json file in the future: Could we create it as a temporary file and let the OS handle where to create it, and to delete it after some time?

e2e/src/main/java/gov/hhs/cdc/trustedintermediary/e2e/SentPayloadReader.java
```diff
@@ -6,22 +6,13 @@
...
-
-        boolean doesFileExist = Files.exists(expectedFilePath);
-
-        if (!doesFileExist) {
-            expectedFilePath = Path.of("..", "localfileorder.json");
-        }
-
-        return expectedFilePath;
+    public static void delete() throws IOException {
+        Files.deleteIfExists(SENT_PAYLOAD_PATH);
```
jorg3lopez, line 16: Nice!

jorg3lopez (approved):
Great job, the localfileorder file gets deleted after the test.

halprin:
> Looks good. I just have a question to potentially change the handling of the localfileorder.json file in the future: Could we create it as a temporary file and let the OS handle where to create it, and to delete it after some time?

That's a great idea that we can do now or later!

PR #505, opened 2023-08-29
Add provenance for tracking changes

ghost:
# Add Provenance to Implementation Guide

- need to check for correctness

## Checklist


ig/input/fsh/logical-adt-01.fsh
```diff
@@ -10,6 +10,21 @@ Description: "The expected input for the demographic data for the newborn screen
...
 
+Instance: adt01-initial-history
+InstanceOf: Provenance
+Title: "Initial creation of ADT01 message changelog"
+Usage: #definition
+* target[+] = Reference(Logical/adt-01-logical-model)
+* recorded = "2023-01-30T13:00:00.0000Z"
+* occurredDateTime = "2023-08-29"
+* reason = Code(http://terminology.hl7.org/CodeSystem/v3-ActReason, "METAMGT")
+* reason.text = "Initial Logical as per the C2S layout"
```
sfradkin: this should be the comment for the history entry, so something a bit more detailed than the `Title`

ig/input/fsh/logical-adt-01.fsh
```diff
@@ -10,6 +10,21 @@ Description: "The expected input for the demographic data for the newborn screen
...
+* activity = Code(http://terminology.hl7.org/CodeSystem/v3-DataOperation, "CREATE")
```
sfradkin, line 23: For the initial history entry this would be "CREATE", but subsequent will most likely be "UPDATE"

ig/input/fsh/logical-adt-01.fsh
```diff
@@ -10,6 +10,21 @@ Description: "The expected input for the demographic data for the newborn screen
...
+* agent[0].type = Coding(http://terminology.hl7.org/CodeSystem/provenance-participant-type, "author")
+* agent[0].who.display = "T. R. Johnson"
+* agent[1].type = Coding(http://terminology.hl7.org/CodeSystem/provenance-participant-type, "verifier")
```
sfradkin: I think we can leave out verifier, unless we want "verifier" to refer to anyone other than the author who helped verify the updates

ig/input/fsh/logical-adt-01.fsh
tjohnson7021: Ok  that makes sense, thanks!

sfradkin (approved):
This looks good as the beginning of the changelogs for each of these.  As long as we verify how this looks when the IG is regenerated, let's go ahead and merge.

basiliskus (commented):
I don't have experience with the `fsh` language, but it looks good to me

halprin (approved):
Did a quick glance.  Ship it.

PR #512, opened 2023-09-01
Update ORC segment IG documentation

basiliskus:
Updated ORC segment IG documentation for required and currently captured fields

## Issue

#492 

## Checklist


ig/input/fsh/nbs-orc-segment.fsh
```diff
@@ -2,15 +2,20 @@ Logical: ORCSegment
...
-* orderControl 1..1 SU string "ORC-1 Order Control"
-* ORCPlacerOrderNumber 1..1 SU string "ORC-2 Placer Order Number"
-* fillerOrderNumber 1..1 SU string "ORC-3 Filler Order Number"
-* placerGroupNumber 1..1 SU string "ORC-4 Placer Group Number"
-* dateTimeOfTransaction 1..1 SU string "ORC-9 Date/Time of Transaction"
-* orderingProvider 1..1 SU string "ORC-12 Ordering Provider"
-* orderingFacilityName 1..1 SU string "ORC-11 prdering Facility Name"
-// other optional fields
-
+* orderControl 1..1 SU id "ORC-1 Order Control"
```
halprin, line 5: I would make this `code`.  [An `id`](http://hl7.org/fhir/R4/datatypes.html#primitive) is a...
> Any combination of upper- or lower-case ASCII letters ('A'..'Z', and 'a'..'z', numerals ('0'..'9'), '-' and '.', with a length limit of 64 characters. (This might be an integer, an un-prefixed OID, UUID or any other identifier pattern that meets these constraints.)

While that description does apply to this field, I feel that `code` is more appropriate.
> Indicates that the value is taken from a set of controlled strings defined elsewhere (see [Using codes](http://hl7.org/fhir/R4/terminologies.html) for further discussion). Technically, a code is restricted to a string which has at least one character and no leading or trailing whitespace, and where there is no whitespace other than single spaces in the contents

Specifically, the order control field is taken from the [0119 controlled set of strings](https://hl7-definition.caristix.com/v2/HL7v2.5.1/Tables/0119).

```suggestion
* orderControl 1..1 SU code "ORC-1 Order Control"
```

ig/input/fsh/nbs-orc-segment.fsh
```diff
@@ -2,15 +2,20 @@ Logical: ORCSegment
...
+* placerOrderNumber 1..1 SU Identifier "ORC-2 Placer Order Number"
```
halprin, line 6: I wouldn't use any complex data types here because the complex types suggest additional fields and structure where there isn't in these HL7 fields.  For example, [`Identifier` states](http://hl7.org/fhir/R4/datatypes.html#identifier) it has a sub-field of `use`, `system`, and `value`.  This HL7 field doesn't have those as sub-fields.  So, I'd just stick to the [primitive types](http://hl7.org/fhir/R4/datatypes.html#primitive).

```suggestion
* placerOrderNumber 1..1 SU string "ORC-2 Placer Order Number"
```

ig/input/fsh/nbs-orc-segment.fsh
```diff
@@ -2,15 +2,20 @@ Logical: ORCSegment
...
+* placerOrderNumberEntityIdentifier 1..1 SU string "ORC-2.1 Placer Order Number Entity Identifier"
+* placerOrderNumberNamespaceID 1..1 SU code "ORC-2.2 Placer Order Number Namespace ID"
+* placerOrderNumberUniversalID 1..1 SU string "ORC-2.3 Placer Order Number Universal ID"
+* placerOrderNumberUniversalIDType 1..1 SU id "ORC-2.4 Placer Order Number Universal ID Type"
```
halprin, line 10: I think the general gist is that when a field (like `ORC-2.2`) is controlled but is completely [user defined](https://hl7-definition.caristix.com/v2/HL7v2.5.1/Tables/0363), then it is a `string` because it's fairly open ended still.  On the other hand, `ORC-2.4` is controlled by a very distinct set of controlled options ([table 0301](https://hl7-definition.caristix.com/v2/HL7v2.5.1/Tables/0301) specifically).

```suggestion
* placerOrderNumberNamespaceID 1..1 SU string "ORC-2.2 Placer Order Number Namespace ID"
* placerOrderNumberUniversalID 1..1 SU string "ORC-2.3 Placer Order Number Universal ID"
* placerOrderNumberUniversalIDType 1..1 SU code "ORC-2.4 Placer Order Number Universal ID Type"
```

ig/input/fsh/nbs-orc-segment.fsh
```diff
@@ -2,15 +2,20 @@ Logical: ORCSegment
...
+* orderStatus 0..1 SU string "ORC-5 Order Status"
+* dateTimeOfTransaction 1..1 SU dateTime "ORC-9 Date/Time of Transaction"
+* orderingFacilityName 1..2 SU string "ORC-21 Ordering Facility Name"
+* orderingFacilityNameOrganizationName 1..1 SU string "ORC-21.1 Ordering Facility Organization Name"
+* orderingFacilityNameTypeCode 0..1 SU code "ORC-21.2 Ordering Facility Type Code"
+* orderingFacilityNameAssigningAuthority 1..1 SU code "ORC-21.6 Ordering Facility Assigning Authority"
+* orderingFacilityNameIdentifierTypeCode 1..1 SU id "ORC-21.7 Ordering Facility Identifier Type Code"
+* orderingFacilityNameOrganizationIdentifier 1..1 SU string "ORC-21.10 Ordering Facility Organization Identifier"
```
halprin, line 18: Similar to the above comment, I think `ORC-21.6` is a user defined such that it should be `string`.  Everything else looks good.

halprin (commented):
I've got some suggestions for changing some of the types.

basiliskus:
> I've got some suggestions for changing some of the types.

@halprin thanks for review and documentation behind the proposed changes! I have a better idea on how to choose the types now. I made the changes

halprin (approved):
Woo hoo!  Ship it.

PR #521, opened 2023-09-07
DoD for OBX Segment for outbound OML 

ghost:
# Updated required OBX Segment fields for Implementation Guide 
- added initially supported fields 
- added provenance update

## Checklist


sfradkin (approved):

ig/input/fsh/nbs-obx-segment.fsh
```diff
@@ -2,16 +2,32 @@ Logical: OBXSegment
...
+* identifier 1..1 string "OBX-3.1 Identifier"
+* observationtText 1..1 string "OBX-3.2 Text"
+* observationNameOfCodingSystem 1..1 id "OBX-3.3 Name of Coding System"
+* observationSubID 1..1 string "OBX-4 (LOINC 54089-8) Observation Sub-ID"
+* observationValue 1..1 string "OBX-5 (LOINC or 99MDH) Observation Value"
+* observationValueUnits 1..1 code "OBX-6 Units (Observation Value)"
+* obxValueUnitsIdentifier 1..1 code "OBX-6.1 (Unit) Identifier"
+* obxValueUnitsText 1..1 code "OBX-6.2 (Unit) Text"
+* obxValueUnitsNameOfCodingSystem 1..1 code "OBX-6.3 (Unit) Name of Coding System"
+* observationReferenceRange 1..1 string "OBX-7 Observation Reference Range"
```
basiliskus, line 17: Even though we may have the mapping for these additional OBX fields (OBX-7, OBX-8 and OBX-15) because we got them from ORU, I'm not sure if we should document them until we actually have them in an OML message and have verified that the conversion is working for those

basiliskus (commented):
I left a few comments/questions. I thought we were supposed to add the names for each field exactly as they are in the documentation, but seeing your PR it seems I had the wrong assumption. Do we we need to describe the fields in plain language or do we need to document the names matching the HL7 guideline?

PR #523, opened 2023-09-10
Split Projects

halprin:
# Split Projects

Split our mono-project into three.

1. `shared` - contains all the shared functionality that would be used across plugins.
2. `etor` - the ETOR plugin that depends on `shared` and contains only the things associated with ETOR.
3. `app` - the Javalin application that depends on `shared` and `etor`.

We now have a top level `build.gradle` that sets the settings for the subprojects.

To run the unit tests, execute `./gradlew clean allUnitTests`.  To run the code coverage verification, execute `./gradlew clean allUnitTests testCodeCoverageReport jacocoTestCoverageVerification`.

### Notes

1. `AuthControllerTest.groovy` - the deleted `import` statements are due to them not being needed in the first place.  They weren't (and still aren't) being used in the test.
2. `FhirParseExceptionTest.groovy` - the deleted `import` statement is due to it not being needed.  Because the test and `FhirParseException.java` being in the same package, you don't need an explicit import.
3. `FormatterProcessingExceptionTest.groovy` - the changed `package` statement is due to moving the test under the `formatter` sub-package.  It was moved because it tests a class under the `formatter` sub-package.  Tests should be in the same package as the class it tests.  That is then the reason why the subsequent `import` statement is deleted.

## Issue

_None._

## Checklist


build.gradle
```diff
@@ -0,0 +1,103 @@
...
+        '**/slf4j/LocalLogger*',
+        '**/slf4j/DeployedLogger*',
+        '**/slf4j/LoggerHelper*',
+        '**/hapi/HapiFhirImplementation*',
+        '**/jjwt/JjwtEngine*',
+        '**/apache/ApacheClient*',
+        '**/azure/AzureSecrets*'
+]
+
+tasks.register('allUnitTests') {
```
basiliskus, line 88: Could we have a similar task for all builds? e.g. `allBuilds`

basiliskus (approved):
Looks good! I just left one comment/question

build.gradle
halprin: Are you thinking it would be used for the the Sonar scan?  In any case, sure, happy to.

basiliskus: Yes, I thought about it after I saw the sonar scan build call, but it could also be useful as a shorthand

tjohnson7021 (approved):

PR #530, opened 2023-09-12
Rename fsh provenance instance id to avoid duplication

basiliskus:
Renamed fsh provenance instance id to avoid duplication. [This fixes the IG build failure](https://github.com/CDCgov/trusted-intermediary/actions/runs/6162751090/job/16725206712#step:8:774)


halprin (approved):
Success!

PR #546, opened 2023-09-27
Remove unneeded `SU` tag from IG pages

ghost:
# Remove `SU` tag from IG pages

It was decided a while back [here](https://flexion.slack.com/archives/C055XTF22B0/p1694209960453609) that we would stop using the `SU` tag in the IG because it had no meaning in the logical models. 
 
This PR removes the `SU` tag from the models. 

[## Issue](https://flexion.slack.com/archives/C055XTF22B0/p1694209960453609)


jcrichlake (approved):

sfradkin (approved):
Thanks for doing this!

Issue #368, opened 2023-06-08
Revert Docker Compose Version Workaround


halprin:
# DevEx/OpEx

Revert PR #367 at a later point in the future when the workaround is no longer needed.  GitHub Action runners already have Docker Compose installed, but the current version (as of this writing) used in the runner is broken.  At some point, GitHub will update their runner to have an even later version that isn't broken.  At that point we should revert the work in this PR so we depend on the default version of Docker Compose provided in the runner.

Consider checking a month from now, around July 8, 2023.

## Tasks

- [x] Revert the work in the PR.

## Additional Context

https://github.com/CDCgov/trusted-intermediary/pull/367

PR #549, opened 2023-09-29
reverting docker compose workaround

jcrichlake:
# Using Github actions docker image

We had to put a workaround in to stop relying on the Github actions docker-compose image since there was an error with that specific version.  This is removing the workaround and putting us back to the standard image used by the actions.

## Issue

#368

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,14 +41,6 @@ jobs:
       - name: ShadowJar
         run: ./gradlew shadowJar
 
-      - name: Install Latest Version of Docker Compose v2
-        run: |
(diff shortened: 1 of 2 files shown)


halprin:
> ## Issue
> [#368](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

Mind changing the the text to be just `#368` instead of `[#368](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)`?  That way GitHub automatically links to that issue instead of linking to that generic GitHub documentation webpage.

Issue #496, opened 2023-08-24
Automated Deployment to CDC Dev Environment


halprin:
# Backlog Task

Automated Deployment to CDC Dev Environment

## Completion Criteria

- [x] A full deployment successfully completes of our dev environment into our CDC Dev environment.
- [x] The full deployment to the CDC Dev environment can happen programmatically as it does today to the Flexion Dev environment.
- [x] We no longer deploy to the Flexion Dev environment.

## Tasks

- [x] Figure out if we can do it the same way we do it for Flexion.
- [x] Figure out how RS does it.
- [x] Make a help desk ticket to create an App Registration that we have permission over.  In lieu of having permission, ask the help desk to create the federated credentials for us.
- [x] Reached out to call phone support to ping the person assigned to our ticket.
- [x] Wait to hear back from help desk.
- [x] Work with help desk.
- [x] Rename Azure app integration.
- [x] Create ticket to add additional permissions to Azure app integration so it can create/read/update/delete resources in our CDC dev environment.
- [x] Get the additional permissions
- [x] Add federated credentials
- [x] Experiment with dev deployment! - @halprin 
- [x] Delete the resource group - @halprin 
- [x] Triage IG deployment
- [x] `azurerm_key_vault_access_policy.allow_github_deployer` needs to have the `object_id` passed instead of hardcoded - @halprin 
- [x] `internal` environment deploy from the `internal` branch. - @halprin 
- [x] Better documentation - @halprin 
- [x] Have RS update their organization settings to point to the new staging endpoint. - @halprin 
- [x] Fix-up why things aren't delivering to the `flexion.simulated-lab`.

## Other Notes

- Our -SU accounts do not have permission to create an App Registration, nor permission to add federated credentials to an existing App Registration.
- [Discussion with Jeremy](https://flexion.slack.com/archives/C048S3GPSQG/p1694641017795869) on how ReportStream does their deployments.

PR #553, opened 2023-10-02
496: `internal` Environment Deploy

halprin:
# `internal` Environment Deploy

- Renamed the `flexion` environment to `internal`.
- Create a GitHub action that deploys the `internal` environment when the `internal` branch is committed to.

## Issue

#496.

## Checklist


sfradkin (approved):
looks good

PR #554, opened 2023-10-02
496: Documentation on Environments

halprin:
# Documentation on Environments

Updated our main `README.md` to document our different environments, their purpose, and how they deploy.  I also updated our documentation for Initial Azure and GitHub Configuration.

## Issue

#496.

## Checklist


README.md
```diff
@@ -92,53 +92,80 @@ the swarm parameters for the test and the local url where the app is running
...
-branch instead of your original branch.
+Then the [deploy](https://github.com/CDCgov/trusted-intermediary/actions/workflows/internal-deploy.yml) will run.
+Remember that you now have the `internal` branch checked out locally.  If you make subsequent code changes, you will
+make them on the `internal` branch instead of your original branch.
+
+##### Dev
+
+The Dev environment is similar to the Internal environment but deploys to a CDC Azure Entra domain and subscription.  It
+is also meant to be the Wild West.  Dev deploys similarly to the Internal environment, but you interact with the
+`internal` branch.
```
sfradkin, line 130: is this supposed to be `dev` branch?

README.md
```diff
@@ -92,53 +92,80 @@ the swarm parameters for the test and the local url where the app is running
...
+
+##### Staging
```
sfradkin, line 132: Should we put a TDB on the staging environment until the new environment is created?  where is it deploying otherwise?  or... do we have 2 non-CDC environments now?

sfradkin (changes requested):

README.md
halprin: Oops. Good catch.

README.md
halprin: We have two non-CDC environments: `internal` and `staging`. Soon we'll update this section to remove the `non-` part.

halprin: And it's active today. It's been active since the beginning. The deletion I mentioned in Slack was just temporary. It's back in the same non-CDC Entrance domain as before.

README.md
halprin: Fixed.

sfradkin (approved):
Thanks for the change

PR #595, opened 2023-10-23
Update issue templates

scleary1cs:
Updated Spike template


scleary1cs:
> Did you mean to remove all the labels of the other templates? Also, I'm not entirely sure the newlines you injected into the YAML part of the Markdown will work.

No, not at all, the only template I changes was the Spike.

PR #597, opened 2023-10-23
Adding Template for Spikes

jcrichlake:
# Adding Spike template

Shawn was experiencing an issue where labels were getting removed from templates he hadn't edited for some reason.  Pushing the new template for Spikes so that he avoids impacting more files than intentded.

### DIFF
--- a/.github/ISSUE_TEMPLATE/spike.md
+++ b/.github/ISSUE_TEMPLATE/spike.md
@@ -1,0 +1,17 @@
+---
+name: Spike
+about: Describe this issue template's purpose here.
+title: ''
+labels: spike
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/spike.md
```diff
@@ -0,0 +1,17 @@
+---
+name: Spike
+about: Describe this issue template's purpose here.
```
halprin, line 3: There should probably be a description for when this template should be used.

halprin (changes requested):

halprin (approved):

Issue #487, opened 2023-08-23
Create `.env` file to store project environment variables


basiliskus:
# DevEx/OpEx

An `.env` file will make it easier to use and manage environment variables needed in the project, in a local dev environment 

## Tasks

- [x] Create PR for `task/eng-create_local_env_file` branch
- [x] Consider not relying in third party library to read `.env` files if custom handling is straightforward --> did some research and it's not straightforward
- [x] Make sure the sample `.env.example` has entries for all currently used env vars
- [x] Dynamically create a sample `.env` to use in the tests
- [x] Add tests

basiliskus:
We currently have a [draft PR](https://github.com/CDCgov/trusted-intermediary/pull/488) with work so far. Still pending, we need to find a good way to handle .env files for e2e test, unit test and app scopes. I moved .env files to resources folder, but we may move them elsewhere if more appropriate

PR #598, opened 2023-10-23
Add .env file handling for local environment variables

basiliskus:
Add .env file handling for local environment variables

## Issue

#487 

## Checklist


shared/src/main/java/gov/hhs/cdc/trustedintermediary/context/Environment.java
```diff
@@ -0,0 +1,21 @@
+package gov.hhs.cdc.trustedintermediary.context;
+
+public class Environment {
+
+    private Environment() {}
+
+    public static String getProperty(String key) {
+        if (DotEnv.isNull()) {
```
jcrichlake, line 8: Will DotEnv ever be null? It's getting initialized in the DotEnv class on instantiation

basiliskus: Good point. The .env file may not exist, but in that case the library will throw an exception instead of returning null, so we need to handle that

shared/src/main/java/gov/hhs/cdc/trustedintermediary/context/DotEnv.java
```diff
@@ -0,0 +1,17 @@
+package gov.hhs.cdc.trustedintermediary.context;
+
+import io.github.cdimascio.dotenv.Dotenv;
+
+class DotEnv {
```
halprin, line 5: Let's add some class level JavaDocs.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/context/Environment.java
```diff
@@ -0,0 +1,21 @@
+package gov.hhs.cdc.trustedintermediary.context;
+
+public class Environment {
```
halprin, line 3: Let's add some class level JavaDocs.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/context/Environment.java
jorg3lopez: We opted to not continue with the Environment class as the Dotenv library is doing some of that functionality behind the scenes.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/context/DotEnv.java
jorg3lopez: Excellent!

shared/src/main/java/gov/hhs/cdc/trustedintermediary/context/DotEnv.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * <p>The <code>load</code> method is primarily intended for testing purposes, allowing for custom
+ * Dotenv instances to be set. It can be useful when configuring specific environments for testing
+ * scenarios.
+ *
+ * @see <a href="https://github.com/cdimascio/dotenv-java">Dotenv Java GitHub Repository</a>
+ * @see io.github.cdimascio.dotenv.Dotenv
+ */
+class DotEnv {
+    private static Dotenv DOTENV;
```
jorg3lopez, line 22: This is no longer a final var, any thoughts on renaming it?

app/src/test/resources/.env.test
```diff
@@ -0,0 +1,3 @@
+ENV=local
```
jcrichlake, line 1: Is this file needed? I don't see it being loaded in the tests anywhere.

generate_env.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# Define the content of the .env file
+env_content="ENV=local\nKEY_VAULT-NAME=ti-key-vault-name\nREPORT_STREAM_URL_PREFIX=http://localhost:7071"
```
jcrichlake, line 4: Do we need a .env.example as well as a script to generate it?

jorg3lopez: Good catch. The .env.example file was made before the script. It is no longer needed.

app/src/test/resources/.env.test
jorg3lopez: File got deleted. Good one!

README.md
```diff
@@ -32,6 +32,16 @@ The additional requirements needed to contribute towards development are...
...
+### Generating .env File
+
+To set up the necessary environment variables, you can use the `generate_env.sh` script. This script will create a `.env` file in the resource folder with the required configuration. Follow these steps:
+
+1. Navigate to the project directory.
+
+2. Run the `generate_env.sh` script:
+
+   ```bash
+   ./generate_env.sh
```
halprin, line 44: This is code block isn't closed.

halprin (changes requested):
Just a small thing in the README file.  Other than that, this looks good.

README.md
jorg3lopez: Good catch, just added the closing.

halprin (approved):
Nice!

PR #599, opened 2023-10-23
ADR on the Project Structure

halprin:
# ADR on the Project Structure

Created an ADR describing our updated project structure.

## Issue

_None_.

## Checklist

### DIFF
--- a/adr/015-project-structure.md
+++ b/adr/015-project-structure.md
@@ -1,0 +1,57 @@
+# 15. Project Structure
+
+Date: 2023-10-23
+
+## Decision
+
+The overall project will consist of 4+ subprojects.
+
+- `app` - The entry point, where our HTTP or calling library is used, and plugins are initialized.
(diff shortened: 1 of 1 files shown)


sfradkin (approved):
Not sure if you'd like to include any information about the choice for using a plugin style architecture.  Do we have documentation somewhere on what is required to create another plugin and some example of why another plugin would be needed?

halprin:
> Not sure if you'd like to include any information about the choice for using a plugin style architecture. Do we have documentation somewhere on what is required to create another plugin and some example of why another plugin would be needed?

We have an existing [ADR](https://github.com/CDCgov/trusted-intermediary/blob/main/adr/006-domain-plugin.md) on the plugin paradigm.  It doesn't exactly call out the reasons one would make a new plugin for.  It alludes to the fact that you would make one for a separate business domain.

sfradkin:
Excellent

jcrichlake (commented):
lgtm

PR #601, opened 2023-10-24
Update Example FHIR OML

halprin:
# Update Example FHIR OML

We've updated what we do to the inbound FHIR ORM to make it be an OML.  It also looks like the inbound FHIR ORM has changed a little bit too.  Either way, `oml_message.json` should be updated to reflect the latest and greatest stuff we do to transform the ORM -> OML.

I also updated some of our sample HL7 ORM/OML messages to have the correct number of digits in the phone number.

## Issue

_None_.


examples/fhir/oml_message.json
```diff
@@ -931,29 +996,76 @@
...
+            "telecom": [
+              {
+                "extension": [
+                  {
+                    "url": "https://reportstream.cdc.gov/fhir/StructureDefinition/text",
+                    "valueString": "(763)555-5555"
+                  }
+                ],
+                "system": "phone",
+                "value": "(763) 555 555",
```
jcrichlake, line 1035: Missing a digit here

halprin: I thought it wasn't a bug in RS but instead it is because our original sample ORM/OML messages we were given were missing the last `5` digit.  I've updated those sample files to now have that last `5` digit.  I know my ORM that I sent into RS to make this updated file was also missing the final `5` digit.

jcrichlake (approved):

Issue #600, opened 2023-10-23
Clear (or Don't Store) the Cached Credentials if We Can't Login to ReportStream


halprin:
# DevEx/OpEx

Let's clear our private key from the cache, or not even cache it in the first place, if we fail to login to RS.

## Tasks

- [x] Change code.
- [x] Tests.

## Additional Context

We cache our private key after retrieving it.  This prevents us from having to continually call Azure for our secret each time we need to login to ReportStream (which is about every 5 minutes).  But, if the key is bad for some reason, we would fail to login to RS in perpetuity until we restart our application (which would clear the cache).  This happens because once we retrieve the secret, we cache it without ever clearing it.  Instead let's clear it from the cache, or not even cache it in the first place, if we fail to login to RS.

[Slack link](https://flexion.slack.com/archives/C055XTF22B0/p1696268515699579).

PR #604, opened 2023-10-24
600: Cached Our Private Key Only After Successful Auth with RS

halprin:
# Cached Our Private Key Only After Successful Auth with RS

Before this PR, we cached our private key after retrieving it no matter what happened.  This prevented us from having to continually call Azure for our secret each time we need to login to ReportStream (which is about every 5 minutes).  But, if the key was bad for some reason, we would fail to login to RS in perpetuity until we restarted our application (which would clear the cache).  This happened because once we retrieved the secret, we cached it without ever clearing it.

Now, we don't even cache our private key in the first place if we fail to login to RS.  We also only cache the key if it wasn't already cached.  This prevents us from continuously caching the key whenever we successfully authenticate with a key that was already cached.

## Issue

#600.

## Checklist

### DIFF
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamOrderSender.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamOrderSender.java
@@ -136,10 +136,11 @@ public class ReportStreamOrderSender implements OrderSender {
     protected String requestToken() throws UnableToSendOrderException {
         logger.logInfo("Requesting token from ReportStream");
 
+        String ourPrivateKey;
         String token;
 
         try {
-            String ourPrivateKey = retrievePrivateKey();
+            ourPrivateKey = retrievePrivateKey();
             String senderToken =
                     jwt.generateToken(
                             CLIENT_NAME,
@@ -156,6 +157,9 @@ public class ReportStreamOrderSender implements OrderSender {
                     "Error getting the API token from ReportStream", e);
         }
 
+        // only cache our private key if we successfully authenticate to RS
+        keyCache.put(OUR_PRIVATE_KEY_ID, ourPrivateKey);
+
         return token;
     }
 
@@ -166,7 +170,7 @@ public class ReportStreamOrderSender implements OrderSender {
         }
 
         key = secrets.getKey(OUR_PRIVATE_KEY_ID);
-        keyCache.put(OUR_PRIVATE_KEY_ID, key);
+
         return key;
     }
 
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamOrderSender.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamOrderSender.java
@@ -37,6 +37,9 @@ public class ReportStreamOrderSender implements OrderSender {
                     .map(urlPrefix -> urlPrefix.replace("https://", "").replace("http://", ""))
                     .orElse("");
 
+    private static final String OUR_PRIVATE_KEY_ID =
+            "trusted-intermediary-private-key-" + ApplicationContext.getEnvironment();
+
     private static final String CLIENT_NAME = "flexion.etor-service-sender";
 
     private String rsTokenCache;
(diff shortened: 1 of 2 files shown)


etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamOrderSender.java
```diff
@@ -131,42 +136,53 @@ protected String sendRequestBody(@Nonnull String json, @Nonnull String bearerTok
...
-        String senderToken = null;
-        String token = "";
-        String body;
-        Map<String, String> headers = Map.of("Content-Type", "application/x-www-form-urlencoded");
+        String ourPrivateKey;
+        String token;
+
         try {
-            senderToken =
+            ourPrivateKey = retrievePrivateKey();
```
jcrichlake, line 143: Should we bother with storing this in a variable since we are only using it once? Or do we want to shy on the side of readability?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamOrderSender.java
```diff
@@ -131,42 +136,53 @@ protected String sendRequestBody(@Nonnull String json, @Nonnull String bearerTok
...
                             CLIENT_NAME,
                             CLIENT_NAME,
                             CLIENT_NAME,
                             RS_DOMAIN_NAME,
                             300,
-                            retrievePrivateKey());
-            body = composeRequestBody(senderToken);
-            String rsResponse = client.post(RS_AUTH_API_URL, headers, body);
+                            ourPrivateKey);
+            String body = composeRequestBody(senderToken);
```
jcrichlake, line 152: I also don't see where we are not caching if we fail to log in to RS.  Maybe I'm just not seeing it in the diff

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamOrderSender.java
halprin: I assigned it to variable because a) yes, readability, and b) I need to pass the value into the `cachePrivateKeyIfNotCachedAlready` method below.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamOrderSender.java
```diff
@@ -131,42 +136,53 @@ protected String sendRequestBody(@Nonnull String json, @Nonnull String bearerTok
...
+            String rsResponse = client.post(RS_AUTH_API_URL, RS_AUTH_API_HEADERS, body);
             token = extractToken(rsResponse);
         } catch (Exception e) {
             throw new UnableToSendOrderException(
                     "Error getting the API token from ReportStream", e);
         }
+
+        // only cache our private key if we successfully authenticate to RS
+        cachePrivateKeyIfNotCachedAlready(ourPrivateKey);
```
halprin, line 161: @jcrichlake, this is where we cache the private key on success.  This is in relation to your [comment above](https://github.com/CDCgov/trusted-intermediary/pull/604/files#r1371859868).

jcrichlake (approved):

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamOrderSender.java
```diff
@@ -131,42 +136,53 @@ protected String sendRequestBody(@Nonnull String json, @Nonnull String bearerTok
...
         return token;
     }
 
     protected String retrievePrivateKey() throws SecretRetrievalException {
-        var senderPrivateKey =
-                "trusted-intermediary-private-key-" + ApplicationContext.getEnvironment();
-        String key = this.keyCache.get(senderPrivateKey);
+        String key = keyCache.get(OUR_PRIVATE_KEY_ID);
         if (key != null) {
             return key;
```
jorg3lopez, line 169: For record keeping. We still need to have a system in place that will change our private key in the cache, when it expires or when it is not valid.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamOrderSender.java
jorg3lopez, line 161: Great work here. I wonder since `cachePrivateKeyIfNotCachedAlready()` deals with only our private key, what are your thoughts on renaming it `cacheOurPrivatekeyIfNotCachedAlready()`?

jorg3lopez (approved):
Great work. I only had some comments but everything looks good.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamOrderSender.java
halprin: Exactly!  Thanks for calling that out and adding that item to our engineering task list channel.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamOrderSender.java
halprin: Fixed.

Issue #613, opened 2023-10-25
Use Existing Cache Paradigm for the RS Token


halprin:
# DevEx/OpEx

We cache the RS login token because it is valid for 5-ish minutes.  We currently cache it in a homegrown way in the `ReportStreamLabOrderSender` instead of using our existing `Cache` paradigm.

## Tasks

- [x] Convert homegrown version to our normal paradigm.

## Additional Context

Add any other context or screenshots about the work here.

PR #614, opened 2023-10-25
613: RS Token Cache Improvement

halprin:
# RS Token Cache Improvement

We had a homebrew caching mechanism for caching the RS token when we login.  This PR migrates to using the our `Caching` pattern.  This also will be more performant because we'll do less call to a locking method.

## Issue

#613.

## Checklist


jorg3lopez (approved):
Looks good, nice job!

basiliskus (approved):

PR #635, opened 2023-11-01
Use Existing "Research" Label

halprin:
# Use Existing "Research" Label

Our issue template was set to auto-assign `design/research` as the label, but we don't really use that label.  We mostly use the `research` label.  So, I codified this in our template.

## Issue

_None_.


scleary1cs (approved):
Great catch Peter.

JohnNKing (approved):
+1 to keeping our tag set clean

PR #646, opened 2023-11-06
Fixing load tests

jcrichlake:
When we updated the key management library we had to strip newlines off of the end of our jwts. We missed one in our load test python file.  This removes that newline and makes the tests run successfully.


halprin (approved):
🎸

tjohnson7021 (approved):

Issue #498, opened 2023-08-24
CDC Production Environment Created


halprin:
# Backlog Task

CDC Production Environment Created

## Completion Criteria

- [x] We have a production environment.
- [x] The team has account access to the production environment.
- [x] The full deployment to the CDC Production environment can happen programmatically when a GitHub release is created.

## Tasks

- [x] File CDC help desk ticket to create our prod environments.  Fill out the resource group template and attach. - @halprin 
- [x] Ask Jeremy, of ReportStream, what they do differently for production deploys, if anything.  Ask what, if any, "paperwork" they generate for a production deploy.  What, if any, sign-offs they get before a production deployment.  Document these. - @halprin 
- [x] Get answers to the above questions.
- [x] Decide whether we need to adopt these differences as well.
- [x] Build-out is complete by the CDC Help Desk.
- [x] Create the prod environment in our Terraform environment. - @halprin 
- [x] Create the CI/CD pipeline that deploys to production when a GitHub release is created on our repository. - @halprin 
- [x] Team to confirm they have access.
  - [x] @halprin 
  - [x] @basiliskus 
  - [x] @jcrichlake 
  - [x] @tjohnson7021 
  - [x] @jorg3lopez 
  - [x] @JohnNKing 
- [x] Prod environment renamed to `prd` in the Terraform - @halprin 
- [x] GitHub deployer whitelists our GitHub repository - @halprin 

## Other Notes

- ~Ticket `RITM0229496` for the CDC help desk to create the resource group in Azure.~
- Updated ticket number `CHG0032675`.
- They are starting on this on December 1, 2023.
- They estimate its completion by December 8, 2023.

halprin:
After [talking with the RS DevOps team](https://flexion.slack.com/archives/C048S3GPSQG/p1701200952580579), there isn't really much that is special about prod compared to staging.  Perhaps the only thing special are sign-offs (from a DevOps engineer) are required for a deploy to production.  I, at this time, do not believe we need someone on this team to sign-off on a deploy before it proceeds.  Maybe in the future, but we don't need red tape now.

halprin:
I created the ticket with the CDC help desk.  Ticket `RITM0229496`.

halprin:
Maggie reached out to me to fill out two additional forms.  There seemed to be a misunderstanding that we need a VM/server created, but we don't.  I said as much and asked if we still needed to fill out the additional forms.

halprin:
[...]

PR #693, opened 2023-11-30
498: Production Deploy on Release Publish

halprin:
# Production Deploy on Release Publish

Created a GitHub Action workflow that deploys to our production environment when a release is published in our GitHub repository.

## Issue

#498.


.github/workflows/prod-deploy.yml
```diff
@@ -0,0 +1,38 @@
+name: Production Deploy
+
+on:
+  release:
+    types:
+      - published
```
jcrichlake, line 6: Should this be on - published or on - released?  Seems like there are some small wrinkles as defined here 
https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=released#release

halprin: `released`
> A release was published, or a pre-release was changed to a release.

`publish`
> A release, pre-release, or draft of a release was published.

Hmmm, it's a little hard to piece together.  Do you think we're missing the part when a pre-release is promoted to a release?  But I'm also not sure what we'd use a pre-release for in the first place.

jcrichlake: I agree it's not the clearest difference.  The thing that raised my brow was a draft of a release being published, would that mean we would deploy a draft release when it's created in the UI?

halprin: Ah, I did look into that.  When a draft of a release happens, then only `created` happens.  When the draft goes out of draft, it is `published`.

jorg3lopez (approved):
Great work!

Issue #689, opened 2023-11-30
Invalid Proxy Exception on Request Execution After Apache Client 5.2.2 Upgrade


jorg3lopez:
# Bug

## Describe the Bug
After upgrading Apache HTTP Client from version 5.2.1 to 5.2.2, we encountered significant changes in how POST requests are built due to deprecated methods. A critical issue arose with this update, specifically related to header management (using org.apache.hc.core5.http.header). This was causing an internal exception that prevented TI from sending a payload to RS. Despite adjusting our code for these changes (during testing), we're consistently encountering an "Invalid Proxy Exception" when executing requests. This occurs even in the absence of any proxy settings.

## Impact
This bug is severely impacting our ability to execute HTTP requests reliably, causing disruptions in functionalities that depend on external API calls. The inability to bypass this exception is hindering our development and deployment processes.

## To Reproduce

1. Live test `ReportStreamOrderSender::getRSToken`, this will give you the import error
2. Once you fix the import error, debug the test case and you will get the `invalid proxy` error.
3. Delete local branch for live testing

## Expected Behavior

The expected behavior is that HTTP requests execute without throwing an Invalid Proxy Exception, especially in environments where no proxy settings are configured.

## Screenshots
Incompatible import
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOXdoQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--bf72dcba842c2d6e78363ed42066d0b0c74d11a0/image.png)

Invalid proxy exception
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOUloQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--13f627086434ca30c66c7bc6b4a5a9d5ceeb0bef/image.png)

## Logs

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOWNoQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f6a1a940dab47935480e0246e1b8db36dde053fc/image.png)

## Version

Apache HTTP Client version 5.2.2

[...]

PR #694, opened 2023-11-30
Apache Client Bug Fix

jorg3lopez:
# Apache Client Bug Fix
This PR addresses the issues caused by updating Apache client 5 from version 5.2.1 to 5.2.2. The main issues were incompatibility from the `org.apache.hc.core5.http.header` import and the client would throw an `invalid proxy` error whenever a post request was executed.

## Changes
- `org.apache.hc.core5.http.header` import has been deleted
- new code that deals with the header management has been employed.
- downgrade to version 5.2.1
- `convertMapToHeader(Map<String, String> headerMap)` helper method has been deleted along with test case

## Issue
#689 

## Checklist

- Build and all tests pass


e2e/build.gradle
```diff
@@ -11,8 +11,8 @@ java {
 
 dependencies {
     //client
-    implementation 'org.apache.httpcomponents.client5:httpclient5:5.2.2'
-    implementation 'org.apache.httpcomponents.client5:httpclient5-fluent:5.2.2'
+    implementation 'org.apache.httpcomponents.client5:httpclient5:5.2.1'
```
halprin, line 14: If we changed our code, do we need to also downgrade?  Or can we keep the latest version?

jorg3lopez: For the import, we can keep the old code. But once we upgrade again, we will have to make the same changes as the header import won't be supported. Overview of the [package](https://hc.apache.org/httpcomponents-core-5.2.x/current/httpcore5/apidocs/overview-summary.html)

shared/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/apache/ApacheClientTest.groovy
```diff
@@ -6,26 +6,6 @@ import spock.lang.Specification
...
 class ApacheClientTest extends Specification {
 
-    def "convertMapToHeader works"() {
-        given:
-        def headerMap = [
-            "key": "value",
-            "name": "dogCow",
-            "first": "last"
-        ]
-        def bhArr = [
```
jcrichlake, line 16: Would it be worth commenting these tests and having an item to uncomment them in the devex/opex backlog in the future when we get bumped to a bug free version?

basiliskus: @jcrichlake I think that's a good idea

jorg3lopez: Since the `org.apache.hc.core5.http.Header` is not supported in the 5.3.x version, we will not be utilizing the `convertMapToHeader(Map<String, String> headerMap)` helper function. This unit test was designed to test the  functionality of the helper function. Once the helper function gets deleted, there is no need for this test case.

jcrichlake: Gotcha, thanks for clarifying

jcrichlake (approved):
Looks good to me

halprin (changes requested):
I actually just looked.  [5.2.3](https://central.sonatype.com/artifact/org.apache.httpcomponents.client5/httpclient5/5.2.3/overview) was just released on November 28th.  Can we just upgrade to that to fix the issue?

jorg3lopez:
Version 5.2.3 is out and the Header import works with it.

Issue #672, opened 2023-11-21
Metadata: Hospital Order Confirmation


JohnNKing:
# Story
As a hospital, so I can confirm my order has been received by ReportStream, I need to be able to view basic information about all orders I've submitted recently.

## Acceptance Criteria
- [ ] The hospital is able to view the unique identifier, sender, receiver, time received, and a hash of their original order.
- [ ] The order metadata is available in the form of FHIR OperationalOutcome
- [ ] The hospital is able to view all orders submitted by them over the past 90 days.

### Research
- [ ] *Research work needed to complete the story*
- [ ] Foundational: *High-level research that will support this and future efforts*

#### Foundational

- [x] Decide on the Q&C groups.
  -  Azure Storage/Local File: Basilio, Peter.
  -  Database: Jeff, Tiff, Jorge.
- [x] Define an interface(s) that's agnostic to how we store the metadata (e.g. in memory, in a file, in a Azure blob storage, or a database). - do as a group.
- [x] Create a local file implementation.
  - [x] Change implementation to use an app path instead of temp path @basiliskus 
- [x] Create an Azure blob store implementation.
  - [x] Request required access for CDC's azure environments (dev, staging, prod)  @basiliskus 
  - [x] Check if the required access was given correctly. - @halprin 
  - [x] Return an empty `Optional` if no metadata is found given a unique ID.  @basiliskus 
- [ ] Create a database implementation.
  - [x] Spin up database locally via Docker
  - [x] Create a table for this story
  - [x] Update Intermediary to use the pgJDBC driver to upsert to database 
  - [x] Connect database to intermediary
  - [x] Find out if we need to request anything specific in Azure
  - [x] Set up Select method in Dao 
  - [x] Update Terraform scripts so that the DB is deployable and can interact with the application
  - [x] Extract local parameters to secrets/env file
  - [x] Return an empty `Optional` if no metadata is found given a unique ID.
  - [ ] Below Tasks are for if the db path is selected
    - [ ] Figure out the full table schema and make sure it's normalized
    - [ ] Implement db migration serivce, or something similar for schema migrations updates
    - [ ] Verify if connection pool is set up or not and if we need it
    - [ ] Do we need to limit roles on application's db user to keep them from modifying db schema and limiting what tables can be read?
    - [x] Setup Azure snapshots for backups
    - [ ] Update intermediary to support connection pooling

--

- [x] Decide what our unique identifier is.
[...]

PR #701, opened 2023-12-04
672: Write Dummy Metadata

halprin:
# Write Dummy Metadata

- Write some sample partner-facing metadata when we are sent an order.
- Converted the partner-facing metadata to use an `Instant` instead of a `LocalTime`.

## Issue

#672.


JohnNKing (approved):
Looks good here. I assume we'll eventually want to rename MetricMetaData to MetricMetadata, but that was pre-existing.

PR #726, opened 2023-12-13
Update PartnerMetadataStorage.readMetadata to return Optional<PartnerMetadata>

basiliskus:
# Update PartnerMetadataStorage.readMetadata to return Optional<PartnerMetadata>

We're missing a failure path in the case the metadata was not found, so we can return a `404` error instead of `400`. For this purpose I'm changing the return value of `readMetadata` to `Optional<PartnerMetadata>` so we can return and handle an empty value when the metadata is not found

## Issue

#672 

## Checklist

### DIFF
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/localfile/FilePartnerMetadataStorage.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/localfile/FilePartnerMetadataStorage.java
@@ -12,6 +12,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.attribute.FileAttribute;
 import java.nio.file.attribute.PosixFilePermissions;
+import java.util.Optional;
 import javax.inject.Inject;
 
 /** Implements the {@link PartnerMetadataStorage} using local files. */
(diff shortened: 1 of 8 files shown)


etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/localfile/FilePartnerMetadataStorage.java
```diff
@@ -42,11 +43,18 @@ public static FilePartnerMetadataStorage getInstance() {
     }
 
     @Override
-    public PartnerMetadata readMetadata(final String uniqueId) throws PartnerMetadataException {
+    public Optional<PartnerMetadata> readMetadata(final String uniqueId)
+            throws PartnerMetadataException {
         Path filePath = getFilePath(uniqueId);
         try {
+            if (!Files.exists(filePath)) {
+                logger.logInfo("Metadata file not found: " + filePath);
```
halprin, line 51: Info or warning?

basiliskus: Good catch. Fixed

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/localfile/FilePartnerMetadataStorage.java
```diff
@@ -42,11 +43,18 @@ public static FilePartnerMetadataStorage getInstance() {
     }
 
     @Override
-    public PartnerMetadata readMetadata(final String uniqueId) throws PartnerMetadataException {
+    public Optional<PartnerMetadata> readMetadata(final String uniqueId)
+            throws PartnerMetadataException {
         Path filePath = getFilePath(uniqueId);
         try {
+            if (!Files.exists(filePath)) {
+                logger.logWarning("Metadata file not found: " + filePath);
```
halprin, line 51: Let's update the `logWarning` to accept parameters so we can use `{}`.

basiliskus: Done

halprin (approved):

PR #729, opened 2023-12-14
Use ReportId to identify metadata for testing purposes

basiliskus:
# Use ReportId to identify metadata for testing purposes

While waiting for [this ReportStream ticket](https://github.com/CDCgov/prime-reportstream/issues/12624) in order to get a submission id, we'll temporarily use the `RecordId` as a placeholder to work on testing

## Issue

#672 

## Checklist

### DIFF
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
@@ -126,8 +126,17 @@ public class EtorDomainRegistration implements DomainConnector {
     DomainResponse handleOrders(DomainRequest request) {
         Order<?> orders;
 
+        String submissionId = request.getHeaders().get("RecordId");
+        // TODO: need to verify what we should do if we don't receive the RecordId header or if it's
+        // empty
+        // For now, returning a 400 error
+        if (submissionId == null) {
+            String errorMessage = "Missing required header: RecordId";
+            logger.logError(errorMessage);
+            return domainResponseHelper.constructErrorResponse(400, errorMessage);
+        }
+
         try {
-            String submissionId = request.getHeaders().get("RecordId");
             orders = orderController.parseOrders(request);
             sendOrderUseCase.convertAndSend(orders, submissionId);
         } catch (FhirParseException e) {
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
@@ -127,8 +127,9 @@ public class EtorDomainRegistration implements DomainConnector {
         Order<?> orders;
 
         String submissionId = request.getHeaders().get("RecordId");
-        if (submissionId == null) {
-            logger.logError("Missing required header: RecordId");
+        if (submissionId == null || submissionId.isEmpty()) {
+            submissionId = null;
+            logger.logError("Missing required header or empty: RecordId");
         }
 
         try {
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
@@ -127,13 +127,8 @@ public class EtorDomainRegistration implements DomainConnector {
         Order<?> orders;
 
         String submissionId = request.getHeaders().get("RecordId");
-        // TODO: need to verify what we should do if we don't receive the RecordId header or if it's
-        // empty
-        // For now, returning a 400 error
         if (submissionId == null) {
-            String errorMessage = "Missing required header: RecordId";
-            logger.logError(errorMessage);
-            return domainResponseHelper.constructErrorResponse(400, errorMessage);
+            logger.logError("Missing required header: RecordId");
         }
 
         try {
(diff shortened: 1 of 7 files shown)


etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
```diff
@@ -126,9 +126,19 @@ DomainResponse handleDemographics(DomainRequest request) {
     DomainResponse handleOrders(DomainRequest request) {
         Order<?> orders;
 
+        String submissionId = request.getHeaders().get("RecordId");
+        // TODO: need to verify what we should do if we don't receive the RecordId header or if it's
+        // empty
+        // For now, returning a 400 error
+        if (submissionId == null) {
```
basiliskus, line 133: Need to discuss what to do when we don't receive the RecordId

basiliskus: After discussion with the team, decided not break the flow and log it as an error

halprin (approved):

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
```diff
@@ -126,9 +126,14 @@ DomainResponse handleDemographics(DomainRequest request) {
     DomainResponse handleOrders(DomainRequest request) {
         Order<?> orders;
 
+        String submissionId = request.getHeaders().get("RecordId");
+        if (submissionId == null) {
+            logger.logError("Missing required header: RecordId");
```
jcrichlake, line 131: Should we also set submissionId to a string that says null in this case? Not sure if it will save as null or empty

basiliskus: I checked that it will return null if either RecordId is not present or the value is empty. Not sure if this is what you're referring to

jcrichlake: I thought we were going to write this submissionId eventually into the file, will it error if we pass a null value in for the String?  Or would it be better if we made the string "null" so that it is able to print to the file

basiliskus: Good point. We need to check for null when trying to write to the file. I don't think we want to pass a "null" string as the file could be overwritten by different submissions and we wouldn't have certainty that the metadata is unique to a specific order anymore

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/SendOrderUseCase.java
```diff
@@ -22,10 +22,11 @@ public static SendOrderUseCase getInstance() {
...
-    public void convertAndSend(final Order<?> order) throws UnableToSendOrderException {
+    public void convertAndSend(final Order<?> order, String submissionId)
+            throws UnableToSendOrderException {
         var partnerMetadata =
                 new PartnerMetadata(
-                        "uniqueId", "senderName", "receiverName", Instant.now(), "abcd");
+                        submissionId, "senderName", "receiverName", Instant.now(), "abcd");
         try {
             partnerMetadataStorage.saveMetadata(partnerMetadata);
         } catch (PartnerMetadataException e) {
```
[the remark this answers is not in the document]
halprin: Oh, good catch.  Uh yes.  We can probably choose to edit this here or do it in the follow-on PR when we edit the usecase to call ReportStream's history API, etc.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
basiliskus: That's what I was thinking as well @halprin: check for null in SendOrderUseCase and skip if null

basiliskus: Made the change in this PR

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/SendOrderUseCase.java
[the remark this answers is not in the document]
basiliskus: Fixed in this PR

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
jorg3lopez: When looking into `DomainRequest.java`, `getHeaders()` returns a Map<String, String>. Following this information, if we search for a key that doesn't exist, it will return null. If we search for a key that has an empty value, it will return empty. What are your thoughts on adding `|| submissionId.isEmpty()` to the if statement so it will check for null and for empty?

halprin: That's a good point.  We should check for empty because that would be the same as if RS provided the header but didn't provide a value.  And if it is empty, we should probably set `submissionId` to `null` so we can make the subsequent logic in the usecase easier.

basiliskus: I tested it with an empty value for the `RecordId` header and the `request.getHeaders().get("RecordId")` call returned null. I'll see if I can find the docs or the code to validate this assumption

halprin: Oh, so maybe we don't need this the extra part of the `if` statement.

jorg3lopez (approved):
Good work!

tjohnson7021 (approved):

PR #730, opened 2023-12-14
Update FilePartnerMetadataStorage to use app path instead of temp path

basiliskus:
# Update FilePartnerMetadataStorage to use app path instead of temp path

Changed because we need better persistence of the metadata file 

## Issue

#672

### DIFF
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/localfile/FilePartnerMetadataStorage.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/localfile/FilePartnerMetadataStorage.java
@@ -10,8 +10,7 @@ import gov.hhs.cdc.trustedintermediary.wrappers.formatter.TypeReference;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.attribute.FileAttribute;
-import java.nio.file.attribute.PosixFilePermissions;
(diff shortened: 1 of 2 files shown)


etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/localfile/FilePartnerMetadataStorage.java
```diff
@@ -23,14 +22,13 @@ public class FilePartnerMetadataStorage implements PartnerMetadataStorage {
...
-    private static final Path metadataTempDirectory;
+    private static final Path metadataDirectory;
 
     static {
         try {
-            FileAttribute<?> onlyOwnerAttrs =
-                    PosixFilePermissions.asFileAttribute(
-                            PosixFilePermissions.fromString("rwx------"));
-            metadataTempDirectory = Files.createTempDirectory("metadata", onlyOwnerAttrs);
+            Path projectRoot = Paths.get(System.getProperty("user.dir"));
```
basiliskus, line 29: Is this a reliable way to get the root path?

basiliskus: After discussing with the team, decided to change this implementation to create a folder in the system temp path to put the files in

halprin (approved):
Lookin' good!

PR #739, opened 2023-12-20
Refactor setHeaders to make non case-sensitive

basiliskus:
# Refactor setHeaders to make non case-sensitive

The implementation to get headers we're using from Javalin is case-sensitive, while HTTP headers are non case-sensitive. We decided to transform the header names to lowercase while setting them to avoid this issue

## Issue

#672 

## Checklist


halprin (approved):
Ship it.

PR #744, opened 2023-12-26
Call the history API for outbound orders

basiliskus:
# Call the history API for outbound orders

- Changed OrderSender.sendOrder to return the sent submissionId (as optional)
- Refactored `ReportStreamOrderSender` and created a more generic `ReportStreamEndpointClient`
- Added call to RS history API to get the receiver name and update metadata 
- Changes to `PartnerMetadata` record
  -  Added `sentSubmissionId` and renamed `uniqueId` to `receivedSubmissionId`
  - Overloaded constructor and added `withSentSubmissionId` and `withReceiver` to update record
-  Changes to `HttpClient`
  - Added `get` http method and implemented in `ApacheClient` 

## Issue

#672 

## Checklist

### DIFF
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
@@ -22,8 +22,8 @@ public class PartnerMetadataOrchestrator {
     private static final PartnerMetadataOrchestrator INSTANCE = new PartnerMetadataOrchestrator();
 
     @Inject PartnerMetadataStorage partnerMetadataStorage;
-    @Inject private ReportStreamEndpointClient rsclient;
-    @Inject private Formatter formatter;
+    @Inject ReportStreamEndpointClient rsclient;
+    @Inject Formatter formatter;
 
     public static PartnerMetadataOrchestrator getInstance() {
         return INSTANCE;
@@ -47,8 +47,7 @@ public class PartnerMetadataOrchestrator {
         partnerMetadataStorage.saveMetadata(partnerMetadata);
     }
 
-    public void updateMetadataForSentOrder(
-            String receivedSubmissionId, String sentSubmissionId, Order<?> order)
+    public void updateMetadataForSentOrder(String receivedSubmissionId, String sentSubmissionId)
             throws PartnerMetadataException {
 
         String receiver;
@@ -78,7 +77,7 @@ public class PartnerMetadataOrchestrator {
         return partnerMetadataStorage.readMetadata(submissionId);
     }
 
-    private String getReceiverName(String clientResponse) throws FormatterProcessingException {
+    String getReceiverName(String clientResponse) throws FormatterProcessingException {
         Map<String, Object> responseObject =
                 formatter.convertJsonToObject(clientResponse, new TypeReference<>() {});
         Object destinationsObj = responseObject.get("destinations");
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
@@ -31,7 +31,7 @@ public class PartnerMetadataOrchestrator {
 
     private PartnerMetadataOrchestrator() {}
 
-    public void updateMetadataForReceivedOrder(String submissionId, Order<?> order)
+    public void updateMetadataForReceivedOrder(String receivedSubmissionId, Order<?> order)
             throws PartnerMetadataException {
         // will call the RS history API given the submissionId, currently blocked by:
         // https://github.com/CDCgov/prime-reportstream/issues/12624
@@ -43,7 +43,7 @@ public class PartnerMetadataOrchestrator {
         Instant timeReceived = Instant.now();
         String hash = String.valueOf(order.hashCode());
         PartnerMetadata partnerMetadata =
-                new PartnerMetadata(submissionId, sender, timeReceived, hash);
+                new PartnerMetadata(receivedSubmissionId, sender, timeReceived, hash);
         partnerMetadataStorage.saveMetadata(partnerMetadata);
     }
 
@@ -66,7 +66,7 @@ public class PartnerMetadataOrchestrator {
         partnerMetadataStorage.saveMetadata(partnerMetadata);
     }
 
-    public Optional<PartnerMetadata> getMetadata(String submissionId)
+    public Optional<PartnerMetadata> getMetadata(String receivedSubmissionId)
             throws PartnerMetadataException {
         // call the metadata storage to get the metadata.
         // check if the receiver is filled out, and if it isn't, call the RS history API to get the
@@ -74,12 +74,12 @@ public class PartnerMetadataOrchestrator {
         // if had to call the history API, extract the receiver and call the metadata storage to
         // save the metadata with the receiver added.
         // return the metadata.
-        return partnerMetadataStorage.readMetadata(submissionId);
+        return partnerMetadataStorage.readMetadata(receivedSubmissionId);
     }
 
-    String getReceiverName(String clientResponse) throws FormatterProcessingException {
+    String getReceiverName(String responseBody) throws FormatterProcessingException {
         Map<String, Object> responseObject =
-                formatter.convertJsonToObject(clientResponse, new TypeReference<>() {});
+                formatter.convertJsonToObject(responseBody, new TypeReference<>() {});
         Object destinationsObj = responseObject.get("destinations");
         if (!(destinationsObj instanceof ArrayList<?> destinationsList)) {
             throw new FormatterProcessingException(
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
@@ -9,6 +9,7 @@ import gov.hhs.cdc.trustedintermediary.wrappers.formatter.TypeReference;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import javax.inject.Inject;
 
@@ -50,6 +51,13 @@ public class PartnerMetadataOrchestrator {
     public void updateMetadataForSentOrder(String receivedSubmissionId, String sentSubmissionId)
             throws PartnerMetadataException {
 
+        PartnerMetadata partnerMetadata =
+                partnerMetadataStorage.readMetadata(receivedSubmissionId).orElseThrow();
+        if (!Objects.equals(partnerMetadata.sentSubmissionId(), sentSubmissionId)) {
+            partnerMetadata = partnerMetadata.withSentSubmissionId(sentSubmissionId);
+            partnerMetadataStorage.saveMetadata(partnerMetadata);
+        }
+
         String receiver;
         try {
             String bearerToken = rsclient.getRsToken();
@@ -60,9 +68,7 @@ public class PartnerMetadataOrchestrator {
                     "Unable to retrieve metadata from RS history API", e);
         }
 
-        PartnerMetadata partnerMetadata =
-                partnerMetadataStorage.readMetadata(receivedSubmissionId).orElseThrow();
-        partnerMetadata = partnerMetadata.withSentSubmissionFields(sentSubmissionId, receiver);
+        partnerMetadata = partnerMetadata.withReceiver(receiver);
         partnerMetadataStorage.saveMetadata(partnerMetadata);
     }
 
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
@@ -83,6 +83,20 @@ public class PartnerMetadataOrchestrator {
     }
 
     String getReceiverName(String responseBody) throws FormatterProcessingException {
+        // Before we can get the organization_id and service from the response,
+        // we need to validate that the response has the correct structure
+        // the expected json structure is:
+        // {
+        //    ...
+        //    "destinations" : [ {
+        //        ...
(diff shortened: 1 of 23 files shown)


etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
```diff
@@ -15,45 +23,89 @@ public class PartnerMetadataOrchestrator {
...
-        // time received.
+        // will call the RS history API given the submissionId, currently blocked by:
+        // https://github.com/CDCgov/prime-reportstream/issues/12624
+        // from the response, extract the "sender" and "timestamp" (timeReceived)
         // we will calculate the hash.
         // then we call the metadata storage to save this stuff.
 
+        String sender = "unknown";
+        Instant timeReceived = Instant.now();
+        String hash = String.valueOf(order.hashCode());
```
jcrichlake, line 45: This is not a secure method of hashing the order which is sensitive info.  We should figure out the type of hash we want to use and implement something like the below link based on the algorithm. 
[hashing](https://reflectoring.io/creating-hashes-in-java/)

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
```diff
@@ -15,45 +23,89 @@ public class PartnerMetadataOrchestrator {
...
+    public void updateMetadataForSentOrder(String receivedSubmissionId, String sentSubmissionId)
             throws PartnerMetadataException {
-        // call the metadata storage and add the sent order's submission ID to the existing metadata
-        // entry
-        // PartnerMetadata may need to be updated to store both the received order's submission ID
-        // _and_ the sent order's submission ID.
+
+        PartnerMetadata partnerMetadata =
+                partnerMetadataStorage.readMetadata(receivedSubmissionId).orElseThrow();
+        if (!Objects.equals(partnerMetadata.sentSubmissionId(), sentSubmissionId)) {
```
jcrichlake, line 56: You should be able to refactor this into something like 

```suggestion
        if (sentSubmissionId != null && !sentSubmissionId.equals`(partnerMetadata.sentSubmissionId()) {
```

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
basiliskus: Good point, though I think this hash will be more of a placeholder for now. RS should be the one creating a sending the hash to us, since they'll have the original order

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
basiliskus: Thanks, I'll use your suggestion to compare. We don't need to check sentSubmissionId for null since we check in `saveSentOrderSubmissionId`.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
```diff
@@ -15,45 +22,89 @@ public class PartnerMetadataOrchestrator {
...
+                    "destinations is not an ArrayList", new Exception());
+        }
+
+        if (destinationsList.isEmpty()
+                || !(destinationsList.get(0) instanceof Map<?, ?> destination)) {
+            throw new FormatterProcessingException(
+                    "First item in destinations is not a Map", new Exception());
+        }
+
+        var organizationId = destination.get("organization_id");
```
jcrichlake, line 100: Can we make both of these Strings? Just more explicit instead of relying on the jdk to figure out the type

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
```diff
@@ -15,45 +22,89 @@ public class PartnerMetadataOrchestrator {
...
+    String getReceiverName(String responseBody) throws FormatterProcessingException {
+        Map<String, Object> responseObject =
+                formatter.convertJsonToObject(responseBody, new TypeReference<>() {});
+        Object destinationsObj = responseObject.get("destinations");
+        if (!(destinationsObj instanceof ArrayList<?> destinationsList)) {
+            throw new FormatterProcessingException(
+                    "destinations is not an ArrayList", new Exception());
+        }
+
+        if (destinationsList.isEmpty()
```
jcrichlake, line 94: Can we put some comments on these if blocks? It's a little unclear what these are doing

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/SendOrderUseCase.java
```diff
@@ -22,44 +22,44 @@ public static SendOrderUseCase getInstance() {
...
+
+        saveSentOrderSubmissionId(receivedSubmissionId, sentSubmissionId);
     }
 
-    private void savePartnerMetadataForReceivedOrder(String submissionId, final Order<?> order) {
-        if (submissionId == null) {
+    private void savePartnerMetadataForReceivedOrder(
+            String receivedSubmissionId, final Order<?> order) {
+        if (receivedSubmissionId == null) {
             return;
```
jcrichlake, line 44: Do we need to log in these null SubmissionId scenarios? Just so we have awareness?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/SendOrderUseCase.java
```diff
@@ -22,44 +22,44 @@ public static SendOrderUseCase getInstance() {
...
+            logger.logError(
+                    "Unable to save metadata for receivedSubmissionId " + receivedSubmissionId, e);
         }
     }
 
-    private void saveSentOrderSubmissionId(
-            String receivedSubmissionId, String sentSubmissionId, final Order<?> order) {
+    private void saveSentOrderSubmissionId(String receivedSubmissionId, String sentSubmissionId) {
         if (sentSubmissionId == null || receivedSubmissionId == null) {
             return;
```
jcrichlake, line 57: Same question here, do we want to log in this case?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/localfile/FilePartnerMetadataStorage.java
```diff
@@ -65,14 +65,18 @@ public Optional<PartnerMetadata> readMetadata(final String uniqueId)
...
     @Override
     public void saveMetadata(final PartnerMetadata metadata) throws PartnerMetadataException {
-        Path metadataFilePath = getFilePath(metadata.uniqueId());
+        Path metadataFilePath = getFilePath(metadata.receivedSubmissionId());
         try {
             String content = formatter.convertToJsonString(metadata);
             Files.writeString(metadataFilePath, content);
-            logger.logInfo("Saved metadata for " + metadata.uniqueId() + " to " + metadataFilePath);
+            logger.logInfo(
+                    "Saved metadata for "
```
jcrichlake, line 73: I think you want to parameterize the inputs for log statements.  So something like 

```suggestion
                    logger.logInfo("Saved metadata for {} to {}", metadata.receivedSubmissionId(), metadataFilePath);
```

It's cleaner and more efficient

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamEndpointClient.java
```diff
@@ -0,0 +1,195 @@
...
+import javax.inject.Inject;
+
+public class ReportStreamEndpointClient {
+    private static final String RS_URL_PREFIX_PROPERTY = "REPORT_STREAM_URL_PREFIX";
+    private static final String RS_DOMAIN_NAME =
+            Optional.ofNullable(ApplicationContext.getProperty(RS_URL_PREFIX_PROPERTY))
+                    .map(urlPrefix -> urlPrefix.replace("https://", "").replace("http://", ""))
+                    .orElse("");
+    private static final String RS_WATERS_API_URL =
+            ApplicationContext.getProperty(RS_URL_PREFIX_PROPERTY) + "/api/waters";
```
jcrichlake, line 30: Can we put this property into a variable so we don't have to keep fetching it from the map over and over?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamEndpointClient.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    public String requestHistoryEndpoint(@Nonnull String submissionId, @Nonnull String bearerToken)
+            throws ReportStreamEndpointClientException {
+        logger.logInfo("Requesting history API from ReportStream");
+
+        Map<String, String> headers = Map.of("Authorization", "Bearer " + bearerToken);
+
+        try {
+            var url = RS_HISTORY_API_URL.replace("{id}", submissionId);
```
jcrichlake, line 117: Can we use the specific type (I assume String) instead of var

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamEndpointClient.java
```diff
@@ -0,0 +1,195 @@
...
+            return client.get(url, headers);
+        } catch (HttpClientException e) {
+            throw new ReportStreamEndpointClientException(
+                    "Error GETing the history from ReportStream", e);
```
jcrichlake, line 121: I think this should be GETting

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamEndpointClient.java
```diff
@@ -0,0 +1,195 @@
...
+
+    public String getRsToken() throws ReportStreamEndpointClientException {
+        logger.logInfo("Looking up ReportStream token");
+
+        var token = cache.get(RS_TOKEN_CACHE_ID);
+
+        if (token != null && isValidToken(token)) {
+            logger.logDebug("valid cache token");
+            return token;
+        }
```
jcrichlake, line 147: Would you need an else condition here to remove an invalid token? Otherwise in theory this map will be growing forever with expired tokens

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamOrderSender.java
```diff
@@ -64,141 +32,43 @@ public static ReportStreamOrderSender getInstance() {
...
-        key = secrets.getKey(OUR_PRIVATE_KEY_ID);
-
-        return key;
-    }
-
-    void cacheOurPrivateKeyIfNotCachedAlready(String privateKey) {
-        String key = cache.get(OUR_PRIVATE_KEY_ID);
-        if (key != null) {
-            return;
+            var rsResponse =
```
jcrichlake, line 64: String or Map?

jcrichlake (changes requested):

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
basiliskus: Will do

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
basiliskus: Would it help if I add a comment with the expected json format that require all these validations?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/SendOrderUseCase.java
basiliskus: Yes, good point

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/localfile/FilePartnerMetadataStorage.java
basiliskus: Good catch, I'll do that

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
jcrichlake: Mostly around what the expected business case is, like we have a destinationsObject that we check to make sure it's a list and then we have to check on the List? Why are we getting these?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamOrderSender.java
basiliskus: Map<String, Object>. Will change

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
basiliskus: It is mostly to avoid null pointer exceptions in case the expected values are not there. Would it be better just to catch the null pointer exception and re throwing as FormatterProcessingException instead of making all these checks?

jcrichlake: Null checks are pretty widely accepted as the preferred method for handling/avoiding null pointer exceptions.  We can leave this as is for now though

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamEndpointClient.java
jcrichlake: Sounds good!

jcrichlake (approved):

PR #746, opened 2023-12-28
Call the history API for inbound orders

basiliskus:
# Call the history API for inbound orders

Implemented updateMetadataForReceivedOrder and getMetadata

Currently blocked by #749 

## Issue

#672 

## Checklist


tjohnson7021 (approved):

tjohnson7021 (approved):

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
```diff
@@ -33,15 +35,41 @@ private PartnerMetadataOrchestrator() {}
...
+        Instant timeReceived;
+        String hash;
+        try {
+            String bearerToken = rsclient.getRsToken();
+            String responseBody =
+                    rsclient.requestHistoryEndpoint(receivedSubmissionId, bearerToken);
+            Map<String, Object> responseObject =
+                    formatter.convertJsonToObject(responseBody, new TypeReference<>() {});
+
+            var senderObj = responseObject.get("sender");
```
jcrichlake, line 54: Same here as on the other, I'd just make these Strings

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
```diff
@@ -67,19 +99,23 @@ public void updateMetadataForSentOrder(String receivedSubmissionId, String sentS
...
+        PartnerMetadata partnerMetadata =
+                partnerMetadataStorage.readMetadata(receivedSubmissionId).orElseThrow();
+
+        if (partnerMetadata.receiver() == null && partnerMetadata.sentSubmissionId() != null) {
+            logger.logInfo("Receiver name not found in metadata, looking up from RS history API");
+            updateMetadataForSentOrder(receivedSubmissionId, partnerMetadata.sentSubmissionId());
+            return partnerMetadataStorage.readMetadata(receivedSubmissionId);
+        }
+
+        return Optional.of(partnerMetadata);
```
jcrichlake, line 118: Does this ever return an Optional.empty? If not should it still return an optional?

basiliskus: Yes, it should return Optional.empty if the stored metadata was not found

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
basiliskus: Done

jcrichlake (approved):

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
```diff
@@ -71,19 +98,23 @@ public void updateMetadataForSentOrder(String receivedSubmissionId, String sentS
...
             throws PartnerMetadataException {
-        // call the metadata storage to get the metadata.
-        // check if the receiver is filled out, and if it isn't, call the RS history API to get the
-        // receiver.
-        // if had to call the history API, extract the receiver and call the metadata storage to
-        // save the metadata with the receiver added.
-        // return the metadata.
-        return partnerMetadataStorage.readMetadata(receivedSubmissionId);
+        PartnerMetadata partnerMetadata =
+                partnerMetadataStorage.readMetadata(receivedSubmissionId).orElseThrow();
```
halprin, line 109: The `.orElseThrow()` throws a `NoSuchElementException` which is an unchecked `RuntimeException`.  We should instead throw our `PartnerMetadataException`, right?  If that's right, then this would turn into...

```suggestion
                partnerMetadataStorage.readMetadata(receivedSubmissionId).orElseThrow(() -> new PartnerMetadataException("Can't update metadata for sent order, no metadata exists for ID " + receivedSubmissionId));
```

halprin: Also, this should apply to the other `orElseThrow` in the `getMetadata` method.

halprin (changes requested):
Some comments on the thrown exception.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
jorg3lopez: Good catch. I think this was a something that we thought about before the PartnerMetadataException was made, that we  needed a custom exception.

halprin:
I broke it!  ;)  I think we need a new constructor added to the `PartnerMetadataException` that only takes a `String` argument.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/PartnerMetadataOrchestrator.java
basiliskus: I'll refactor `updateMetadataForSentOrder` to pass PartnerMetadata so we don't need to read the metadata again when we call it

halprin (approved):

PR #747, opened 2023-12-29
Add retry mechanism to call RS history API

basiliskus:
# Add retry mechanism to call RS history API

- Added `SyncRetryTask` as util that we can reuse

## Issue

#672 

## Checklist


halprin (changes requested):
Hmmm.  There's a few possible issues with the current approach.

1. There doesn't seem to be support for concurrent calls to `RetryTask#scheduleRetry`.  E.g. there's two tasks that are retrying, then one finishes successfully, it kills the scheduler, and the second task is no longer being executed.
2. Looks like we couldn't call `RetryTask#scheduleRetry` a second time after the first time since the `scheduler` is shutdown and never started back up again.
3. Is the purpose of this retry mechanism to be synchronous or asynchronous?  This looks asynchronous to me, but do we need it to be synchronous?  For example, if there's still a failure after the last retry, do we want to bubble up that last exception so it is thrown to the top level call to `RetryTask#scheduleRetry`?

halprin:
Another thought is using an existing third-party library for this that already has robust retry capability.  E.g Spring Retry (you can use this without bringing all of Spring into our application).  This would get you the benefit of not having to write all the intricacies of this and give you the benefits of exponential backoff with jitter.  But then again, maybe it's easier to write a simple for loop (but then it's all synchronous and maybe you don't want that?).  So anyway, those are some thoughts.

basiliskus:
> 1. There doesn't seem to be support for concurrent calls to `RetryTask#scheduleRetry`.  E.g. there's two tasks that are retrying, then one finishes successfully, it kills the scheduler, and the second task is no longer being executed.
> 2. Looks like we couldn't call `RetryTask#scheduleRetry` a second time after the first time since the `scheduler` is shutdown and never started back up again.
> 3. Is the purpose of this retry mechanism to be synchronous or asynchronous?  This looks asynchronous to me, but do we need it to be synchronous?  For example, if there's still a failure after the last retry, do we want to bubble up that last exception so it is thrown to the top level call to `RetryTask#scheduleRetry`?

Good points. I'd need to do some more research on when and how ScheduledThreadPool is expected to be shut down. Have't used it before so I'm not very familiar. And you're right, this is async but that will create some issues. I'd need to put more though into this because it was a quick idea that I meant to try after the other PRs were done

halprin:
I should ask, what was the thinking behind needing retries?  Is the history API flaky or something?  We should consider about using this when we send our orders to RS (not in this PR).

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/SendOrderUseCase.java
```diff
@@ -3,18 +3,24 @@
...
 import gov.hhs.cdc.trustedintermediary.wrappers.MetricMetadata;
+import java.util.concurrent.Callable;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import javax.inject.Inject;
 
 /** The overall logic to receive, convert to OML, and subsequently send a lab order. */
 public class SendOrderUseCase {
     private static final SendOrderUseCase INSTANCE = new SendOrderUseCase();
+    private final ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
```
halprin, line 17: Looks like we don't need this field, right?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
```diff
@@ -82,6 +83,7 @@ public Map<HttpEndpoint, Function<DomainRequest, DomainResponse>> domainRegistra
         ApplicationContext.register(OrderSender.class, ReportStreamOrderSender.getInstance());
         ApplicationContext.register(
                 PartnerMetadataOrchestrator.class, PartnerMetadataOrchestrator.getInstance());
+        ApplicationContext.register(SyncRetryTask.class, SyncRetryTask.getInstance());
```
halprin, line 86: Since `SyncRetryTask` is a shared utility beyond just the ETOR plugin, we should register it in `App.java`.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/utils/SyncRetryTask.java
```diff
@@ -0,0 +1,48 @@
...
+                logger.logWarning(
+                        "Attempt " + attempt + ": Retrying in " + (waitTime * 2) / 1000 + "s");
+
+                try {
+                    Thread.sleep(waitTime);
+                } catch (InterruptedException ie) {
+                    Thread.currentThread().interrupt(); // Restore interrupted status
+                    logger.logError("Thread interrupted during wait before retry", ie);
+                    break;
+                }
```
halprin, line 39: Very nice with handling the interrupt exception!

shared/src/main/java/gov/hhs/cdc/trustedintermediary/utils/SyncRetryTask.java
```diff
@@ -0,0 +1,48 @@
...
+    private static final SyncRetryTask INSTANCE = new SyncRetryTask();
+    @Inject Logger logger;
+
+    public static SyncRetryTask getInstance() {
+        return INSTANCE;
+    }
+
+    private SyncRetryTask() {}
+
+    public <T> boolean retry(Callable<T> task, int maxRetries, long waitTime) {
```
halprin, line 21: I recommend that we return `T` instead of boolean and that this method throws `Exception` that will originate from the `Callable` on the last retry failure.

This get's us two benefits.
1. We get the return value back from the `Callable` if we pass in one that actually returns something.  E.g. `Callable<Integer>`.
2. More importantly, we maintain the Exception cause chain.  If an Exception is thrown inside the `Callable`, that eventually results in a failure, we will want to have that stacktrace of that exception.  The way the code is written today, we loose that cause chain by always catching the Exception and eventually returning `false` for a failure.

What do you think?

shared/src/main/java/gov/hhs/cdc/trustedintermediary/utils/SyncRetryTask.java
```diff
@@ -0,0 +1,48 @@
...
+        int attempt = 0;
+        while (attempt < maxRetries) {
+            try {
+                task.call();
+                return true;
+            } catch (Exception e) {
+                attempt++;
+
+                logger.logWarning(
+                        "Attempt " + attempt + ": Retrying in " + (waitTime * 2) / 1000 + "s");
```
halprin, line 31: This should be converted to use the logging parameters.  E.g...

```suggestion
                logger.logWarning(
                        "Attempt {}: Retrying in {}s", attempt, waitTime * 2 / 1000);
```

halprin (changes requested):
I like the overall approach, but I would recommend some improvements.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/SendOrderUseCase.java
basiliskus: Good catch

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
basiliskus: Done

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/SendOrderUseCase.java
basiliskus: Done

shared/src/main/java/gov/hhs/cdc/trustedintermediary/utils/SyncRetryTask.java
basiliskus: Done

shared/src/main/java/gov/hhs/cdc/trustedintermediary/utils/SyncRetryTask.java
basiliskus: Those are great points, and I agree that's a better implementation. I'll make the changes

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/SendOrderUseCase.java
```diff
@@ -62,17 +65,19 @@ private void saveSentOrderSubmissionId(String receivedSubmissionId, String sentS
...
-            partnerMetadataOrchestrator.updateMetadataForSentOrder(
-                    receivedSubmissionId, sentSubmissionId);
-        } catch (PartnerMetadataException e) {
+        Callable<Void> task =
+                () -> {
+                    partnerMetadataOrchestrator.updateMetadataForSentOrder(
+                            receivedSubmissionId, sentSubmissionId);
+                    return null;
+                };
+        boolean taskExecutedSuccessfully = retryTask.retry(task, 3, 1000);
```
jcrichlake, line 74: Can we increment the retry based on counts of retries? Perhaps double or triple the sleep time after each failure?

basiliskus: The waitTime already multiplies by 2 each iteration

shared/src/main/java/gov/hhs/cdc/trustedintermediary/utils/RetryFailedException.java
```diff
@@ -0,0 +1,11 @@
+package gov.hhs.cdc.trustedintermediary.utils;
+
+public class RetryFailedException extends Exception {
```
halprin, line 3: Let's add some JavaDocs.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/utils/RetryFailedException.java
```diff
@@ -0,0 +1,11 @@
...
+    public RetryFailedException(String message) {
+        super(message);
+    }
```
halprin, line 6: Looks like this constructor is only used in tests.  We could probably remove it and then the test can pass in a bogus `new Exception()` as the second argument.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/utils/RetryFailedException.java
halprin, line 3: We should write a test for this.  See something like `HttpClientExceptionTest` for an example.

halprin:
I also wonder if there's a way to get the original exception to be thrown instead of the new `RetryFailedException`?  There seems to be some issues because Java insists we have a return statement or the last statement be a `throw` though?

basiliskus:
> I also wonder if there's a way to get the original exception to be thrown instead of the new `RetryFailedException`? There seems to be some issues because Java insists we have a return statement or the last statement be a `throw` though?

I wrapped it with a custom exception because we don't know what kind of exception would be thrown. We still have access to the original exception through. I didn't have any issues with the last statement, but we do have to wait until we've pass the max retries before we can throw the exception

shared/src/main/java/gov/hhs/cdc/trustedintermediary/utils/RetryFailedException.java
basiliskus: Done

shared/src/main/java/gov/hhs/cdc/trustedintermediary/utils/RetryFailedException.java
basiliskus: Done

halprin (approved):
Woohoo!  Thanks for hanging in there with all my reviews.

jorg3lopez:
> > 1. There doesn't seem to be support for concurrent calls to `RetryTask#scheduleRetry`.  E.g. there's two tasks that are retrying, then one finishes successfully, it kills the scheduler, and the second task is no longer being executed.
> > 2. Looks like we couldn't call `RetryTask#scheduleRetry` a second time after the first time since the `scheduler` is shutdown and never started back up again.
> > 3. Is the purpose of this retry mechanism to be synchronous or asynchronous?  This looks asynchronous to me, but do we need it to be synchronous?  For example, if there's still a failure after the last retry, do we want to bubble up that last exception so it is thrown to the top level call to `RetryTask#scheduleRetry`?
> 
> Good points. I'd need to do some more research on when and how ScheduledThreadPool is expected to be shut down. Have't used it before so I'm not very familiar. And you're right, this is async but that will create some issues. I'd need to put more though into this because it was a quick idea that I meant to try after the other PRs were done

Yes, this is what I was talking about when I brought up having a method that would do the shutdown so if there were other tasks in the scheduler, then they wouldn't be wiped out.

basiliskus:
> > > 1. There doesn't seem to be support for concurrent calls to `RetryTask#scheduleRetry`.  E.g. there's two tasks that are retrying, then one finishes successfully, it kills the scheduler, and the second task is no longer being executed.
> > > 2. Looks like we couldn't call `RetryTask#scheduleRetry` a second time after the first time since the `scheduler` is shutdown and never started back up again.
> > > 3. Is the purpose of this retry mechanism to be synchronous or asynchronous?  This looks asynchronous to me, but do we need it to be synchronous?  For example, if there's still a failure after the last retry, do we want to bubble up that last exception so it is thrown to the top level call to `RetryTask#scheduleRetry`?
> > 
> > 
> > Good points. I'd need to do some more research on when and how ScheduledThreadPool is expected to be shut down. Have't used it before so I'm not very familiar. And you're right, this is async but that will create some issues. I'd need to put more though into this because it was a quick idea that I meant to try after the other PRs were done
> 
> Yes, this is what I was talking about when I brought up having a method that would do the shutdown so if there were other tasks in the scheduler, then they wouldn't be wiped out.

Right, sorry I didn't understand it at the time but that is a good point. I changed the implementation to be completely synchronous now

shared/src/main/java/gov/hhs/cdc/trustedintermediary/utils/SyncRetryTask.java
```diff
@@ -0,0 +1,48 @@
...
+    private static final SyncRetryTask INSTANCE = new SyncRetryTask();
+    @Inject Logger logger;
+
+    public static SyncRetryTask getInstance() {
+        return INSTANCE;
+    }
+
+    private SyncRetryTask() {}
+
+    public <T> T retry(Callable<T> task, int maxRetries, long waitTime)
```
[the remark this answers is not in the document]
basiliskus: I'm not familiar with atomic integers. What would be the advantage of using it?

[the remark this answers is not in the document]
jorg3lopez: I think we can keep it as an integer as it is passed in as a parameter, making it thread safe.

jorg3lopez (approved):
Great job.

PR #755, opened 2024-01-08
Story/672/marshal results into fhir

jcrichlake:
#672 

Marshaling our stored data into a FHIR object

## Issue

Add a link to the issue here.  Consider using
[672](https://app.zenhub.com/workspaces/cdc-ti-633de70d939f6c00155f3722/issues/gh/cdcgov/trusted-intermediary/672)

## Checklist

- [ ] I have added tests to cover my changes
- [ ] I have added logging where useful (with appropriate log level)
- [ ] I have added JavaDocs where required
- [ ] I have updated the documentation accordingly

**Note**: You may remove items that are not applicable

PR #760, opened 2024-01-09
672: Metadata Database Support for Sent Submission ID and Extra Null Checks

halprin:
# Metadata Database Support for Sent Submission ID and Extra Null Checks

_Review #755 first._

The database was missing support for the sender submission ID.  I added support for that.
There were also some missing `null` checks when working with the received timestamp and the database.

## Issue

#672.

## Checklist


etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/PostgresDao.java
```diff
@@ -121,11 +129,18 @@ public synchronized PartnerMetadata fetchMetadata(String receivedSubmissionId)
                 return null;
             }
 
+            Instant timeReceived = null;
+            Timestamp timestamp = result.getTimestamp("time_received");
+            if (timestamp != null) {
+                timeReceived = timestamp.toInstant();
+            }
+
```
jcrichlake, line 137: If timestamp is null this will still throw a null pointer

halprin: Correct, but that statement is wrapped in that `if` statement that ensures that `timestamp` is _not_ `null`.

jcrichlake (approved):

PR #763, opened 2024-01-09
672: Don't Call RS History API for Sent Order on Order Endpoint

halprin:
# Don't Call RS History API for Sent Order on Order Endpoint

We are no longer calling the RS history API for the _sent_ order in the order endpoint.  We only call it from the metadata endpoint now.

This allows us to delete the retry logic because it was only used to retry the call to the RS history API for the _sent_ order in the hope that we'd get the receiver name at the end of the order endpoint.

Updated the unit tests.

## Issue

#672

## Checklist


jcrichlake:
Nicely done

basiliskus (approved):
Thanks for cleaning up the code, looks good!

Issue #610, opened 2023-10-25
Metadata: Hospital Order Status


scleary1cs:
# Story
As a hospital, so I can confirm my order has been received by the lab, I need to be able to view the delivery status of all orders I’ve submitted recently.

## Acceptance Criteria
- [x] The hospital is able to view the delivery status (e.g. pending, delivered, failed) of their orders.
- [x] The order metadata is available in the form of FHIR OperationOutcome
- [x] The hospital is able to view all orders submitted by them over the past 90 days.

### Research
- [ ] *Research work needed to complete the story*
- [ ] Foundational: *High-level research that will support this and future efforts*

###  Engineering
- [x] Figure out the tasks.
- [x] Create new enum for the different states for the delivery status.
- [x] Add new delivery status "column" to `PartnerMetadata` and associated other DB updates.
- [x] Update PostgresDao to retrieve new column
- [x] Update PostgresDao to write new data to database
- [x] Add stored data to metadata object for endpoint to return
- [x] Set delivery status to `FAILED` when the intermediary wouldn't forward on a message after accepting it (e.g. anything that happens within the `EtorDomainRegistration#handleOrders` method). - @tjohnson7021, @halprin 
- [x] Whenever we call the RS history API for the outbound order, we need to map their statuses to one of our status. - @halprin 
- [x] (See PR #770) Fix `Caused by: org.postgresql.util.PSQLException: ERROR: column "delivery_status" is of type message_status but expression is of type character varying.  Hint: You will need to rewrite or cast the expression.  Position: 54`. - @halprin 
- [x] When deploying a second time, we get the following issue: `ERROR:  type "message_status" already exists`.  Looks like we need to make the SQL statement `CREATE TYPE message_status...` conditional. - @jcrichlake, @halprin 

## Definition of Done
- [x] Documentation tasks completed
  - [x] Documentation and diagrams created or updated
    - [x] Implementation guide (`/ig` folder)
    - [x] ADRs (`/adr` folder)
    - [x] Main `README.md`
    - [x] Other READMEs in the repo
    - [x] If applicable, update the `ReportStream Setup` section in `README.md`
  - [x] Threat model updated
  - [x] API documentation updated
  - [x] Source code documentation created when the code is not self-documenting
- [x] Code quality tasks completed
  - [x] Logging added where useful
  - [x] Code refactored for clarity and no design/technical debt
  - [x] Adhere to separation of concerns; code is not tightly coupled, especially to 3rd party dependencies
[...]

PR #766, opened 2024-01-11
Story 610 metadata order status

jcrichlake:
# Adding message status to database and ReST endpoint

This PR adds the Order status data to both the database and the ReST endpoint.  We will need to discover the statuses ReportStream uses and add the logic to determine the status as needed to complete the story.

## Issue
#610 

## Checklist

### DIFF
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/PostgresDao.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/PostgresDao.java
@@ -143,11 +143,6 @@ public class PostgresDao implements DbDao {
                 timeReceived = timestamp.toInstant();
             }
 
-            String status =
-                    result.getString("delivery_status") != null
-                            ? result.getString("delivery_status")
-                            : "PENDING";
-
             return new PartnerMetadata(
                     result.getString("received_message_id"),
                     result.getString("sent_message_id"),
@@ -155,7 +150,7 @@ public class PostgresDao implements DbDao {
                     result.getString("receiver"),
                     timeReceived,
                     result.getString("hash_of_order"),
-                    PartnerMetadataStatus.valueOf(status));
+                    PartnerMetadataStatus.valueOf(result.getString("delivery_status")));
         }
     }
 }
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/PostgresDao.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/PostgresDao.java
@@ -4,7 +4,6 @@ import gov.hhs.cdc.trustedintermediary.context.ApplicationContext;
 import gov.hhs.cdc.trustedintermediary.etor.metadata.PartnerMetadata;
 import gov.hhs.cdc.trustedintermediary.etor.metadata.partner.PartnerMetadataStatus;
 import gov.hhs.cdc.trustedintermediary.external.azure.AzureClient;
-import gov.hhs.cdc.trustedintermediary.wrappers.DbDao;
 import gov.hhs.cdc.trustedintermediary.wrappers.Logger;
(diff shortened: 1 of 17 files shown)


etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/PostgresDao.java
```diff
@@ -135,13 +143,19 @@ public synchronized PartnerMetadata fetchMetadata(String receivedSubmissionId)
                 timeReceived = timestamp.toInstant();
             }
 
+            String status =
+                    result.getString("delivery_status") != null
+                            ? result.getString("delivery_status")
+                            : "PENDING";
```
halprin, line 149: Oh yeah, I wanted to ask about this.  Is it correct to default to `PENDING`?  Or should we keep it as `null` so that `PartnerMetadata` has `null` for the delivery status?  What was your thinking behind defaulting it to `PENDING`?

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/database/PostgresDaoTest.groovy
```diff
@@ -200,7 +203,7 @@ class PostgresDaoTest extends Specification {
         mockPreparedStatement.executeQuery() >> mockResultSet
         mockResultSet.next() >> true
         mockResultSet.getTimestamp("time_received") >> null
-
```
halprin, line 203: In a new test similar to this one, can you test that when the database returns `null` for the `delivery_status`, that it is converted to `PENDING`?  This comment, of course, depends on the answer of the comment asking about whether defaulting to `PENDING` is right or not.

halprin (commented):
Oooo goodie!  We're making some great progress.  Sorry I was barely there to contribute.  :(

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/PostgresDao.java
jcrichlake: So I defaulted to `PENDING` for 2 reasons.  1. Not having a status doesn't seem like correct behavior, if we have a message it should have a status. 2. We've set the status column up to be a custom type.  I'm not sure if this would accept a null value when it's a type looking for 1 of 3 specific strings in the db.

halprin: That makes sense.  Let's not do this defaulting in the db layer but codify it in `PartnerMetadata`, the business class in charge of the enum.  You can add something like this...

```java
public PartnerMetadata {
    if (deliveryStatus == null) {
        deliveryStatus = PartnerMetadataStatus.PENDING;
    }
}
```

jcrichlake: Okay, sounds good

jcrichlake: Done

halprin (approved):
Score!

basiliskus:
Will the status handling be added to the `FilePartnerMetadataStorage` implementation as well? Is there a task for that?

halprin:
> Will the status handling be added to the `FilePartnerMetadataStorage` implementation as well? Is there a task for that?

It should just work.  The status handling has moved to `PartnerMetadata` so we don't need to replicate it elsewhere.

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/database/PostgresDaoTest.groovy
halprin: Not needed anymore since we moved the logic to the `PartnerMetadata`.

Issue #615, opened 2023-10-26
Receiving Results via REST


scleary1cs:
# Story
As a lab, so that test results can forwarded to a hospital, I need the intermediary to accept results via REST.

## Pre-conditions
- ReportStream is already able to send us a FHIR results message based upon an ORU from the lab.
- The incoming message is a valid FHIR lab test results message

## Acceptance Criteria
- [x] An indication of success or failure is returned to the calling service (e.g. an HTTP response)
- [x] The incoming FHIR results are only accepted if and only if it is sent from a trusted organization (e.g. ReportStream).
- [x] For production & staging: Results data is transmitted securely

###  Engineering
- [x] Create the TI results endpoint
  - [x] results endpoint skeleton (branch)
    - [x] handleResults method in EtorDomainRegistration @jorg3lopez 
  - [x] Enforce authentication
- [x] ResultController @saquino0827 
- [x] SendResultUseCase @basiliskus 
  - [x] Create `MessageSender` interface
    - [x] Initial implementation for `ReportStreamResultSender`
    - [x] Refactor `OrderSender` to use interface
  - [x] Create `SendMessageUseCase` interface
    - [x] Initial implementation for `SendResultUseCase`
    - [x] Refactor `ConvertAndSendDemographicsUseCase` to use interface
  - [x] Add `UnableToSendMessageException`
    - [x] Use exception in result sending classes
    - [x] Use exception in orders sending classes
    - [x] Use exception in demographics sending classes
- [x] Parse results
- [x] ResultResponse
  - [x] Return success/failure notice via HTTP response
- [x] Result Interface
  - [x] interface @jorg3lopez 
  - [x] HapiResult @saquino0827 @jorg3lopez 
- [x] Configure ReportStream to route ORU messages to the TI results endpoint @basiliskus 
  - [x] Create PR in RS to add settings ([created here](https://github.com/CDCgov/prime-reportstream/pull/13114))
  - [x] Get PR merged 
  - [x] Apply settings to RS staging
  - [x] Re add client keys to staging: `flexion.simulated-hospital`, `flexion.etor-service-sender`, `flexion.simulated-lab`
- [x] DoD items that will require work:
  - [x] Update swagger docs
  - [x] Update thread model
  - [x] Update load tests
  - [x] README docs
  - [x] Update `ReportStream Setup` in readme @basiliskus 

## Definition of Done
- [x] Documentation tasks completed
  - [x] Documentation and diagrams created or updated
[...]

PR #803, opened 2024-01-24
SendResultUseCase implementation

basiliskus:
# SendResultUseCase implementation

Initial implementation for `SendResultUseCase` + needed result interfaces

## Issue

#615 

## Checklist

### DIFF
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messages/SendMessageUseCase.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messages/SendMessageUseCase.java
@@ -1,4 +1,4 @@
-package gov.hhs.cdc.trustedintermediary.etor.message;
+package gov.hhs.cdc.trustedintermediary.etor.messages;
 
 public interface SendMessageUseCase<T> {
 
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messages/SendMessageUseCase.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messages/SendMessageUseCase.java
@@ -1,5 +1,11 @@
 package gov.hhs.cdc.trustedintermediary.etor.messages;
 
+/**
+ * This interface represents a use case for sending a generic message. It provides a method to
+ * convert and send the message.
+ *
+ * @param <T> the type of message to be sent
+ */
 public interface SendMessageUseCase<T> {
 
     void convertAndSend(final T message, String receivedSubmissionId)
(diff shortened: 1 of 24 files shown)


etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messages/SendMessageUseCase.java
```diff
@@ -0,0 +1,13 @@
+package gov.hhs.cdc.trustedintermediary.etor.messages;
+
+/**
+ * This interface represents a use case for sending a generic message. It provides a method to
+ * convert and send the message.
+ *
+ * @param <T> the type of message to be sent
```
jcrichlake, line 7: Is this a param?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/results/SendResultUseCase.java
```diff
@@ -0,0 +1,24 @@
...
+    @Inject MessageSender<Result<?>> sender;
+
+    private SendResultUseCase() {}
+
+    public static SendResultUseCase getInstance() {
+        return INSTANCE;
+    }
+
+    @Override
+    public void convertAndSend(Result<?> result, String receivedSubmissionId)
```
jcrichlake, line 20: Should the conversion happen in the use case or should that be a separate class to handle that?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messages/SendMessageUseCase.java
basiliskus: Generics use the same @param tag in Javadocs: https://stackoverflow.com/a/2015998

saquino0827 (commented):
I like the Message typing. I was thinking of combining the Result and Order objects to a Message object as well but I wanted to discuss with you and @jorg3lopez . I'll wait until he takes a look at this PR and we can see where we want to go but the PR looks good!

jcrichlake:
Really nice PR! Just a couple of questions that the answer may just be "No it's fine" hah

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/results/SendResultUseCase.java
basiliskus: I think it should. That's what we're doing in `SendOrderUseCase`. Maybe the use case should be named `ConvertAndSendResultUseCase` but I don't know if that's necessary

saquino0827 (approved):

PR #820, opened 2024-01-29
updated OpenAPI documentation to include summary API

rajeev-flexion:


etor/src/main/resources/openapi_etor.yaml
```diff
@@ -105,6 +105,38 @@ paths:
             application/json:
             schema:
               $ref: '#/components/schemas/NotFoundError'
+  /v1/etor/metadata/summary/{sender}:
+    get:
+      summary: Returns all the order and their status for particular hospital
```
basiliskus, line 110: Is this metadata summary endpoint just for orders or will it work for results as well?

etor/src/main/resources/openapi_etor.yaml
```diff
@@ -105,6 +105,38 @@ paths:
...
+            type: string
+      security:
+        - bearerAuthorization: [ ]
+      responses:
+        '200':
+          description: orders and their status returned successfully
+          content:
+            application/json:
+            schema:
+              $ref: 'https://github.com/LinuxForHealth/FHIR/blob/main/fhir-openapi/src/main/webapp/META-INF/openapi.json?raw=true#/components/schemas/OperationOutcome'
```
basiliskus, line 125: Is the return type an `OperationOutcome` message? Is it a list of `OperationOutcome`? Or something else?

etor/src/main/resources/openapi_etor.yaml
```diff
@@ -105,6 +105,38 @@ paths:
...
+        '401':
+          description: Authentication failed due to invalid token or unknown organization
+          content:
+            application/json:
+            schema:
+              $ref: '#/components/schemas/UnauthorizedRequestError'
+        '404':
```
basiliskus, line 132: Is the endpoint returning 404? I didn't find it. If it's not, it would be a good idea to add the condition to return 404 in case nothing is found

basiliskus (changes requested):
Looks good but I left a few comments/questions

etor/src/main/resources/openapi_etor.yaml
rajeev-flexion: results as in?

basiliskus: Lab results. We have lab orders as one part of the flow from the hospital to the lab, and then we're adding lab results that will go from the lab to the hospital

etor/src/main/resources/openapi_etor.yaml
jcrichlake: It is something different

etor/src/main/resources/openapi_etor.yaml
jcrichlake: I think this should be 500 not 404

etor/src/main/resources/openapi_etor.yaml
jcrichlake: We can update this to include orders when or if we get to this point

etor/src/main/resources/openapi_etor.yaml
basiliskus: 500 error codes are for unexpected internal errors. Not finding any metadata should be an expected outcome

etor/src/main/resources/openapi_etor.yaml
```diff
@@ -105,6 +105,30 @@ paths:
             application/json:
             schema:
               $ref: '#/components/schemas/NotFoundError'
+  /v1/etor/metadata/summary/{sender}:
+    get:
+      summary: Returns all the order and their status for particular hospital
```
basiliskus, line 110: Couple of typos: `Returns all the orders and their status for a particular hospital`

jcrichlake (approved):

basiliskus (approved):
Looks good! Just pointed out some typos

Issue #676, opened 2023-11-22
Metadata: Hospital Consolidated Order Status


JohnNKing:
# Story
As a hospital, so I can confirm that all my orders have been successfully delivered to the lab, I need to be able to view the delivery status of orders coming from me recently.

## Acceptance Criteria
- [x] The hospital is able to view the delivery status (e.g. pending, delivered, failed) of all their orders in a consolidated response.
- [x] The hospital is able to view all orders submitted by them for at least the past 90 days.

###  Engineering
- [x] Add new endpoint to Etor Domain Registration, it should take the sender name as part of the path and use that to search the database. 
- [x] Add handle method for the new endpoint
- [x] Update PartnerMetadataOrchestrator and it's Interface to have a new method to return a  map composed of received submission id (Map key), the status and the status message (Stored in the map as 1 value to be put together and delimited by a character.)
- [x] Retrieve data needed from the database
- [x] Compose the queried data into the map
- [x] Return the map to the handle method in the API call 
- [x] Update the openAPI documentation for the new endpoint responses - @rajeev-flexion 
- [x] Test new endpoint in e2e tests.  - @jcrichlake 

## Definition of Done
- [ ] Documentation tasks completed
  - [ ] Documentation and diagrams created or updated
    - [x] Implementation guide (`/ig` folder)
    - [x] ADRs (`/adr` folder)
    - [x] Main `README.md`
    - [x] Other READMEs in the repo
    - [x] If applicable, update the `ReportStream Setup` section in `README.md`
  - [x] Threat model updated
  - [x] API documentation updated
  - [x] Source code documentation created when the code is not self-documenting
- [ ] Code quality tasks completed
  - [x] Logging added where useful
  - [x] Code refactored for clarity and no design/technical debt
  - [x] Adhere to separation of concerns; code is not tightly coupled, especially to 3rd party dependencies
  - [x] Code is reviewed or developed by pair; 1 approval is needed but consider requiring an outside-the-pair reviewer
  - [x] Code quality checks passed
- [x] Security & Privacy tasks completed
  - [x] Security & privacy gates passed
- [x] Testing tasks completed
  - [x] Load tests passed
  - [x] Unit test coverage of our code >= 90%
- [ ] Build & Deploy tasks completed
  - [x] Build process updated
  - [x] API(s) are versioned
  - [x] Feature toggles created and/or deleted.  Document the feature toggle
  - [ ] Source code is merged to the main branch

## Research Questions
[...]

PR #822, opened 2024-01-30
Story 676/load tests

jcrichlake:
# Renaming input files for load tests

We re-orged and re-named a lot of files that are used for load testing but we didn't propagate those changes into the test itself.  This PR fixes that
## Issue

#676


operations/locustfile.py
```diff
@@ -74,14 +76,22 @@ def post_v1_etor_orders(self):
...
             self.orders_api_called = True
 
-    @task(2)
+    @task(1)
     def get_v1_etor_metadata(self):
         if self.orders_api_called:
             self.client.get(
                 f"{METADATA_ENDPOINT}/{self.submission_id}",
                 headers={"Authorization": self.access_token},
             )
```
saquino0827, line 86: I assume this will be added later on for `CONSOLIDATED_ENDPOINT`? Or was this accidentally commented out?

saquino0827 (commented):
A quick question but the PR looks good!

operations/locustfile.py
```diff
@@ -74,14 +76,22 @@ def post_v1_etor_orders(self):
...
 
+    # @task(1)
+    # def get_v1_metadata_consolidated(self):
+    #     if self.orders_api_called:
+    #         self.client.get(
+    #             f"{CONSOLIDATED_ENDPOINT}/{self.sender}",
+    #             headers={"Authorization": self.access_token},
+    #         )
+
```
saquino0827, line 94: I assume this will be added back for `CONSOLIDATED_ENDPOINT` or was this commented out by accident?

operations/locustfile.py
saquino0827: Resolving this as I commented in the wrong spont.

Issue #838, opened 2024-02-02
Bug: Mock Secrets Not Being Accessed


jorg3lopez:
# Bug

API fails with a NullPointerException due to a null secret variable.

## Describe the Bug

When running the api in a local environment, it fails the process of retrieving our local secrets from the mock_credentials folder.

## Impact

Unable to process reports locally for testing.

## To Reproduce

1. run the api
2. request token
3. send payload to the reports endpoint

## Expected Behavior

Error message saying it failed to retrieve credentials from resources.

## Screenshots

<img width="1605" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNkVoQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--892cec3a5b760079597943fb7e72a6e8c120497b/image.png" alt="image.png" />

<img width="1094" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNkFoQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b9c7dcfcb4c818322bab1e89d1239e3605e7702c/image.png" alt="image.png" />

PR #839, opened 2024-02-02
Add Mock_Credentials As A Resource To ETOR Module

jorg3lopez:
# Mock_Credentials As A Resource 

This PR adds the mock_credentials folder to the shared module. It also adds the the steps to take to debug the api inside the container, to the `README.md` file.

### Notes

`InputStream secretStream =
                getClass().getClassLoader().getResourceAsStream(secretName + ".pem")`

is the way that the local secrets are accessed by the etor module. The `mock_credentials` folder is at the same level as the etor module folder. The `getClass.getClassLoader().getResourceAsStream()` checks the resource folder of the etor module. So it never reaches outside of it's module scope. To fix this, the `mock_credentials` folder can be included as a resource to the shared module, which is included as a resource in the etor module. This approach keeps the module decoupled from the resource folder.

## Issue
#838


jorg3lopez:
Does anyone know what is the difference between these two methods? If the `readSecretFromFileSystem` is accessing the `mock_credentials` folder as well, then we can use the `getClass().getClassLoader().getResourceAsStream()` approach to get the credentials from the `mock_credentials` folder. Thoughts?

image

Dockerfile
```diff
@@ -10,7 +10,7 @@ WORKDIR /app
 COPY ${JAR_LIB_FILE} app.jar
 
 # Run the api
-CMD ["java", "-jar", "app.jar"]
+CMD ["java", "-agentlib:jdwp=transport=dt_socket,address=*:5005,server=y,suspend=n", "-jar", "app.jar"]
```
jorg3lopez, line 14: RS debug port is `5005` as well. This might create conflicts if running both applications, in docker, at the same time. Thoughts on changing to another port number?

halprin:
> Does anyone know what is the difference between these two methods? If the `readSecretFromFileSystem` is accessing the `mock_credentials` folder as well, then we can use the `getClass().getClassLoader().getResourceAsStream()` approach to get the credentials from the `mock_credentials` folder. Thoughts?

`readSecretFromFileSystem` tries to read the secret from the pre-existing `/mock_credentials/` folder.  This form only works when running TI through Gradle.  E.g. `./gradlew clean app:run` of if the current working directory when running the jar file is a sub-directory below the top-level of the repository.

`readSecretFromResources` reads from resources.  You'll see that the `/shared/src/main/resources/` folder contains some duplicate secrets from `/mock_credentials/` folder.  This form always works no matter what the current working directory is when you run the jar file because they are bundled into the jar file.

etor/build.gradle
```diff
@@ -53,3 +53,12 @@ jacocoTestReport {
...
 }
+
+sourceSets {
+    main {
+        resources {
+            // Include the mock_credentials folder as part of the resources
+            srcDir '../mock_credentials'
+        }
+    }
+}
```
halprin, line 64: I'd move this to the `shared` sub-project and delete the other secrets that are already stored in `/shared/src/main/resources/`.

Dockerfile
```diff
@@ -10,7 +10,7 @@ WORKDIR /app
 COPY ${JAR_LIB_FILE} app.jar
 
 # Run the api
-CMD ["java", "-jar", "app.jar"]
+CMD ["java", "-agentlib:jdwp=transport=dt_socket,address=*:6006,server=y,suspend=n", "-jar", "app.jar"]
```
halprin, line 13: Is this safe to run in production?  This `Dockerfile` is used when running our application in production.  Does it also not degrade performance?

basiliskus:
> `readSecretFromResources` reads from resources. You'll see that the `/shared/src/main/resources/` folder contains some duplicate secrets from `/mock_credentials/` folder. This form always works no matter what the current working directory is when you run the jar file because they are bundled into the jar file.

@halprin Do we need these mock credential files duplicated?

Dockerfile
jorg3lopez: Yes, it can degrade performance. I'll exclude it.

docker-compose.yml
```diff
@@ -6,5 +6,5 @@ services:
       REPORT_STREAM_URL_PREFIX: http://localhost:7071
     ports:
       - "8080:8080" # default api endpoint port
-      - "5005:5005" # Java debug port
+      - "6006:6006" # Java debug port
```
jorg3lopez, line 9: RS runs the debug port on `5005`. This can create conflicts if TI uses the same port for debugging the container.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/localfile/LocalSecrets.java
```diff
@@ -33,26 +30,17 @@ public String getKey(String secretName) throws SecretRetrievalException {
...
         try {
-            key = readSecretFromFileSystem(secretName);
-        } catch (SecretRetrievalException exception) {
-            logger.logWarning(
-                    "Not finding the " + secretName + " on the filesystem, searching in resources");
             key = readSecretFromResources(secretName);
+        } catch (SecretRetrievalException exception) {
+            logger.logWarning("Not finding the " + secretName + " in resources");
+            throw new SecretRetrievalException(
+                    "Could not find " + secretName + "in resources", exception);
```
halprin, line 37: No-need to rewrap the `SecretRetrievalException` that comes out of the `readSecretFromResources` method.  You can completely nuke the try/catch stuff and just call the `readSecretFromResources` method directly.

halprin (commented):
Thanks for the documentation!  I just have one suggestion in `LocalSecrets.java`

saquino0827 (approved):

saquino0827:
I was looking at the wrong PR when and cycled to this one for approval.

halprin (approved):

Issue #721, opened 2023-12-12
Initial Production Deploy and its Approval


halprin:
# Backlog Task

Aquire any approvals we need for our initial deployment to our production environment and then do the production deployment.

## Completion Criteria

- [x] Approval for a production deploy is given.
- [x] Our intermediary is deployed to production.

## Tasks

- [x] Given permission.
- [x] Give GitHub Deployer appropriate permission to deploy to prod environment.
- [x] Create any required Azure bucket's for remote Terraform state.
- [x] Deploy to production.
- [x] Set secrets.

## Other Notes

We are unsure whether we need to get permission from our ISSOs for our initial deployment to production.  
So, we rather be safe than sorry and get permission.  The permission we are seeking is to deploy our application into the production environment _but_ not to process PII or PHI just yet.  Just fake/test data for now.  There's #175 for seeking permission to start handling PHI and PII in our production environment.  With that said, we should reach out to the correct ISSOs, ask for permission, and if given, deploy to production.  If they require us to get approval to handle PHI and PII before any deployment to production occurs, then this backlog item should be scrapped and #175 should be prioritized.

@sfradkin talked with Mo and others about this process.  Looks like we are cleared to deploy to production, but will still need approval to handle PHI and PII.

scleary1cs:
Combine with #175?

scleary1cs:
Combine with #175?

halprin:
Uploaded the secrets.  Only secret that isn't filled in is the `organization-report-stream-public-key-prd`.  That will require ReportStream to onboard us as a receiver in their production and for them to generate public key for us there.

PR #847, opened 2024-02-05
Use Environments Name for All the Different Deploy Steps

halprin:
# Use Environments Name for All the Different Deploy Steps

Our deployments use OIDC to authenticate itself to Azure.  GitHub's OIDC implementation uses a couple of different subjects when talking over OIDC depending on the situation.  During a normal deploy from the `main` branch, it uses the following: `repo:CDCgov/trusted-intermediary:head:main`.  When doing a deploy from a `v1.0.0` tag, it uses the following: `repo:CDCgov/trusted-intermediary:tag:v1.0.0`.

Azure doesn't support wildcards subjects.  For example, I cannot tell Azure to allow `repo:CDCgov/trusted-intermediary:tag:v*`.  Azure is [working](https://github.com/Azure/azure-workload-identity/issues/373) on support.  So we don't need to constantly create new allowed tags in Azure for every new deploy we want to do, we are labeling every job that interacts with Azure as an environment.  This will make GitHub's OIDC handling logic advertise itself to Azure as `repo:CDCgov/trusted-intermediary:environment:stg`, as an example for the staging environment.  This allows us to set Azure to only allow our named environments.

## Issue

#721.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -19,6 +19,7 @@ jobs:
     needs: ci
     uses: ./.github/workflows/terraform-deploy_reusable.yml
     with:
+      ENVIRONMENT: stg
       TERRAFORM_DIRECTORY: operations/environments/staging
     secrets:
       AZURE_CLIENT_ID: ${{ secrets.AZURE_CDC_CLIENT_ID }}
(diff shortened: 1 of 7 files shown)


jcrichlake (approved):
👍

Issue #753, opened 2024-01-05
Database CICD and Connection Pooling


jcrichlake:
# DevEx/OpEx

High level description of what what needs to be improved.

### Connection Pooling
- [x] Determine the connection pooling library we should implement
  - We will be using [Hikari](https://github.com/brettwooldridge/HikariCP). Based on excellent documentation and seemingly having the best performance by a significant margin.
- [x] Add HikariCP dependency to the project
- [x] Remove Synchronized from postgresDao and see if the load tests pass 
- [x] Add  HikariCP connection manager using the datasourceClassName of `org.postgresql.ds.PGSimpleDataSource`
- [x] Swap the existing `DriverManager` in the `PostgresDao` class out for the `DataSource` class defined in the previous step. 
- [x] Reimplement Azure credentialing support.
- [x] Make sure NTP is configured in all vms so the db doesn't have random issues based on timing 
- [x] Verify that connection works via manual testing
- [x] Verify connection works by running E2E tests
- [x] Discuss if we should update the load test script to run the .env script to default to the database implementation
- [x] Run the load tests with the database implementation running locally
- [x] After selecting the connection pool library, re-visit tasks and determine what configuration items are needed
 

### Database CICD
 - [x] Determine database migration tool that we want to implement inside of the intermediary
   - We will be implementing Liquibase as the migration engine, primarily due to the rollback feature being part of the free open source community edition.   
 - [x] Implement the migration tool
 - [x] Remove the database setup scripts from the terraform file
 - [x] Add script to local documentation for migrating locally
 - [x] Add the scripts to liquibase, Especially user permission scripts
 - [x] Add the appropriate pipeline steps to our terraform to allow the migrations to run on deploy
 - [x] Add rollback steps to terraform to be compliant with blue/green production readiness standards
 - [x] Decide if we want the rollback inputs to require a specific yml file or if we need to re-structure the files we have
 - [x] Validate and understand the rollback behavior in a non-local environment (Staging), document any interesting discoveries

## Definition of Done
- [x] Documentation tasks completed
  - [x] Documentation and diagrams created or updated
    - [x] Update readme or onboarding docs to include liquibase install? 
    - [x] ADRs (`/adr` folder)
      - [x] Connection Pool
      - [x] Liquibase Migration Tool
    - [x] Main `README.md`
    - [x] Other READMEs in the repo
  - [x] Threat model updated
  - [x] API documentation updated
  - [x] Source code documentation created when the code is not self-documenting
[...]

PR #863, opened 2024-02-09
Add Rollback for Github Action

jcrichlake:
# Rollback db changes

We need the ability to rollback db migrations from Github action runners.  

## Issue
#753


.github/workflows/db_rollback.yml
```diff
@@ -0,0 +1,68 @@
+name: Rollback DB
+
+on:
+  workflow_dispatch:
+    inputs:
```
halprin, line 5: Add an input for which deployed environment we're picking.  Super cool if it could be a dropdown.

.github/workflows/db_rollback.yml
```diff
@@ -0,0 +1,68 @@
...
+      rollbackFile:
+        description: "File with migrations you want to rollback"
+        required: true
+        default: Filepath to yml
+
+jobs:
+  terraform-deploy:
+    name: Terraform Deploy
+    environment:
+      name: ${{ inputs.ENVIRONMENT }}
```
halprin, line 19: Here's a `inputs.`, so that doesn't work in this instance.  This will need to convert to get the environment name from the above workflow dispatch input.

.github/workflows/db_rollback.yml
```diff
@@ -0,0 +1,68 @@
...
+    runs-on: ubuntu-latest
+    env:
+      ARM_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
+      ARM_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
+      ARM_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
```
halprin, line 24: We have different secrets depending on whether we are deploying to the Flexion or CDC domain.  Maybe we can do a ternary if statement based on the workflow dispatch environment input?

Something like `${{ github.event.inputs.environment == 'internal' ? secrets.AZURE_CLIENT_ID : secrets.AZURE_CDC_CLIENT_ID }}`.

.github/workflows/db_rollback.yml
```diff
@@ -0,0 +1,68 @@
...
+      ARM_USE_OIDC: true
+    permissions:
+      id-token: write
+      contents: read
+    defaults:
+      run:
+        working-directory: ${{ inputs.TERRAFORM_DIRECTORY }}
```
halprin, line 31: Another `inputs.` that won't work in this context.  This might be something like `./operations/environments/${{ github.event.inputs.environment }}`?

.github/workflows/db_rollback.yml
```diff
@@ -0,0 +1,68 @@
...
+      REGISTRY: ${{ steps.export-terraform-output.outputs.REGISTRY }}
+      APP: ${{ steps.export-terraform-output.outputs.APP }}
+
+    steps:
+
+      - uses: actions/checkout@v4
+
+      - name: Terraform Init
+        id: init
+        run: terraform init ${{ inputs.TERRAFORM_INIT_PARAMETERS }}
```
halprin, line 43: We don't need Terraform init parameters for our main environments.
```suggestion
        run: terraform init
```

.github/workflows/db_rollback.yml
```diff
@@ -0,0 +1,68 @@
...
+
+      - name: Login via Azure CLI
+        uses: azure/login@v1
+        with:
+          client-id: ${{ secrets.AZURE_CLIENT_ID }}
+          tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+          subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
```
halprin, line 50: Like the comment above, maybe we can use a ternary if statement to change which secret we use.

.github/workflows/db_rollback.yml
```diff
@@ -0,0 +1,68 @@
...
+        run: |
+          DATABASE_HOSTNAME=$(terraform output -raw database_hostname)
+          DATABASE_PASSWORD=$(az account get-access-token --resource-type oss-rdbms --query "[accessToken]" -o tsv)
+          echo "::add-mask::$DATABASE_HOSTNAME"
+          echo "::add-mask::$DATABASE_PASSWORD"
+          echo "DATABASE_HOSTNAME=$DATABASE_HOSTNAME" >> "$GITHUB_ENV"
+          echo "DATABASE_PASSWORD=$DATABASE_PASSWORD" >> "$GITHUB_ENV"
+
+      - name: Run Db Rollback
+        uses: liquibase-github-actions/rollback-count-sql@v4.25.1
```
halprin, line 62: Do we really want `-sql` version?  That might just output the SQL to do the rollback without actually doing the rollback.  That's at least how it works on the command line.  Maybe the GitHub action is different?

.github/workflows/db_rollback.yml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Extract database hostname and password into GitHub Env
+        run: |
+          DATABASE_HOSTNAME=$(terraform output -raw database_hostname)
+          DATABASE_PASSWORD=$(az account get-access-token --resource-type oss-rdbms --query "[accessToken]" -o tsv)
+          echo "::add-mask::$DATABASE_HOSTNAME"
+          echo "::add-mask::$DATABASE_PASSWORD"
+          echo "DATABASE_HOSTNAME=$DATABASE_HOSTNAME" >> "$GITHUB_ENV"
+          echo "DATABASE_PASSWORD=$DATABASE_PASSWORD" >> "$GITHUB_ENV"
+
+      - name: Run Db Rollback
```
halprin, line 61: We'll need to change the current working directory before we run this step.  This job runs, by default, in the Terraform directory.  See the above `working-directory:`.  Our Liquibase migrations work when ran from the repository root.

halprin:
This PR looks 🎸 awesome!

halprin (approved):

jorg3lopez (approved):
Great job. I was going to bring this up in the engineering sync, but I don't have to anymore.

Issue #641, opened 2023-11-03
FHIR => HL7 ORU


JohnNKing:
# Story
As a hospital, so that ReportStream can send NBS results in a format that my EHR can receive, I need ReportStream to send HL7 ORU messages based on an incoming FHIR result.

## Pre-conditions
- An inbound FHIR message can be received successfully by ReportStream.
- The incoming message is a valid FHIR lab test results message

## Acceptance Criteria
- [x] The outgoing ORU is a valid HL7 v2.5.1 message.
- [x] The translated result message is transmitted soon after the FHIR result has been received.
- [x] Only the intended hospital receives the result message
- [x] The incoming FHIR results are always sent through the TI service (e.g. they don't go directly to the hospital)

###  Engineering
- [x] Identify cause of inconsistent transformation seen by our development team
- [x] FHIR to HL7 mapping for ORU
  - [x] Test current mapping 
  - [x] Document missing fields
    - `OBX-4`: Observation Sub-ID
    - `OBX-29` (value `RSLT`, `SCI`, `QST`) 
  - [x] Update ORU mapping RS 
    - [x] Create `metadata/hl7_mapping/receivers/Flexion/TIResults` to replace `metadata/hl7_mapping/ORU_R01/ORU_R01-base` @basiliskus 
    - [x] Check what is required to map `OBX-4`
  - [x] Remove batch headers and footers in final HL7
  - [x] Update example files for ORU samples after transformation @basiliskus 
    - [x] Fix e2e tests 
  - [x] Look into encoding issue in final message: `The�purpose of newborn screening...`
    - The problem was in the original ORU example file. The file has been fixed
- [x] Fix results routing issue
  - [x] Identify solution to routing based on the original sender (i.e. the lab)
    - [x] Decided to use `MSH.3`, `MSH.4`, `MSH.5` and `MSH.6` for routing 
  - [x] Consult the RS team regarding our proposed solution
    - [x] Will use `jurisdictionalFilter` in org settings to filter by `Organization` (`MSH.X`) 
      - [x] Make sure we have the required FHIR resources from MSH to create the filters @basiliskus 
        - `Bundle.entry.resource.ofType(MessageHeader).destination.receiver.resolve().identifier.where(value = 'simulated-hospital').exists()` 
  - [x] Test that filters work as intended and are sufficient
- [x] RS configured to route the result to the hospital @basiliskus 
[...]

PR #909, opened 2024-02-22
Reorganize test sample files and add ORU test examples

basiliskus:
# Reorganize test sample files and add ORU test examples

## Issue

#641


saquino0827 (approved):
Nice flattened files 👍

basiliskus:
> Good idea. What are your thoughts on creating the test folder as part of the e2e resources? This might duplicate files but it would isolate them from unwanted access. Thoughts on doing something similar for integration testing?

I like that idea, but I don't think we should duplicate the files. Can we just move them to resources and remove them from the examples folder?

Issue #341, opened 2023-05-23
Field Type Validation of NBS Orders


JohnNKing:
# Story
As a hospital, so that NBS orders with invalid information can be fixed more quickly, I need validation of orders that identifies if any data is of the wrong type (e.g. date, string, number).

## Context
Are there key values that, if of the wrong format, would cause things to break when received by the LIMS? For example:
* What if birth weight is provided in ounces instead of grams? Or vice versa? What if "1100g" is provided instead of "1100"?
* Gestational age not provided in weeks?
* What if the bar code is not provided in an expected format?
* What if yes/no question values are not in the desired format? (e.g. 0 vs 1, or "yes" vs "no")
* Utilize LOINC code value sets?

## Pre-conditions
- A single set of validation rules are used (i.e. they are not state or hospital specific).

## Acceptance Criteria
- [x] A list of shared validation rules, based upon data gap analysis work to date, have been documented.
- [x] The shared validations rules are used to validate incoming orders.
- [x] We provide some means of capturing & persisting validation warnings (so they can be communicated back to the sender in the future).
- [x] None of the validation rules added prevent the order from being delivered.

### Research
- [x] Document list of standard validation rules
  - [x] Review data gap for consistency across states
  - [x] Test ReportStream's validation that may reject a message before it reached the ETOR/TI service (e.g. OBX NM without numeric in OBX.5; invalid OBX.6)

###  Engineering
- [x] Document what RS is validating
- [x] For each validation rule, document fhirpath for resource that needs validation (related to #636)
- [x] Look into [HapiFHIR validation library](https://hapifhir.io/hapi-fhir/docs/validation/instance_validator.html)
- [x] Look into [fhirpath](https://hl7.org/fhirpath/) validations
- [x] Are these validations applicable? Taking into account the hapifhir marshaling
- [x] Determine how we want to implement validation rules
- [x] Implement validation rules

#### Proposed Solution

Use a variation on the Rules Engine Design Pattern (read more [here](https://tenmilesquare.com/resources/software-development/basic-rules-engine-design-pattern/) and [here](https://softwarehut.com/blog/tech/design-patterns-rules-engine) and [here](https://yiniski.medium.com/rule-engine-pattern-8a3f0e0c2d81))

The rules would be defined in a json file with this shape:
##### Version 1
```
{
  "rules": [
    {
      "fieldName": "patientName",
      "fhirPath": "Bundle.entry.resource.ofType(Patient)",
      "required": true,
[...]

PR #943, opened 2024-03-12
Validation Rules Engine

basiliskus:
# Validation Rules Engine

Implementation of the Validation Rule Engine, which loads a rule definitions file with conditions and validations. The purpose of the engine is to validate a FHIR resource. Any validation that fails will log a warning. For a validation rule to apply to a specific resource, it needs to meet the condition associated with the rule.

The validations and conditions use the FHIRPath language

## Issue

#341 

## Checklist


etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleDefinition.java
```diff
@@ -0,0 +1,9 @@
+package gov.hhs.cdc.trustedintermediary.etor.ruleengine;
+
+import java.util.List;
+
+public class RuleDefinition {
```
halprin, line 5: Looks like you could merge `RuleDefinition` with `ValidationRule`.  Basically this could `implements Rule`?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
```diff
@@ -0,0 +1,60 @@
...
+        return INSTANCE;
+    }
+
+    public void unloadRules() {
+        rulesLoaded = false;
+        rules.clear();
+    }
+
+    public void ensureRulesLoaded() {
+        if (!rulesLoaded) {
```
jcrichlake, line 36: Should this be a boolean or just a check if there's any rules in the list

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
```diff
@@ -0,0 +1,60 @@
...
+            } catch (RuleLoaderException e) {
+                logger.logError(
+                        "Failed to load rules definitions from: " + RULES_DEFINITIONS_PATH, e);
+            }
+        }
+    }
+
+    public void validate(IBaseResource resource) {
+        logger.logDebug("Validating FHIR resource");
+        ensureRulesLoaded();
```
jcrichlake, line 51: Should this be inside of the constructor on line 24 to load the rules when we start the application?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
```diff
@@ -0,0 +1,60 @@
...
+        for (Rule rule : rules) {
+            if (rule.appliesTo(resource)) {
+                if (!rule.isValid(resource)) {
```
jcrichlake, line 54: Should you combine these together with an and ?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/ValidationRule.java
```diff
@@ -0,0 +1,97 @@
...
+        return conditions.stream()
+                .allMatch(
+                        condition -> {
+                            try {
+                                return fhirEngine.evaluateCondition(resource, condition);
+                            } catch (Exception e) {
+                                logger.logError(
+                                        "An error occurred while evaluating the condition: "
+                                                + condition,
+                                        e);
```
halprin, line 92: I'd include the name of this rule in the log message.

etor/src/main/resources/rule_definitions.json
```diff
@@ -0,0 +1,23 @@
...
+    },
+    {
+      "name": "requiredReceiverId",
+      "description": "ORM has required Card Number",
+      "warningMessage": "ORM doesn't have required Card Number",
+      "conditions": [
+        "Bundle.entry.resource.ofType(MessageHeader).event.code = 'O01'"
+      ],
+      "validations": [
+        "Bundle.entry.resource.ofType(Observation).where(code.coding.code = '57723-9').valueCodeableConcept.coding.code.exists()"
```
halprin, line 19: You'll need to confirm this, but FHIR Path allows you to not specify the "type" of a field.

```suggestion
        "Bundle.entry.resource.ofType(Observation).where(code.coding.code = '57723-9').value.coding.code.exists()"
```

halprin (commented):
Here's my first round of feedback.  More to come after stand-up.  This is some sweet sweet code!  I especially like the Java 8 streams.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
luis-pabon-tf: Changed it, it's now a check with one less var to keep track of.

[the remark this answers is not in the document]
luis-pabon-tf: Method FLATTENED!

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/Rule.java
```diff
@@ -0,0 +1,25 @@
+package gov.hhs.cdc.trustedintermediary.etor.ruleengine;
+
+import java.util.List;
+import org.hl7.fhir.instance.model.api.IBaseResource;
```
halprin, line 4: So... this is a Hapi FHIR import.  This means this validation framework is tightly coupled to the Hapi FHIR library.

1. Do we want that?
2. Could we add some abstraction to this?
3. Maybe we just rename all of these classes by prepending `Hapi`?

See what we wrote in ADR 13 (`/adr/013-wrapped-domain-objects.md`).

I'm not pushing you to a specific answer, and I want to know what your thoughts are on this.

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/orders/OrderControllerTest.groovy
```diff
@@ -16,6 +17,7 @@ class OrderControllerTest extends Specification {
         TestApplicationContext.init()
         TestApplicationContext.register(OrderController, OrderController.getInstance())
         TestApplicationContext.register(MetricMetadata, Mock(MetricMetadata))
+        TestApplicationContext.register(RuleEngine, Mock(RuleEngine))
```
halprin, line 20: We should probably put an assertion in some test (or a new test) that ensures we call the `RuleEngine`.  If we accidentally remove the call to the rules engine, I would hope a unit test would fail.

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/hapi/HapiFhirImplementationTest.groovy
```diff
@@ -0,0 +1,163 @@
...
+
+import ca.uhn.fhir.fhirpath.FhirPathExecutionException
+import gov.hhs.cdc.trustedintermediary.context.TestApplicationContext
+import org.hl7.fhir.instance.model.api.IBaseResource
+import org.hl7.fhir.r4.model.Bundle
+import org.hl7.fhir.r4.model.DiagnosticReport
+import org.hl7.fhir.r4.model.ServiceRequest
+import spock.lang.Specification
+
+class HapiFhirImplementationTest extends Specification {
```
halprin, line 11: This file is in the wrong location.  It needs to be in the `shared` subproject.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/wrappers/HapiFhir.java
```diff
@@ -13,4 +15,8 @@ <T extends IBaseResource> T parseResource(String fhirResource, Class<T> clazz)
             throws FhirParseException;
 
     String encodeResourceToJson(Object resource);
+
+    List<Base> evaluate(IBaseResource root, String expression);
```
halprin, line 19: This method, and the implementation in `HapiFhirImplementation` is only used in tests.  This method should be deleted and any tests that specifically test this method can be deleted.

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/hapi/HapiFhirImplementationTest.groovy
halprin: You are welcome to add tests for `HapiFhirImplementation`, but do realize it was meant to be a "humble object" and therefore was added to the JaCoCo code coverage ignore list.  See `/build.gradle:95`.  If you keep these tests, consider removing the JaCoCo ignore statement.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiFhirImplementation.java
```diff
@@ -1,17 +1,21 @@
...
+import org.hl7.fhir.r4.model.Base;
+import org.hl7.fhir.r4.model.BooleanType;
 
 /** Concrete implementation that calls the Hapi FHIR library. */
 public class HapiFhirImplementation implements HapiFhir {
 
     private static final HapiFhirImplementation INSTANCE = new HapiFhirImplementation();
-
     private static final FhirContext CONTEXT = FhirContext.forR4();
+    private static final IFhirPath pathEngine = CONTEXT.newFhirPath();
```
halprin, line 18: Make this all caps like the other fields nearby because this is a `static final` field.

```suggestion
    private static final IFhirPath PATH_ENGINE = CONTEXT.newFhirPath();
```

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
luis-pabon-tf: This can be done, but I have a concern in that the method needs the logger to be loaded, and that makes the injects need to be written above the construction.

halprin (commented):
Some more comments.  I'm happy to chat about them in detail with you!

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
halprin: I have a [similar comment below](https://github.com/CDCgov/trusted-intermediary/pull/943#discussion_r1535713313) with another possible solution.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
luis-pabon-tf: Done, no issues there

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleDefinition.java
luis-pabon-tf: Looks like this one was fixed

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
```diff
@@ -0,0 +1,60 @@
...
+/** Manages the application of rules loaded from a definitions file using the RuleLoader. */
+public class RuleEngine {
+
+    private final Path RULES_DEFINITIONS_PATH =
+            Path.of("../etor/src/main/resources/rule_definitions.json");
+    private boolean rulesLoaded = false;
+
+    private static final RuleEngine INSTANCE = new RuleEngine();
+
+    final List<Rule> rules = new ArrayList<>();
```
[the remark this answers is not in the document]
basiliskus: I thought having a list of rules instead of the map will give more flexibility and a consistent self-contained structure. It makes it easier to dynamically add/remove rules, which I think it will happen more often that accessing a specific rule by name. It will also preserve the order which may be important in the future

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import javax.inject.Inject;
+import org.hl7.fhir.instance.model.api.IBaseResource;
+
+/** Manages the application of rules loaded from a definitions file using the RuleLoader. */
+public class RuleEngine {
+
+    private final Path RULES_DEFINITIONS_PATH =
+            Path.of("../etor/src/main/resources/rule_definitions.json");
```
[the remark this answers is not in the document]
basiliskus: Good catch, I'll take a look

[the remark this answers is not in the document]
basiliskus: Actually, we tried using the `getClass().getClassLoader()` and had the issue running from the jar file. That's why we went with this relative path, which seems to work running off jar. May need to double check this with you later

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiFhirImplementation.java
luis-pabon-tf: Done! Couldn't commit the suggestion cause there were a few other instances

shared/src/main/java/gov/hhs/cdc/trustedintermediary/wrappers/HapiFhir.java
luis-pabon-tf: Before removing, I wanted to make sure this is not something that we will be needing in the near future

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/hapi/HapiFhirImplementationTest.groovy
luis-pabon-tf: Unignoring `HapiFhirImplementation` from codecoverage because I stand by those tests!

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/Rule.java
basiliskus: Good catch. I had pending task that I missed to review all the places where we were using the HapiFHIR library to make sure we decouple the dependency. I'll work on refactoring it

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/orders/OrderControllerTest.groovy
luis-pabon-tf: Added an assertion to the mock on the happy path test

shared/src/main/java/gov/hhs/cdc/trustedintermediary/wrappers/HapiFhir.java
luis-pabon-tf: YAGNI - done

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
```diff
@@ -0,0 +1,60 @@
...
+    public static RuleEngine getInstance() {
+        return INSTANCE;
+    }
+
+    public void unloadRules() {
+        rulesLoaded = false;
+        rules.clear();
+    }
+
+    public void ensureRulesLoaded() {
```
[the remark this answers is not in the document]
luis-pabon-tf: Added a sync sub-method that should take care of this. Leaving this one unresolved cause I'd like your confirmation that the implementation is correct. Tests are still passing

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
luis-pabon-tf: Added a potential fix on the related thread.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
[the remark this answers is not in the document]
luis-pabon-tf: Just pushed an update to replicate the logic in `EtorDomainRegistration`. Will resolve once all tests pass...

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
```diff
@@ -0,0 +1,60 @@
...
+        if (!rulesLoaded) {
+            logger.logDebug("Loading rules definitions from " + RULES_DEFINITIONS_PATH);
+            try {
+                var loadedRules = ruleLoader.loadRules(RULES_DEFINITIONS_PATH);
+                rules.addAll(loadedRules);
+                rulesLoaded = true;
+            } catch (RuleLoaderException e) {
+                logger.logError(
+                        "Failed to load rules definitions from: " + RULES_DEFINITIONS_PATH, e);
+            }
```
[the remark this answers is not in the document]
basiliskus: Correct. That's the idea. We don't want to stop the flow if there's an exception here. Specially considering we're only logging warnings as part of the validations

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/Rule.java
basiliskus: I refactored the code to decouple the hapifhir dependency

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
```diff
@@ -0,0 +1,59 @@
...
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+import javax.inject.Inject;
+
+/** Manages the application of rules loaded from a definitions file using the RuleLoader. */
+public class RuleEngine {
+
+    private static final RuleEngine INSTANCE = new RuleEngine();
```
jorg3lopez, line 14: I'm wondering if this will be a rules engine for global configuration? an example would be for validation of fhir format or required fields. So if it is, it would carry the same rules across all api requests. If it is not and we are using it as rules for every client, then the rules will differ from client to client. and each client will have to get a different instance of the engine per api request.

basiliskus: The rule engine implementation is flexible enough that we can have either global or partner-specific validations (in a big part because we're using FHIR Path which is very powerful). The way to do this would be using the `conditions` in the validation rule. A validation rule with empty conditions will apply to all fhir messages. A sender or receiver-specific one would define a condition like `Bundle.entry.resource.ofType(MessageHeader).destination.receiver.resolve().identifier.where(value = 'simulated-lab-id').exists()`

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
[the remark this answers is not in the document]
jorg3lopez: thoughts on using a synchronized block for writing or reading to the list? I think there is a synchronized list and a CopyOnWriteArrayList that can provide thread safety. We would have to see how it will impact the performance in order to use them as options. But if the set of rules will not change frequently, thoughts on using an immutable collection to not get hit on performance as much?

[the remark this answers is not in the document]
basiliskus: That's a good point. I don't expect the json file will be updated frequently so I don't think that would be a concern. I'm not very familiar with immutable collections in Java so I'd need to look into that

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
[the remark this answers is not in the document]
halprin: Yeah, this looks great!  Thanks, Luis!

shared/src/main/java/gov/hhs/cdc/trustedintermediary/wrappers/HapiFhir.java
halprin: Heck yeah!  💥

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/hapi/HapiFhirImplementationTest.groovy
halprin: Nice!  I'm all for tests!

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiFhirResource.java
```diff
@@ -0,0 +1,18 @@
+package gov.hhs.cdc.trustedintermediary.external.hapi;
+
+import gov.hhs.cdc.trustedintermediary.etor.ruleengine.FhirResource;
+import org.hl7.fhir.instance.model.api.IBaseResource;
+
+public class HapiFhirResource implements FhirResource<IBaseResource> {
```
halprin, line 6: JavaDocs please.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/FhirResource.java
```diff
@@ -0,0 +1,5 @@
+package gov.hhs.cdc.trustedintermediary.etor.ruleengine;
+
+public interface FhirResource<T> {
```
halprin, line 3: JavaDocs please.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleLoader.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import javax.inject.Inject;
+
+/** Manages the loading of rules from a definitions file. */
+public class RuleLoader {
+    private static final RuleLoader INSTANCE = new RuleLoader();
+    @Inject Formatter formatter;
+    @Inject Logger logger;
```
halprin, line 16: The `logger` is unused.  Should just delete this.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/ValidationRule.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private final Logger logger = ApplicationContext.getImplementation(Logger.class);
+    private final HapiFhir fhirEngine = ApplicationContext.getImplementation(HapiFhir.class);
+    private String name;
+    private String description;
+    private String violationMessage;
+    private List<String> conditions;
+    private List<String> validations;
+
+    public ValidationRule() {}
```
halprin, line 23: This is an unused constructor, but is required for JSON unmarshaling.  Let's add a comment describing this so someone doesn't accidentally remove this constructor.

halprin (commented):
Last couple of small things, but this is looking hot hot hot!  🔥

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/ValidationRule.java
basiliskus: Good point! Will do

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
jorg3lopez: Ah ok. When we do the work for client rules, we would need to look into the singleton design pattern as each client will need to have a different instance of the class.

basiliskus: There won't be a need for an instance per partner. All validation rules will be under the same instance. In the future this could be a database instead of the json file. Or azure blob storage

jorg3lopez (approved):
This is is nice, great job!

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
jorg3lopez: I see, so retrieve the rules from the database that has the client settings. This will allow us to make the settings changes without needing to deploy. I like the sound of that

PR #972, opened 2024-03-25
Results Metadata Update

saquino0827:
# Add a PR title

Updates around metadata which include:
- Refactor metadata to use a dedicated `PartnerMetadataConverter` over the original `OrderConverter`.
- Add `sent submission id` as a new column when retrieving Metadata from the Metadata endpoint.
- Update unit and e2e tests.

## Issue

## Checklist


e2e/src/test/groovy/gov/hhs/cdc/trustedintermediary/e2e/MetadataTest.groovy
```diff
@@ -29,24 +29,23 @@ class MetadataTest extends Specification {
...
-            "order ingestion",
+            "ingestion",
             "payload hash",
             "delivery status",
-            "status message"
+            "status message",
+            "message type",
+            "sent submission id"
         ].each { String metadataKey ->
+            println(metadataKey)
```
jherrflexion, line 48: Do we still need this?

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/hapi/HapiOrderConverterTest.groovy
```diff
@@ -351,6 +351,6 @@ class HapiOrderConverterTest extends Specification {
         result.getIssue().get(4).diagnostics == time.toString()
         result.getIssue().get(5).diagnostics == PartnerMetadataStatus.DELIVERED.toString()
         result.getIssue().get(6).diagnostics == failureReason
-        result.getIssue().get(4).details.text.contains(messageType.toString().toLowerCase())
+        result.getIssue().get(7).diagnostics == messageType.toString()
```
jherrflexion, line 354: Should this have: 
 result.getIssue().get(8).diagnostics == "sentSubmissionId"

e2e/src/test/groovy/gov/hhs/cdc/trustedintermediary/e2e/MetadataTest.groovy
```diff
@@ -17,7 +18,6 @@ class MetadataTest extends Specification {
         given:
         def expectedStatusCode = 200
         def inboundSubmissionId = UUID.randomUUID().toString()
-        def outboundSubmissionId = "1234567890"
```
basiliskus, line 20: Given you only have one submission id in the test that should work for inbound and outbound regardless, I think it's better to rename `inboundSubmissionId` to `submissionId`. Same for the other tests

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiPartnerMetadataConverter.java
```diff
@@ -0,0 +1,79 @@
...
+                .getIssue()
+                .add(
+                        createInformationIssueComponent(
+                                "message type", metadata.messageType().toString()));
+
+        operation
+                .getIssue()
+                .add(
+                        createInformationIssueComponent(
+                                "sent submission id", metadata.sentSubmissionId()));
```
halprin, line 62: Let's rename this as `outbound submission id`.  In the past, we've referred to it as "outbound" instead of "sent".

```suggestion
                                "outbound submission id", metadata.sentSubmissionId()));
```

halprin, line 62: Since we're including the outbound submission ID, let's also add the inbound submission ID.  Users can call the metadata API with either the inbound or outbound submission ID and get the same information, so we should probably just decide to include both.

e2e/src/test/groovy/gov/hhs/cdc/trustedintermediary/e2e/MetadataTest.groovy
```diff
@@ -29,24 +29,23 @@ class MetadataTest extends Specification {
 
         when:
         def inboundMetadataResponse = metadataClient.get(inboundSubmissionId, true)
-        def outboundMetadataResponse = metadataClient.get(outboundSubmissionId, true)
```
halprin, line 32: We support both the inbound or outbound ID being passed in, so we need to keep this test code here.

halprin (changes requested):

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiPartnerMetadataConverter.java
saquino0827: The inbound submission id is included as the `ID` field. Do we want to rename the ID field?

e2e/src/test/groovy/gov/hhs/cdc/trustedintermediary/e2e/MetadataTest.groovy
saquino0827: The test will change to include the `outboundSubmissionId` based on the comment below.

e2e/src/test/groovy/gov/hhs/cdc/trustedintermediary/e2e/MetadataTest.groovy
saquino0827: Good catch

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiPartnerMetadataConverter.java
saquino0827: Nevermind, I see why this is needed and I also think it's a good idea to add.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiPartnerMetadataConverter.java
```diff
@@ -0,0 +1,86 @@
...
+    private static final HapiPartnerMetadataConverter INSTANCE = new HapiPartnerMetadataConverter();
+
+    public static HapiPartnerMetadataConverter getInstance() {
+        return INSTANCE;
+    }
+
+    private HapiPartnerMetadataConverter() {}
+
+    @Override
+    public FhirMetadata<?> extractPublicMetadataToOperationOutcome(
```
luis-pabon-tf, line 20: While this was an existing method, I think it should get some apidoc

saquino0827: I think the open api docs generate for the metadata endpoint. But also, I added docs to the class interface [here](https://github.com/CDCgov/trusted-intermediary/blob/d557068b91a8e6cc90916b782d1250e8fdeea39f/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/partner/PartnerMetadataConverter.java).

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
```diff
@@ -214,7 +217,7 @@ DomainResponse handleMetadata(DomainRequest request) {
             }
 
             FhirMetadata<?> responseObject =
-                    orderConverter.extractPublicMetadataToOperationOutcome(
+                    partnerMetadataConverter.extractPublicMetadataToOperationOutcome(
```
halprin, line 220: Thanks for pulling this out of the order converter!

PR #978, opened 2024-03-26
Make the public IP address used by the VPN a standard static IP

jcrichlake:
Apparently we need this update.

The creation of VPNs now suddenly require a "standard" (and therefore "static") public IP address.


halprin (approved):
So hot.

halprin:
With my expirementation in the Dev environment, this is turning into something bigger that I don't think needs to be solved right now.  I feel like we should figure this out post-MVP.

Issue #621, opened 2023-10-27
Linking Orders to Results


scleary1cs:
# Story
As a hospital (or lab), so that I can know if a result has been returned for an order I placed (or vice versa), I want order and results metadata to be provided together.

## Pre-conditions
- Orders and results processing are already supported.
- Any modification to the result message based on the order is out of scope of this story.
- Calculating stats for unsolicited orders is out of scope of this story.

## Acceptance Criteria
- [x] When retrieving order metadata, metadata for the corresponding result is included.
- [x] When retrieving result metadata, metadata for the corresponding order is included.

## Note
- Added bonus based on how this was implemented: results to results and orders to orders matching; however test cases would need to be created in the future if that becomes a real use case

### Research
- [x] What unique identifier can we use to link the order and result?
  - [x] Is there already a way to do this in the HL7 specs? **Placer order number: ORC.2 & OBR.2** (they are required, and should be the same value)
    - [x] Which one should we rely on? **ORC.2** & **MSH.4** (decided not to also key based on Patient ID for now) 
    - [x] What does that look like in FHIR?
    - [x] Ask what AIMS is doing for this: They are assigning an internal ID to each order received which is what they use to link back to result. For that internal ID, they are using the placer order number (ORC-2) from the order message to assign that internal ID -- but one that also included the assigning authority.
    - [x] Consider how this may work under reflex testing, re-orders, amended orders, etc.
- [x] There is the potential that this will also allow us to take data from an order and add it to a result, but this may or may not be something the intermediary wants to support.

###  Engineering
- [x] Stream 2 to sync up with Stream 1
  - [x] Test integration of different PRs
- [x] Add fields to metadata table - @halprin 
  - [x] Add `placer_order_number`, `sending_application_id`, `sending_facility_id`, `receiving_application_id`, `receiving_facility_id` to the partner metadata table - @halprin 
  - [x] Extract fields from FHIR message and write them into the metadata table @jorg3lopez 
    - [x] Orders
      - [x] Extract placer order number @jherrflexion 
      - [x] Extract sending application details @jorg3lopez
        - [x]  Refactor to SendingApplicationDetails Object
      - [x] Extract sending facility details @jorg3lopez 
        - [x]  Refactor to SendingFacilityDetails Object 
      - [x] Extract receiving application details @jorg3lopez 
      - [x] Extract receiving facility details @jorg3lopez 
        - [x]  Refactor to ReceivingFacilityDetails Object
[...]

PR #982, opened 2024-03-27
Extract Fields From FHIR Bundle

jorg3lopez:
# Extract Fields From FHIR Bundle

This PR adds the logic to extract the following fields from the FHIR bundle:
- placer order number
- sending application id
- sending facility id
- receiving application id
- receiving facility id

These fields are extracted from orders and results.

## Issue
#621 

## Checklist


halprin:
I talked with the HL7 SMEs.  They stated that the MSH fields' most important sub-fields are the first two, and the ORC field's most important sub-fields are the first three.  This basically means that we shouldn't just return a single ID for each of these new methods we're adding.  Honestly, we might as well include all of the sub-fields.

We should either add additional methods for each sub-field, have the existing method return all the important sub-fields with some delimiter (e.g. `1234^other_stuff^moof` using `^` as the delimiter), have the existing method return a POJO (Plain Old Java Object) that has the sub-fields, or some other solution that I'm not thinking of.

jorg3lopez:
> I talked with the HL7 SMEs. They stated that the MSH fields' most important sub-fields are the first two, and the ORC field's most important sub-fields are the first three. This basically means that we shouldn't just return a single ID for each of these new methods we're adding. Honestly, we might as well include all of the sub-fields.
> 
> We should either add additional methods for each sub-field, have the existing method return all the important sub-fields with some delimiter (e.g. `1234^other_stuff^moof` using `^` as the delimiter), have the existing method return a POJO (Plain Old Java Object) that has the sub-fields, or some other solution that I'm not thinking of.

Yes. I noticed that we would have to get more than a sub-field. I'm taking the namespace-id, universal-id, and universal-type-id for the most part. I already have a method that concatenates and joins the fields with a caret. I still haven't looked into isolating each id for later use. Right now, I'm only spitting out something like: `Epic^1.2.840.114350.1.13.145.2.7.2.695071^ISO`

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiOrder.java
```diff
@@ -42,4 +50,170 @@ public String getPatientId() {
...
+                                } else if ("https://reportstream.cdc.gov/fhir/StructureDefinition/universal-id-type"
+                                        .equals(extension.getUrl())) {
+                                    universalIdType =
+                                            Objects.toString(
+                                                    extension.getValue().primitiveValue(), "");
+                                }
+                            }
+
+                            return concatenateWithCaret(
+                                    namespaceId, universalId, universalIdType, endpoint);
```
jorg3lopez, line 91: I'm not sure if the endpoint is needed here. I think we can go without it, thoughts?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/Order.java
```diff
@@ -11,4 +11,14 @@ public interface Order<T> {
...
+
+    String getPlacerOrderNumber();
+
+    String getSendingApplicationDetails();
+
+    String getSendingFacilityDetails();
+
+    String getReceivingApplicationDetails();
+
+    String getReceivingFacilityId();
```
basiliskus, line 23: For consistency, should we have `getReceivingFacilityDetails` instead of `getReceivingFacilityId`?

basiliskus: As an idea, instead of `String` we could return a custom object that will split namespace, id and type

jorg3lopez: Yes. We haven't started working on this method. Initially, we had  the methods return single IDs, now they are returning a string with multiple IDs. Yes, we are looking at the possibility of having a pojo and using fhir path as well.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/Order.java
```diff
@@ -11,4 +11,14 @@ public interface Order<T> {
     String getFhirResourceId();
 
     String getPatientId();
+
+    String getPlacerOrderNumber();
+
+    String getSendingApplicationDetails();
```
basiliskus, line 17: Could we return a `MessageHdDataType` here instead of `String`?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiMessageHelper.java
```diff
@@ -0,0 +1,282 @@
...
+        return fhirEngine.getStringFromFhirPath(
+                messageBundle, fhirPaths.get(RECEIVING_APPLICATION_UNIVERSAL_ID_TYPE));
+    }
+
+    private Map<String, String> loadFhirPaths() {
+        Map<String, String> tempPaths = new HashMap<>();
+        tempPaths.put(
+                RECEIVING_FACILITY_NAMESPACE,
+                """
+            Bundle.entry.resource.ofType(MessageHeader).destination.receiver.resolve().identifier.where(
```
basiliskus, line 211: I think it'd be good elevate this fhirpaths into static strings at the beginning of the class. I'm thinking we could refactor later so they are accessible in other parts of the project and we have a central place to define them

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/OrderMock.groovy
```diff
@@ -10,25 +10,61 @@ class OrderMock<T> implements Order<T> {
...
     }
 
     @Override
     String getPatientId() {
         return patientId
     }
+
+    @Override
+    String getPlacerOrderNumber() {
+        return this.placerOrderNumber
```
basiliskus, line 48: I've seen some places in the project where we don't use `this` explicitly. Do we have a convention for this?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/Order.java
jorg3lopez: Yes. I agree with this approach. This is what all the methods with details will return.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiMessageHelper.java
jorg3lopez: We took the map out and went with this approach. Good suggestion!

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/OrderMock.groovy
jorg3lopez: I have seen it as well. Personally, I like to use `this` as it adds a little more reference to the viewer. Maybe we can decide on a convention to follow for the project.

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/hapi/HapiMessageHelperTest.groovy
saquino0827, line 1: I feel reluctant to add a `HapiMessageHelperTest` class. Mainly because the purpose of this helper was to reduce code duplication of `HapiOrder` and `HapiResult` which we have test cases for both of those classes. I'm in favor of remove this class as I feel it's only duplicate testing at this point. Thoughts?

jorg3lopez: I concur. @halprin is ok with not including it as well. I'll remove the test file.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiOrder.java
```diff
@@ -11,14 +14,17 @@
  */
 public class HapiOrder implements Order<Bundle> {
 
+    private final HapiMessageHelper MESSAGE_HELPER =
```
basiliskus, line 17: Should this name be in lowercase given it's not static? Was it meant to be static?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiResult.java
```diff
@@ -1,24 +1,78 @@
...
+import gov.hhs.cdc.trustedintermediary.context.ApplicationContext;
+import gov.hhs.cdc.trustedintermediary.etor.messages.MessageHdDataType;
 import gov.hhs.cdc.trustedintermediary.etor.results.Result;
+import java.util.function.Supplier;
 import org.hl7.fhir.r4.model.Bundle;
 
 /** Filler concrete implementation of a {@link Result} using the Hapi FHIR library */
 public class HapiResult implements Result<Bundle> {
 
+    private final HapiMessageHelper MESSAGE_HELPER =
```
basiliskus, line 12: Same comment as above

basiliskus: Also, seems like there is a lot of shared code between `HapiOrder` and `HapiResult`

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/hapi/HapiResultTest.groovy
```diff
@@ -1,16 +1,30 @@
 package gov.hhs.cdc.trustedintermediary.external.hapi
 
-import org.hl7.fhir.r4.model.Bundle
+import gov.hhs.cdc.trustedintermediary.context.TestApplicationContext
+import gov.hhs.cdc.trustedintermediary.etor.messages.MessageHdDataType
+import gov.hhs.cdc.trustedintermediary.wrappers.HapiFhir
+import org.hl7.fhir.r4.model.*
```
basiliskus, line 6: Reminder to not use wildcard imports. You should be able to change this setting in your IDE to avoid it

basiliskus (commented):
Great job on this PR! I just left a few minor comments

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiOrder.java
jorg3lopez: No. Conventionally, final variables are all uppercase and use underscore. Sonarcloud is just angry and wants us to follow the camel case convention. I think we can leave it as is, thoughts?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiResult.java
jorg3lopez: Same...
No. Conventionally, final variables are all uppercase and use underscore. Sonarcloud is just angry and wants us to follow the camel case convention. I think we can leave it as is, thoughts?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiOrder.java
basiliskus: Sure, if that's the naming convention in Java then we should keep it. Thanks Jorge

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiResult.java
jorg3lopez: Yes on the similar codes. There is a Message interface that both Result and Order interfaces extend. HapiOrder implements Order and HapiResult implements Result. We can discuss if we no longer need HapiOrder or HapiResult and just create HapiMessage to do the work. But we already have two separations for Orders and Results., thoughts?

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/hapi/HapiResultTest.groovy
jorg3lopez: Good catch. I hope everyone is using the githooks.

jorg3lopez: fixed.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiResult.java
basiliskus: I know there are some differences between HapiOrder and HapirResult so they may not be able to completely merge

luis-pabon-tf: Checking what we can merge by having a common parent.

basiliskus (approved):
Great work!

saquino0827 (approved):

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiFhirResource.java
```diff
@@ -16,4 +16,9 @@ public HapiFhirResource(IBaseResource innerResource) {
     public IBaseResource getUnderlyingResource() {
         return innerResource;
     }
+
+    @Override
+    public String getFhirResourceId() {
+        return innerResource.getIdElement().getIdPart();
```
jorg3lopez, line 22: Any idea why it shows that this method is not covered by the tests in HapiOrderTest and HapiResultTest?

saquino0827: I merged this before seeing this was commented. Let's keep track of test coverage for this and make another PR to fix it if necessary.

PR #983, opened 2024-03-27
621: Order/Result Linking Fields Added to Metadata Table

halprin:
# Order/Result Linking Fields Added to Metadata Table

Added new columns to the `metadata` table to support our order and result linking.

## Issue

#621.

### DIFF
--- a/etor/databaseMigrations/metadata.yml
+++ b/etor/databaseMigrations/metadata.yml
@@ -84,3 +84,28 @@ databaseChangeLog:
               - column:
                   name: message_type
                   type: message_type
+  - changeSet:
+      id: 5
+      author: halprin
+      labels: update-metadata-table
+      context: metadata
+      comment: update partner metadata table for order and result linking
+      changes:
+        - addColumn:
+            tableName: metadata
+            columns:  # the size of the varchars below are based on the field's HL7 spec size
+              - column:
+                  name: placer_order_number
+                  type: varchar(427)
+              - column:
+                  name: sending_application_id
+                  type: varchar(227)
+              - column:
+                  name: sending_facility_id
+                  type: varchar(227)
+              - column:
+                  name: receiving_application_id
+                  type: varchar(227)
+              - column:
+                  name: receiving_facility_id
+                  type: varchar(227)


etor/databaseMigrations/metadata.yml
```diff
@@ -84,3 +84,28 @@ databaseChangeLog:
...
+      labels: update-metadata-table
+      context: metadata
+      comment: update partner metadata table for order and result linking
+      changes:
+        - addColumn:
+            tableName: metadata
+            columns:  # the size of the varchars below are based on the field's HL7 spec size
+              - column:
+                  name: placer_order_number
+                  type: varchar(427)
```
basiliskus, line 99: 427 seems high, do we expect for the placer order number to take a lot of space?

basiliskus (commented):
I left a couple of comments on the maximums for the fields, otherwise looks good to me

etor/databaseMigrations/metadata.yml
jcrichlake: I think his comment on 96 is that he's basing it on the HL7 spec

jcrichlake (approved):

etor/databaseMigrations/metadata.yml
basiliskus: Right, I missed that. This is an open question, but should we always use the maximum from the HL7 specs for these database fields if they seem unusually high? We could also consider using the `text` data type which doesn't specify the max

halprin: For the `placer_order_number`, I used this: https://hl7-definition.caristix.com/v2/HL7v2.5.1/Fields/ORC.2.  Add up all of the fields and then add three for the `^` separators.

For the other four, I used this: https://hl7-definition.caristix.com/v2/HL7v2.5.1/Segments/MSH.

You can see that there is a lot more than _just_ the ID in that documentation.  There's also the namespace stuff.  Do we think that is also important to capture?

I'm open to reducing the size.

basiliskus: I think we'll only store `ORC-2.3` there, but we can confirm that with Daniel. In any case, it's not a big deal and we can adjust it later if we decide on that

halprin: Talked with the SMEs.

For the MSH headers, the important fields are the first two.  So 20 + 199 + 1 = 220.

For the ORC placer order number, the important fields are the first three.  So 199 + 20 + 199 + 2 = 420.

But here's the thing: the last field is so small that I might as well include it.

Realistically, I can see the full size _not_ being taken up, but I'd hate to truncate if, by chance, the message includes a full size field.

halprin: For now, I'll merge this PR since I have an approval.  As @basiliskus states, if we need to change them later on, we will.

PR #986, opened 2024-03-28
621: Make the `DbDoa` More Generic

halprin:
# Make the `DbDao` More Generic

A bit of DevEx clean-up to better support story 621.  This makes the `DbDoa` more generic.  I'm trying to make the `DbDoa` not need to know about `PartnerMetadata` which makes the `DatabasePartnerMetadataStorage` the exclusive converter between `PartnerMetadata` and how to store it in the database.  This is step one.  Later on changes will address the reading.

## Issue

#621.

## Checklist


etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/DatabasePartnerMetadataStorage.java
```diff
@@ -38,18 +42,66 @@ public Optional<PartnerMetadata> readMetadata(final String uniqueId)
...
+
+        DbColumn dbColumn1 =
+                new DbColumn(
+                        "received_message_id",
+                        metadata.receivedSubmissionId(),
+                        false,
+                        Types.VARCHAR);
+        values.add(dbColumn1);
+
+        DbColumn dbColumn2 =
```
jcrichlake, line 55: Could you refactor this to only use one dbColumn? Just reinitialize and add the column at each step. Then we don't have to have 11+ different columns

halprin: Definitely.

basiliskus: Could this be another option to make it more concise and easier to maintain:
```
List<DbColumn> values = Stream.of(
        new DbColumn("received_message_id", metadata.receivedSubmissionId(), false, Types.VARCHAR),
        new DbColumn("sent_message_id", metadata.sentSubmissionId(), true, Types.VARCHAR),
        new DbColumn("sender", metadata.sender(), false, Types.VARCHAR),
        ...
).collect(Collectors.toList());
```

halprin: @basiliskus, I think the only downside is that some of the `DbColumn`s are conditionally created with their value.  See `time_received`.  So, I could create all the conditional values up front and then do the `Stream.of(` call.  But then the creation of the conditional value is create farther away from where it is used.  Thoughts?

basiliskus: This could be a good use case for ternary inline statements:
`new DbColumn("time_received", metadata.timeReceived() != null ? Timestamp.from(metadata.timeReceived()) : null, false, Types.TIMESTAMP),`

halprin: Oh, right!

halprin: Done.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/PostgresDao.java
```diff
@@ -29,62 +29,63 @@ public static PostgresDao getInstance() {
...
-            statement.setTimestamp(7, timestampDelivered);
-
-            String deliveryStatusString = null;
-            if (deliveryStatus != null) {
-                deliveryStatusString = deliveryStatus.toString();
+        StringBuilder sqlStatementBuilder =
+                new StringBuilder("INSERT INTO ").append(tableName).append(" (");
+
+        values.forEach(dbColumn -> sqlStatementBuilder.append(dbColumn.name()).append(", "));
+        sqlStatementBuilder.delete(
```
basiliskus, line 38: This is minor, but since this way of removing the last `,` is used in multiple places, we could have a private method to do this

basiliskus: My only concern with this approach is that it's harder to read what the query is doing because it's more fragmented

basiliskus (approved):
Left a couple of comments, otherwise looks good to me

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/PostgresDao.java
halprin: I like this idea of having the private method.  I think I'll also add a code comment that gives an example SQL statement that helps us understand what is being constructed.

PR #988, opened 2024-03-29
621: Implement check for linked orders/results

basiliskus:
# Implement check for linked orders/results

- Created `MessageLink` as an abstraction of message links
- Check for messages to link given a received submission id. Use partner metadata (placerOrderNumber, sendingApplicationId and sendingFacilityId) to match to other submissions
  - If there's no match, don't create a link
  - If there's a match and there's an existing message link, add submission id to link
  - If there's a match and there's no existing message link, create a new link with the matched submission ids

## Issue

#621 

## Checklist


etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/partner/PartnerMetadataOrchestrator.java
```diff
@@ -143,43 +144,45 @@ public Optional<PartnerMetadata> getMetadata(String receivedSubmissionId)
...
-                rsStatus = parsedResponseBody[1];
-                rsMessage = parsedResponseBody[2];
-                timeDelivered = parsedResponseBody[3];
-            } catch (ReportStreamEndpointClientException | FormatterProcessingException e) {
-                throw new PartnerMetadataException(
-                        "Unable to retrieve metadata from RS history API", e);
-            }
 
-            var ourStatus = ourStatusFromReportStreamStatus(rsStatus);
+        if (!metadataIsStale(partnerMetadata) || sentSubmissionId == null) {
```
jcrichlake, line 148: Is this what we want? Will sentSumbissionId ever get updated in this case?

basiliskus: I just inverted the if-statement to avoid nesting, didn't change the actual logic

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/PostgresDao.java
```diff
@@ -128,6 +128,62 @@ public PartnerMetadata fetchMetadata(String submissionId) throws SQLException {
...
+            }
+
+            return metadataSet;
+        }
+    }
+
+    @Override
+    public Set<PartnerMetadata> fetchLinkedMessages(String messageId) throws SQLException {
+        var sql =
+                """
```
jcrichlake, line 163: Do we need a nested query here? I would think we can just do a 1 level query to get back the results we need.

basiliskus: I could use a self-join instead, or do you have another idea?

basiliskus: Changed to use join instead

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/metadata/partner/PartnerMetadataOrchestrator.java
basiliskus: Reverted this change for now to reduce noise in this PR

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messagelink/MessageLink.java
```diff
@@ -0,0 +1,53 @@
...
+    public UUID getLinkId() {
+        return linkId;
+    }
+
+    public void setMessageIds(Set<String> messageIds) {
+        this.messageIds = new HashSet<>(messageIds);
+    }
+
+    public Set<String> getMessageIds() {
+        return Collections.unmodifiableSet(messageIds);
```
halprin, line 43: You've put some protections in place in this class to defend against modifications.  For example, above use the copy constructor to get your own copy of the messageIds.  Here, you return a new version of the set that is unmodifiable.

Was there a reason for this?  I don't have an opinion, but I'm curious what you're thinking was.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/SendOrderUseCase.java
```diff
@@ -50,7 +50,8 @@ public void convertAndSend(final Order<?> order, String receivedSubmissionId)
                 EtorMetadataStep.ETOR_PROCESSING_TAG_ADDED_TO_MESSAGE_HEADER);
 
         String sentSubmissionId = sender.send(omlOrder).orElse(null);
-
         sendMessageHelper.saveSentMessageSubmissionId(receivedSubmissionId, sentSubmissionId);
+
+        sendMessageHelper.linkMessage(receivedSubmissionId);
```
halprin, line 55: Do we want to wait until the very end to link?  Or do it right after saving the metadata given the received ID, which is towards the beginning?

halprin, line 55: Also, what about adding this to the Send Result Usecase?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/PostgresDao.java
```diff
@@ -137,6 +142,91 @@ public PartnerMetadata fetchMetadata(String submissionId)
...
+                INSERT INTO message_link (link_id, message_id)
+                SELECT ?, ?
+                WHERE NOT EXISTS (
+                    SELECT 1 FROM message_link WHERE message_id = ?
+                );
+                """;
+
+        try (Connection conn = connectionPool.getConnection();
+                PreparedStatement statement = conn.prepareStatement(sql)) {
+            UUID linkId = Optional.ofNullable(messageLink.getLinkId()).orElse(UUID.randomUUID());
```
halprin, line 219: What do you think about moving this logic to the `PartnerMetadataOrchistrator`?

E.g.
`messageLinkStorage.saveMessageLink(new MessageLink(null, messageIds));` is updated to be `messageLinkStorage.saveMessageLink(new MessageLink(UUID.randomUUID(), messageIds));`

halprin (commented):
More comments.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messagelink/MessageLink.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.UUID;
+
+/**
+ * This class represents a link between messages. Each link has a unique ID and is associated with a
+ * set of message IDs to link.
+ */
+public final class MessageLink {
```
jorg3lopez, line 12: Thoughts on making this a record?

basiliskus: Not sure if it makes sense here as we need to manage state for message ids. It's not supposed to be immutable

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messagelink/MessageLink.java
basiliskus: I based this off of an example. I was just trying to handle Sets in Java, which I don't have much experience with so I'm not very familiar with the syntax. Now that you mention it, I don't think it's really necessary

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/PostgresDao.java
```diff
@@ -137,6 +142,91 @@ public PartnerMetadata fetchMetadata(String submissionId)
...
+
+        if (!messageIds.isEmpty()) {
+            return Optional.of(new MessageLink(linkId, messageIds));
+        } else {
+            return Optional.empty();
+        }
+    }
+
+    @Override
+    public void insertMessageLink(MessageLink messageLink) throws SQLException {
```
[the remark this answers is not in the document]
basiliskus: I'll take a look

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/PostgresDao.java
basiliskus: I'll take a look

jorg3lopez: I agree. After calling `getMessageLink` from the `messageLinkStorage`, the orchestrator can follow up with this logic before calling `saveMessageLink` from the`messageLinkStorage`.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messagelink/MessageLink.java
jorg3lopez: Yes. What remains immutable is the reference to the Set object, but you can modify the contents of the Set itself.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/PostgresDao.java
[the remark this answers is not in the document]
basiliskus: Done. One possible issue I see is that we need to open/close the db connection for each insert in the message_link table, which before was only one open/close per group of message ids

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/PostgresDao.java
basiliskus: Makes sense. Updated it

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messagelink/MessageLinkException.java
```diff
@@ -0,0 +1,13 @@
...
+/** This exception is thrown when there is an error linking messages. */
+public class MessageLinkException extends Exception {
+
+    public MessageLinkException(String message, Throwable cause) {
+        super(message, cause);
+    }
+
+    public MessageLinkException(String message) {
+        super(message);
+    }
```
halprin, line 12: This constructor can be removed because it is only called in tests.  Those tests that call this constructor can instead call the other constructor and pass in `new Exception()` for the second argument.

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/messagelink/MessageLinkTest.groovy
```diff
@@ -0,0 +1,117 @@
...
+
+import gov.hhs.cdc.trustedintermediary.context.TestApplicationContext
+import spock.lang.Specification
+
+class MessageLinkTest extends Specification {
+
+    def setup() {
+        TestApplicationContext.reset()
+        TestApplicationContext.init()
+    }
```
halprin, line 11: There's no `@Inject` or calls to the `ApplicationContext` in `MessageLink`.  You can remove this set-up.

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/messagelink/MessageLinkTest.groovy
```diff
@@ -0,0 +1,117 @@
...
+        given:
+        def messageLink = new MessageLink()
+        def linkId = UUID.randomUUID()
+
+        when:
+        messageLink.setLinkId(linkId)
+
+        then:
+        messageLink.getLinkId() == linkId
+    }
```
halprin, line 104: To test your classic setters and getters, you can use the `PojoTestUtils` helper class.

halprin (commented):
This looks really good!

halprin (approved):
Sweet!  Great changes to the PostgresDao and thanks for moving the logic for reading the links out of the PostgresDao.  That will make it easier to migrate once my PR is merged.

PR #997, opened 2024-04-01
341: Fix Validation Rule Engine concurrency issues

basiliskus:
# Fix Validation Rule Engine concurrency issues

- Fixed concurrency issues when loading rules
- Fixed validation rule fhirpath for "Message has required receiver id". It was previously matching receiver namespace as well

## Issue

#341 

## Checklist

### DIFF
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
@@ -28,7 +28,7 @@ public class RuleEngine {
         rules.clear();
     }
 
-    public void ensureRulesLoaded() {
+    public synchronized void ensureRulesLoaded() {
         if (rules.isEmpty()) {
             loadRules();
         }
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
@@ -28,9 +28,13 @@ public class RuleEngine {
         rules.clear();
     }
 
-    public synchronized void ensureRulesLoaded() {
+    public void ensureRulesLoaded() {
         if (rules.isEmpty()) {
-            loadRules();
+            synchronized (this) {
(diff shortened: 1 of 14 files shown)


etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
```diff
@@ -28,7 +28,7 @@ public void unloadRules() {
         rules.clear();
     }
 
-    public void ensureRulesLoaded() {
+    public synchronized void ensureRulesLoaded() {
```
halprin, line 31: This is in a performance sensitive flow during API flows, and synchronizing here will slow down concurrent calls to our API.

We can prevent this bottleneck by doing "double-checked locking".

```java
public void ensureRulesLoaded() {
    if (rules.isEmpty()) {
        synchronized(this) {
            if (rules.isEmpty()) {
                loadRules();
            }
        }
    }
}
```

This also allows us to remove `synchronized` from the `loadRules` method below.

halprin (changes requested):

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
luis-pabon-tf: That's a really good catch! Updating...

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngineTest.groovy
```diff
@@ -39,6 +39,27 @@ class RuleEngineTest extends Specification {
         1 * mockRuleLoader.loadRules(_ as String) >> [Mock(Rule)]
     }
 
+    def "ensureRulesLoaded loads rules only once on multiple threads"() {
+        given:
+        def threadsNum = 2
```
halprin, line 44: Let's up this to 10 so we have more threads in contention for the lock.  It's easy for 2 to not execute the method at the same time.

```suggestion
        def threadsNum = 10
```

halprin (approved):
Looks great!

PR #1019, opened 2024-04-12
621: Generic Read Metadata

halprin:
# Generic Read Metadata

Made reading from the database generic.  It used to be highly coupled to the metadata concept.  It can now support arbitrary SQL statements and conversions.

## Issue

#621.

## Checklist


etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/DatabasePartnerMetadataStorage.java
```diff
@@ -35,12 +45,25 @@ public static DatabasePartnerMetadataStorage getInstance() {
...
         try {
-            PartnerMetadata data = (PartnerMetadata) dao.fetchMetadata(uniqueId);
-            return Optional.ofNullable(data);
+            PartnerMetadata metadata =
+                    dao.fetchFirstData(
+                            connection -> {
+                                try {
+                                    PreparedStatement statement =
+                                            connection.prepareStatement(
+                                                    "SELECT * FROM metadata where received_message_id = ? OR sent_message_id = ?");
```
basiliskus, line 54: At some point I think we should separate SQL code from Java code. In the meantime I've been placing the sql code in a variable at the start of the method where it's used, but we should probably start putting them in a more centralized place soon. It could be external files or java constants

basiliskus (approved):
Looks great! I just noticed some test coverage is missing for the exception path in fetchManyData

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/DatabasePartnerMetadataStorage.java
```diff
@@ -35,12 +45,25 @@ public static DatabasePartnerMetadataStorage getInstance() {
...
             throws PartnerMetadataException {
         try {
-            PartnerMetadata data = (PartnerMetadata) dao.fetchMetadata(uniqueId);
-            return Optional.ofNullable(data);
+            PartnerMetadata metadata =
+                    dao.fetchFirstData(
+                            connection -> {
+                                try {
+                                    PreparedStatement statement =
+                                            connection.prepareStatement(
```
jorg3lopez, line 53: This is nice. For some time I've been thinking of having a wrapper class that will handle the prepared statements. Thoughts on having the wrapper class so that both storage classes can use it with the dao?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/ResultSetIterator.java
```diff
@@ -0,0 +1,41 @@
...
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+/**
+ * ResultSetIterator iterates over a SQL {@link ResultSet}. This helps stream over a {@link
+ * ResultSet}.
+ */
+public class ResultSetIterator implements Iterator<ResultSet> {
```
jorg3lopez, line 12: Thoughts on implementing the `AutoCloseable` interface to manage the closure of `ResultSet` in case the iterator is not fully traversed until the end?

halprin: That is handled outside of the iterator in `PostgresDao#fetchData` and `PostgresDao#fetchManyData` already.  We put the `ResultSet` inside the try-with-resources immediately as we get the `ResultSet`.

jorg3lopez: I see it.

jorg3lopez (approved):
Great job, looks good.

jorg3lopez:
Nice work, the only thing I see that might need attention is the test coverage. Other than that, everything looks good.

shared/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/jjwt/JjwtEngineTest.groovy
```diff
@@ -14,18 +14,14 @@ class JjwtEngineTest extends Specification {
 
     def "readPrivateKey works"() {
         given:
-        def expected = """SunRsaSign RSA private CRT key, 2048 bits
```
halprin, line 17: For some reason this test started to fail in GitHub for me.  I decided to refactor the test to not depend on `toString()`.

basiliskus (approved):
Looks good to me!

Issue #989, opened 2024-04-01
Metadata Ingestion Time 


scleary1cs:
# Story
As a hospital or PHL, to know when the Intermediary received the message include ingestion time for the metadata in the ResportStream service.

## Pre-conditions
- [ ] NA

## Acceptance Criteria
- [x] The Intermediary service reports the time an ETOR message is ingestion into ReportStream.

### Research
- [ ] *Research work needed to complete the story*
- [ ] Foundational: *High-level research that will support this and future efforts*

###  Engineering

- [x] Provide a tech spec to the RS team and get their buy-in.
- [x] TI 
  - [x] Create new branch. story/989/ingestion-time
  - [x] Add new method to call RS delivery endpoint in `RSEndpointClient` and `ReportStreamEndpointClient` and `MockRSEndpointClient`
  - [x] Update `updateMetadataForReceivedMessage` to use delivery endpoint instead of history endpoint
- [x] RS
  - [x] Create new branch.  stream1/989/ingestion-time
  - [x] Add `ingestionTime` to `DeliveryHistory`.
  - [x] Call `reportService.getRootReport` in `DeliveryFacade` to get the `created_at` and fill into `ingestionTIme`.

## Definition of Done
- [x] Documentation tasks completed
  - [x] Documentation and diagrams created or updated
    - [x] Implementation guide (`/ig` folder)
    - [x] ADRs (`/adr` folder)
    - [x] Main `README.md`
    - [x] Other READMEs in the repo
    - [x] If applicable, update the `ReportStream Setup` section in `README.md`
  - [x] [Threat model](https://lucid.app/lucidchart/8c6e8d37-2612-42a8-ac57-150c83f8e29e/edit) updated
  - [x] API documentation updated
- [x] Code quality tasks completed
  - [x] Code refactored for clarity and no design/technical debt
  - [x] Adhere to separation of concerns; code is not tightly coupled, especially to 3rd party dependencies
  - [x] Code is reviewed or developed by pair; 1 approval is needed but consider requiring an outside-the-pair reviewer
  - [x] Code quality checks passed
- [x] Security & Privacy tasks completed
  - [x] Security & privacy gates passed
- [x] Testing tasks completed
  - [x] Load tests passed
  - [x] Unit test coverage of our code >= 90%
- [x] Build & Deploy tasks completed
  - [x] Build process updated
  - [x] API(s) are versioned
  - [x] Feature toggles created and/or deleted.  Document the feature toggle
[...]

PR #1021, opened 2024-04-12
Story/989/ingestion time fix

somesylvie:
# Add a PR title
When the ingestion time is not found, we were getting a null pointer exception. This PR adds a check that the `originalIngestion` time is not null and is not empty before trying to use it

## Issue
https://github.com/CDCgov/trusted-intermediary/issues/989

## Checklist


basiliskus (approved):
Looks good!

halprin (approved):
Ship it!

Issue #974, opened 2024-03-26
Test Database Migrations in PRs


halprin:
# DevEx/OpEx

Whenever a database migration file is added or modified in a PR, we should start a job that runs a dummy database (through Docker?) and run Liquibase on it with these added/modified migration file(s).  This will check that we wrote the migration scripts correctly during the PR review process.

## Tasks

- [x] Add GitHub action.
- [x] Make GitHub action required to pass.

## Additional Context

Add any other context or screenshots about the work here.

PR #1036, opened 2024-04-23
Added Liquibase migration action

jherrflexion:
# Added Liquibase migration action 

Added Liquibase migration GitHub Action 

## Issue

#974

## Checklist


luis-pabon-tf (approved):
The yaml is valid and it looks good. Let's merge and then we can easily test it on other PRs.

Issue #1024, opened 2024-04-16
UCSD Baseline Results Transformation


JohnNKing:
# Story
As UCSD, so that our EHR can ingest results messages from ReportStream, I need a variety of fields removed, modified, or swapped based on our technical scoping research.

## Pre-conditions
- These transformations should only be applied to either UCSD or CDPH.
- The transformations have been implemented in the RS staging environment

## Acceptance Criteria
- [x] Result messages delivered to UCSD are transformed based on https://docs.google.com/spreadsheets/d/1E7-ARoZ7Eb_8-N25XT3Fb77aybmsHHrD/edit#gid=1250549121

### Research
- [x] Update CA results spreadsheet to include which transformation are specific to UCSD and which are potentially needed by additional CA hospitals receiving CDPH ORUs. (also a task in #955)

###  Engineering
- [x] Transformation Engine
  - [x] Design transformation rule engine
  - [x] Refactor validation rule engine to reuse code for transformation engine
  - [x] Implement transformation engine
    - [x] Decide where to house custom methods (a folder w/1 file per method that implements the `CustomFhirTransformation` interface?) @basiliskus 
    - [x] Implement `TransformationRule.runRule`: get method by name and evaluate with args 
    - [x] Fix `TransformRule` mapping from json file
    - [x] Refactor to use transformation engine instead of current conversion
      - [x] In `ConvertAndSendDemographicsUsecase` replace `OrderConverter` with `transformationEngine.runRules`
      - [x] In `SendOrderUseCase` replace `OrderConverter` with `transformationEngine.runRules`
        - [x] Remove `OrderConverter` interface
        - [x] Turn `HapiOrderConverter` into a class with static methods. Remove singleton and dependency injection
        - [x] Fix tests  
      - [x] In `SendResultUseCase` replace `ResultConverter` with `transformationEngine.runRules`
        - [x] Remove `ResultConverter` interface
        - [x] Remove `HapiResultConverter`
        - [x] Fix tests  
      - [x] Fix e2e tests
      - [x] Add test coverage 
    - [x] Move current transformations to engine
      - [x] `HapiOrderConverter.convertToOrder`  @saquino0827 
      - [x] `HapiOrderConverter.convertToOmlOrder`  @luis-pabon-tf 
      - [x] `HapiOrderConverter.addContactSectionToPatientResource` @luis-pabon-tf 
      - [x] `HapiOrderConverter.addEtorProcessingTag` @jorg3lopez 
[...]

PR #1037, opened 2024-04-23
Rules engine refactoring

basiliskus:
# Rules engine refactoring

Refactoring to make validation rule engine code more generic and reusable for the transformation rule engine

## Issue

#1024 

## Checklist


etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/Rule.java
```diff
@@ -12,17 +12,9 @@ public interface Rule {
...
-     * sure that no PII or PHI is included in the message!
-     */
-    String getViolationMessage();
-
     List<String> getConditions();
 
-    List<String> getValidations();
-
-    boolean isValid(FhirResource<?> resource);
+    boolean shouldRun(FhirResource<?> resource);
```
luis-pabon-tf, line 17: yes yes these are very good renames

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/OrderController.java
```diff
@@ -20,7 +20,7 @@ public class OrderController {
     @Inject HapiFhir fhir;
     @Inject Logger logger;
     @Inject MetricMetadata metadata;
-    @Inject RuleEngine ruleEngine;
+    @Inject ValidationRuleEngine validationEngine;
```
jcrichlake, line 23: You should be injecting the RulesEngine, not the implementation of the interface

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
```diff
@@ -133,7 +133,9 @@ public Map<HttpEndpoint, Function<DomainRequest, DomainResponse>> domainRegistra
                 PartnerMetadataConverter.class, HapiPartnerMetadataConverter.getInstance());
         // Validation rules
         ApplicationContext.register(RuleLoader.class, RuleLoader.getInstance());
-        ApplicationContext.register(RuleEngine.class, RuleEngine.getInstance());
+        ApplicationContext.register(
+                ValidationRuleEngine.class,
```
jcrichlake, line 137: The first value in register should be an interface and the second should be the class that is defining the behavior for an interface

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/Rule.java
```diff
@@ -1,28 +1,85 @@
...
+    private String name;
+    private String description;
+    private String message;
+    private List<String> conditions;
+    private List<String> rules;
 
     /**
-     * Descriptive message when there's a rule violation Note: When implementing this method, make
-     * sure that no PII or PHI is included in the message!
+     * Do not delete this constructor! It is used for JSON deserialization when loading rules from a
```
jcrichlake, line 24: Right, Jackson requires an empty constructor to de-serialize things, don't have to add this but something to think about

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/orders/OrderControllerTest.groovy
```diff
@@ -12,14 +12,14 @@ import org.hl7.fhir.r4.model.Bundle
...
-    def ruleEngine = Mock(RuleEngine)
+    def ruleEngine = Mock(ValidationRuleEngine)
 
     def setup() {
         TestApplicationContext.reset()
         TestApplicationContext.init()
         TestApplicationContext.register(OrderController, OrderController.getInstance())
         TestApplicationContext.register(MetricMetadata, Mock(MetricMetadata))
-        TestApplicationContext.register(RuleEngine, ruleEngine)
+        TestApplicationContext.register(ValidationRuleEngine, ruleEngine)
```
jcrichlake, line 22: Same thing here, I think it should be RuleEngine for the register

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/OrderController.java
basiliskus: We need to inject two different rules engine, one for the validation and another one for the transformation. The transformation engine is being implemented in the other active PR that we have

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
basiliskus: As commented above, we need to be able to identify two different implementations of the rule engine

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/orders/OrderControllerTest.groovy
basiliskus: Same as above

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
jorg3lopez: Thoughts on doing a parent interface (engine), and two child interfaces that will extend the parent interface (RuleEngine, ValidationEngine)? With this setup, you can use the child interface to register the implementation for each. The only change here is the naming for the parent interface, the child interfaces, and the implementations for each child interface. You register both engines with a unique interface for each.

basiliskus: I don't think the value we register in the ApplicationContext needs to be an interface, a lot of the existing ones are not. Is there a benefit in creating the interface even if it will have only one implementation?

jcrichlake: I think we generally wrap 3rd party dependencies in an interface to reduce dependencies

jcrichlake (approved):

Issue #690, opened 2023-11-30
New ReportStream Metadata API & Authentication


JohnNKing:
# Story
As a hospital or lab, so that I can use the existing ReportStream credential flow, I need a new metadata endpoint that retrieves the metadata from the Intermediary service.

## Pre-conditions
- Access to metadata for hospitals and labs has already been implemented.
- Hospital or Lab is able to login to RS.
- Linking work is done.
 
## Acceptance Criteria
- [x] The hospital is able view their order/results metadata through ReportStream if and only if they've successfully authenticated with their RS credentials. 
- [x] The lab is able to view their order/results metadata through ReportStream if and only if they've successfully authenticated with their RS credentials.

### Research
- [x] *Research work needed to complete the story*
- [x] Foundational: *High-level research that will support this and future efforts*

###  Engineering
- [x] Codify formal requirements for RS.  See notes bellow. - @halprin, @scleary1cs 
- [x] Send formal requirements to RS.
- [x] Figure out how to authenticate to TI from Report Stream (Find where RS is calling TI for orders?)
- [x] Add wrapper endpoint to report stream that calls the intermediary metadata endpoint
- [x] Validate that the new endpoint fails when not authenticated
- [x] Unit tests for new endpoint
  - [x] Auth endpoint
- [x] How to pass back TI response body unchanged through RS 
- [x] Ask RS folks about release process if there's anything more than just getting the change into the master branch and then waiting
- [x] TI. origin/stream1/690/passthrough
  - [x] Update use cases to retrieve and store report id instead of submission id
  - [x] Fix any broken test cases related to the above
- [x] RS halprin/ti-690/metadata-passthrough (Branch that contains a mostly complete function sans auth)
  - [x] Add a receiver metadata endpoint.  This uses the existing authorization of other receiver endpoints.
  - [x] Add a sender metadata endpoint.  This uses the existing authorization of other sender endpoints.
  - [x] Add environment variables for TI endpoint in RS call 
  - [x] Both metadata endpoints need to convert the passed-in ID to the ID that TI knows about (walk the lineage).
  - [x] Comment for endpoint stating that it shouldn't be changed except by a member of the TI team
  - [x] Unit tests.
  - [x] Refactor RS functions to pull duplicate code up into `ReportFileFunction.kt` I think the initialization of the vars from response to the jksCredential can be a data class along with 2 other suspending functions to call the duplicate suspending functions.  

## Definition of Done
- [x] Documentation tasks completed
[...]

PR #1039, opened 2024-04-24
690: Save `reportId` for Outbound Messages

halprin:
# Save `reportId` for Outbound Messages

We have pivoted away from saving `submissionId` to saving `reportId` for outbound messages sent to RS.

## Issue

#690.

## Checklist


etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/SendOrderUseCase.java
```diff
@@ -54,7 +54,7 @@ public void convertAndSend(final Order<?> order, String receivedSubmissionId)
                 EtorMetadataStep.ETOR_PROCESSING_TAG_ADDED_TO_MESSAGE_HEADER);
 
         String sentSubmissionId = sender.send(omlOrder).orElse(null);
```
somesylvie, line 56: Should this `sentSubmissionId` variable also be some flavor of `reportId`?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamSenderHelper.java
```diff
@@ -55,23 +55,23 @@ protected Optional<String> sendToReportStream(
...
+        return outboundReportId;
     }
 
-    protected Optional<String> getSubmissionId(String rsResponseBody) {
+    protected Optional<String> getReportId(String rsResponseBody) {
         try {
             Map<String, Object> rsResponse =
                     formatter.convertJsonToObject(rsResponseBody, new TypeReference<>() {});
-            return Optional.ofNullable(rsResponse.get("submissionId").toString());
+            return Optional.ofNullable(rsResponse.get("reportId").toString());
```
somesylvie, line 72: I may be misunderstanding this, but it looks like it's parsing a field from a ReportStream response, and we're now looking for a different field than we were before. Is that the intention/are both `submissionId` and `reportId` real fields and we're switching which we're using?

somesylvie (commented):
Are we planning to refactor away other uses of `submissionId`, or are some appropriate to keep? It looks like it's in a bunch of other places in the codebase still

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamSenderHelper.java
halprin: Exactly.  Both `submissionId` and `reportId` are fields in a real RS response.  We are changing what we're using.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/SendOrderUseCase.java
halprin: Good catch.

halprin:
> Are we planning to refactor away other uses of `submissionId`, or are some appropriate to keep? It looks like it's in a bunch of other places in the codebase still

We are, but I'm not sure if this is the PR to do it everywhere.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/SendOrderUseCase.java
halprin: Fixed.

basiliskus:
> > Are we planning to refactor away other uses of `submissionId`, or are some appropriate to keep? It looks like it's in a bunch of other places in the codebase still
> 
> We are, but I'm not sure if this is the PR to do it everywhere.

Should we create a devex/opex story to rename these in the rest of the codebase?

basiliskus (approved):

halprin:
> > > Are we planning to refactor away other uses of `submissionId`, or are some appropriate to keep? It looks like it's in a bunch of other places in the codebase still
> > 
> > 
> > We are, but I'm not sure if this is the PR to do it everywhere.
> 
> Should we create a devex/opex story to rename these in the rest of the codebase?

Definitely.

Issue #990, opened 2024-04-01
Sender/Receiver Metadata HL7 Format


scleary1cs:
# Story
As a hospital or lab, so I know the original sender/receiver of the message I need metadata to include the sending/receiving facilities in the HL7 message.

## Pre-conditions
- NA

## Acceptance Criteria
- [x] The sender in the metadata uses data from the message itself.
- [x] The receiver in the metadata uses data from the message itself.

## Tasks
- [x] Create a new branch "story-990/sender-receiver-metadata"

### Research
- [x] *Research work needed to complete the story*
- [x] Foundational: *High-level research that will support this and future efforts*

###  Engineering
- [x] Work for `PartnerMetaData.java`
  - [x] Remove `sender` param
  - [x] Remove `receiver` param
  - [x] Fix any unit test that fails due to the changes
- [x] Work for `PartnerMetadataOrchestrator.java`
  - [x] Remove sender and all its occurrences from PartnerMetaData.
  - [x] Remove receiver and all its occurrences from PartnerMetaData.
  - [x] Replace history endpoint with the above values
- [x] Update test cases in PartnerMetadataOrchestratorTest
- [x] Work for `HapiPartnerMetadataConverter.java`
  - [x] Make appropriate changes to reflect the work from `PartnerMetaData`
  - [x] Verify all unit tests pass
- [x] work in PostgresDaoTest
  - [x] Make appropriate changes to reflect the work from `PartnerMetaData`
  - [x] Verify all unit tests pass
- [x] Work for Database
  - [x] delete `sender` and `receiver` colunms
  - [x] Test metadata gets recorded

## Definition of Done
- [x] Documentation tasks completed
  - [x] Documentation and diagrams created or updated
    - [x] Implementation guide (`/ig` folder)
    - [x] ADRs (`/adr` folder)
    - [x] Main `README.md`
    - [x] Other READMEs in the repo
    - [x] If applicable, update the `ReportStream Setup` section in `README.md`
  - [x] [Threat model](https://lucid.app/lucidchart/8c6e8d37-2612-42a8-ac57-150c83f8e29e/edit) updated
  - [x] API documentation updated
- [x] Code quality tasks completed
  - [x] Code refactored for clarity and no design/technical debt
  - [x] Adhere to separation of concerns; code is not tightly coupled, especially to 3rd party dependencies
  - [x] Code is reviewed or developed by pair; 1 approval is needed but consider requiring an outside-the-pair reviewer
[...]

PR #1040, opened 2024-04-24
Revisit 990

jorg3lopez:
# Remove Sender and Receiver entities

This PR removes the sender and receiver columns from the database and from our code base. The sender and receiver can be access through the `sendingFacilityDetails` and the `receivingFacilityDetails`. Due to this, the isolated sender and receiver become redundant code.

The main work was to remove these entities from the `PartnertMetadata` class and from the database.

## Issue
#990 

## Checklist

### DIFF
--- a/etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/metadata/partner/PartnerMetadataOrchestratorTest.groovy
+++ b/etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/metadata/partner/PartnerMetadataOrchestratorTest.groovy
@@ -51,8 +51,7 @@ class PartnerMetadataOrchestratorTest extends Specification {
         receivingFacility = new MessageHdDataType("receiving_facility_name", "receiving_facility_id", "receiving_facility_type")
 
         testMetadata = new PartnerMetadata(receivedSubmissionId,
-                sentSubmissionId
-                ,
+                sentSubmissionId,
                 timeReceived,
                 timeDelivered,
                 hashCode,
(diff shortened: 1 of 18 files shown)


etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/metadata/partner/PartnerMetadataOrchestratorTest.groovy
```diff
@@ -26,7 +26,6 @@ class PartnerMetadataOrchestratorTest extends Specification {
     def hashCode = "hash"
     def bearerToken = "token"
     def sender = "sender"
```
basiliskus, line 28: I think this `sender` can be removed

examples/Test/e2e/orders/002_ORM_O01.fhir
```diff
@@ -1,1406 +1,1626 @@
 {
```
basiliskus, line 1: Why was this file changed? We need to preserve some files in the examples folder as other tests may rely on it (like the load tests). In the case of this particular one, I think we want to preserve it because it was based on an example we got from MN. If necessary, it's better to add a new file and point to that one

basiliskus (commented):
Left a couple of comments, otherwise looks good to me!

examples/Test/e2e/orders/002_ORM_O01.fhir
halprin: This FHIR file indeed came from an MN HL7v2 file, but it was outdated based on the old RS HL7v2/FHIR conversion algorithm.  For example, nothing was coming out as the sender and receiver because this FHIR file was missing how RS now extracts that information.  So, we basically reran the original HL7v2 file through RS again to get the updated FHIR file.

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/metadata/partner/PartnerMetadataOrchestratorTest.groovy
jorg3lopez: Great catch!

basiliskus (approved):

halprin (approved):
Ship it.

PR #1043, opened 2024-04-24
Refactor readMetadataForMessageLinking so it doesn't return itself

basiliskus:
# Refactor readMetadataForMessageLinking so it doesn't return itself

This is a fix in the linking logic so it only matches when there's at least one pair

## Issue

#621 

## Checklist

### DIFF
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messages/SendMessageHelper.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messages/SendMessageHelper.java
@@ -70,6 +70,10 @@ public class SendMessageHelper {
             if (messageIdsToLink == null || messageIdsToLink.isEmpty()) {
                 return;
             }
+
+            // Add receivedSubmissionId to complete the list of messageIds to link
+            messageIdsToLink.add(receivedSubmissionId);
+
             logger.logInfo(
                     "Found messages to link for receivedSubmissionId {}: {}",
                     receivedSubmissionId,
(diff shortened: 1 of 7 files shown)


etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messages/SendMessageHelper.java
```diff
@@ -70,6 +70,10 @@ public void linkMessage(String receivedSubmissionId) {
             if (messageIdsToLink == null || messageIdsToLink.isEmpty()) {
                 return;
             }
+
+            // Add receivedSubmissionId to complete the list of messageIds to link
+            messageIdsToLink.add(receivedSubmissionId);
+
```
halprin, line 76: Does this work correctly now?

Did `partnerMetadataOrchestrator.findMessagesIdsToLink(receivedSubmissionId)` return `receivedSubmissionId` in the returned set, but now it doesn't?  Is it possible that the `if` statement now always returns an empty set if this is the first one?

basiliskus: Correct, findMessagesIdsToLink was returning receivedSubmissionId in the set. The problem with the previous logic and query is that it will return a record even if there's no pair to match, as it will always return itself. In this case, if there's a pair to match, it will return that so if won't be empty. If there's no pair to match, we want to return an empty set

halprin (approved):
Thanks for explaining it to me!

PR #1045, opened 2024-04-25
Proposal example to move file loading to RuleLoader

luis-pabon-tf:
# Add a PR title

## Issue

## Checklist


saquino0827:
I'm more inclined to keep file loading in their RuleEngines themselves as to separate the rule type, while keeping RuleLoader generic and only for loading the rules they are given via parameters.

basiliskus:
> I'm more inclined to keep file loading in their RuleEngines themselves as to separate the rule type, while keeping RuleLoader generic and only for loading the rules they are given via parameters.

I agree. I think we need more separation between the engines to be able to instantiate and call them separately. Aside from that, I like the idea of adding a loadDefinition method and converting the RuleEngine into a singleton for consistency and testing

luis-pabon-tf:
We took what we needed from here and went down a different road.

PR #1046, opened 2024-04-26
Transformation rule engine implementation

basiliskus:
# Transformation rule engine implementation

Transformation rule engine implementation

**Note**: there is code duplication in `ValidationRuleEngine` and `TransformationRuleEngine` that we have decided to refactor in a future PR

## Issue

#1024 

## Checklist


jorg3lopez:
I noticed that for each transformation there is a class created. Thoughts on having one class that has all the transformations to choose from and the only thing we need to tell the transformation engine is which transformation we want, assuming the engine loads a client settings file holding the settings of the clients? I remember agreeing on having one file with all the clients' settings and later on switching to one settings file per client.

basiliskus:
> I noticed that for each transformation there is a class created. Thoughts on having one class that has all the transformations to choose from and the only thing we need to tell the transformation engine is which transformation we want, assuming the engine loads a client settings file holding the settings of the clients?

That's what's already happening unless I'm misunderstanding. There is a settings file (transformation_definitions.json) that has the transformations we want. Are you suggesting we have one settings file per client? The benefit of having one class per transformation is that it can implement the interface that we require for the transformation method. Also, having one class with all the transformations can make it harder to maintain as it grows

jorg3lopez:
> > I noticed that for each transformation there is a class created. Thoughts on having one class that has all the transformations to choose from and the only thing we need to tell the transformation engine is which transformation we want, assuming the engine loads a client settings file holding the settings of the clients?
> 
> That's what's already happening unless I'm misunderstanding. There is a settings file (transformation_definitions.json) that has the transformations we want. Are you suggesting we have one settings file per client? The benefit of having one class per transformation is that it can implement the interface that we require for the transformation method. Also, having one class with all the transformations can make it harder to maintain as it grows

Yes. I was thinking having a transformation class that will house all the transformations. This class can get injected into the engine. All the transformation logic can live in one place only creating one object for the logic.

basiliskus:
> > > I noticed that for each transformation there is a class created. Thoughts on having one class that has all the transformations to choose from and the only thing we need to tell the transformation engine is which transformation we want, assuming the engine loads a client settings file holding the settings of the clients?
> > 
> > 
> > That's what's already happening unless I'm misunderstanding. There is a settings file (transformation_definitions.json) that has the transformations we want. Are you suggesting we have one settings file per client? The benefit of having one class per transformation is that it can implement the interface that we require for the transformation method. Also, having one class with all the transformations can make it harder to maintain as it grows
> 
> Yes. I was thinking having a transformation class that will house all the transformations. This class can get injected into the engine. All the transformation logic can live in one place only creating one object for the logic. But definately, scalability will be better using a class per transformation if the number of transformation become huge.

One of the advantages of the current design is that it will allow partners to self-serve and add transformations in the future with pre-defined methods, as well as update and remove transformations (using a UI and without updating any source code)

basiliskus:
> > > > > I noticed that for each transformation there is a class created. Thoughts on having one class that has all the transformations to choose from and the only thing we need to tell the transformation engine is which transformation we want, assuming the engine loads a client settings file holding the settings of the clients?
> > > > 
> > > > 
> > > > That's what's already happening unless I'm misunderstanding. There is a settings file (transformation_definitions.json) that has the transformations we want. Are you suggesting we have one settings file per client? The benefit of having one class per transformation is that it can implement the interface that we require for the transformation method. Also, having one class with all the transformations can make it harder to maintain as it grows
> > > 
> > > 
> > > Yes. I was thinking having a transformation class that will house all the transformations. This class can get injected into the engine. All the transformation logic can live in one place only creating one object for the logic. But definately, scalability will be better using a class per transformation if the number of transformation become huge.
> > 
> > 
> > One of the advantages of the current design is that it will allow partners to self-serve and add transformations in the future with pre-defined methods, as well as update and remove transformations (using a UI and without updating any source code)
> 
> I like that feature. I thought the settings file would determine what translations to do. But correct me if I'm wrong, the process that we are looking at is to add the transformations that the settings file will have?
> 
> I had added this to the previous quote:
> 
> How about if we combine both? I think this will give us all the benefits from both, centralization, scalability, single responsibility, reusability, and modularity. One Transformation class to house all the transformation class. I'm trying to think about the drawbacks.

The definition file has the condition (i.e. the transformation should only run for CA, or only for ORMs, etc) and the name to the transformation to run. I'm not sure I'm understanding what you see as the drawback of having one class per transformation. We can meet to discuss this

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRule.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+    }
+
+    private static Class<?> loadCustomTransformationClassFromFile(String className)
+            throws ClassNotFoundException {
+        String customPackageName =
+                "gov.hhs.cdc.trustedintermediary.etor.ruleengine.transformation.custom";
+        Path customTransformationPath =
+                Paths.get(
+                        System.getProperty("user.dir"),
```
jorg3lopez, line 63: I know we use the ApplicationContext for getting environment variables. I wonder if it would work in here as well?

basiliskus: Good point. I think we need a standard way to easily construct paths. ApplicationContext seems like a good option

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/demographics/ConvertAndSendDemographicsUsecase.java
```diff
@@ -26,7 +27,8 @@ public static ConvertAndSendDemographicsUsecase getInstance() {
     private ConvertAndSendDemographicsUsecase() {}
 
     public void convertAndSend(Demographics<?> demographics) throws UnableToSendMessageException {
-        Order<?> order = converter.convertToOrder(demographics);
+        transformationEngine.runRules(demographics);
+        var order = new HapiOrder((Bundle) demographics.getUnderlyingResource());
```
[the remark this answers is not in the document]
basiliskus: I'll take a look

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRuleEngine.java
```diff
@@ -0,0 +1,86 @@
...
+            return;
+        }
+
+        rules.forEach(
+                rule -> {
+                    try {
+                        if (rule.shouldRun(resource)) {
+                            rule.runRule((resource));
+                        }
+                    } catch (Exception e) { // Do we need a custom exception for rules?
```
jorg3lopez, line 74: I wrote that comment to remind me. I trhink it would be a good idea, thoughts?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRuleEngine.java
```diff
@@ -0,0 +1,81 @@
...
+    private String ruleDefinitionsFileName;
+    final List<TransformationRule> rules = new ArrayList<>();
+
+    private static final TransformationRuleEngine INSTANCE = new TransformationRuleEngine();
+
+    @Inject Logger logger;
+    @Inject RuleLoader ruleLoader;
+
+    public static TransformationRuleEngine getInstance(String ruleDefinitionsFileName) {
+        INSTANCE.ruleDefinitionsFileName = ruleDefinitionsFileName;
```
jorg3lopez, line 26: This will run into concurrency issues. We probably need to synchronize how we initialize the instance. My question now is how is this going to work when we get to the point of each client loading their settings into the engine. Right now, this is set up to take one file with all the clients' settings. In the future, each client will have a settings file. This is a singleton class, the instance is shared across all API instances that are running. Should each API instance have a different instance of the engine?

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRuleEngineIntegrationTest.groovy
```diff
@@ -0,0 +1,153 @@
...
+        engine.runRules(new HapiFhirResource(bundle))
+
+        then:
+        0 * mockLogger.logError(_ as String, _ as Exception)
+        (1.._) * mockLogger.logInfo(_ as String, _ as String)
+    }
+
+    def "all transformations in the definitions file have existing custom methods"() {
+        when:
+        def transformationMethodNames = engine.rules.collect { rule -> rule.name }
```
jorg3lopez, line 51: Thoughts on accessing the rules list through a `getRules()` method?

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRuleEngineIntegrationTest.groovy
```diff
@@ -0,0 +1,153 @@
...
+        TestApplicationContext.register(HapiFhir, fhir)
+        TestApplicationContext.register(TransformationRuleEngine, engine)
+        TestApplicationContext.register(RuleLoader, RuleLoader.getInstance())
+        TestApplicationContext.register(Logger, mockLogger)
+        TestApplicationContext.register(MetricMetadata, Mock(MetricMetadata))
+
+        TestApplicationContext.injectRegisteredImplementations()
+    }
+
+    def "transformation rules run without error"() {
```
jorg3lopez, line 37: I believe this is the test that is leaking onto the test below (`all transformations in the definitions file have existing custom methods`).

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRuleEngineIntegrationTest.groovy
```diff
@@ -0,0 +1,153 @@
...
+        "addContactSectionToPatientResource" | "e2e/orders/003_2_ORM_O01_short_linked_to_002_ORU_R01_short.fhir"
+        "convertDemographicsToOrder"         | "e2e/demographics/001_Patient_NBS.fhir"
+    }
+
+    def "Testing accuracy of rule: convertDemographicsToOrder"() {
+        given:
+        def untouchedBundle = new Bundle()
+        def bundle = untouchedBundle.copy()
+        engine.ensureRulesLoaded()
+        engine.rules.removeAll(engine.rules.findAll {
```
jorg3lopez, line 85: same as above regarding the use of a method to get the rules list.

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRuleEngineIntegrationTest.groovy
```diff
@@ -0,0 +1,153 @@
...
+        untouchedBundle.entry.isEmpty()
+        bundle.entry[0].getResource().meta.tag.last().code == "ETOR"
+    }
+
+    def "Testing accuracy of rule: convertToOmlOrder"() {
+        given:
+        def untouchedBundle = FhirBundleHelper.createMessageBundle(messageTypeCode: "ORM_O01")
+        def bundle = untouchedBundle.copy()
+        engine.ensureRulesLoaded()
+        engine.rules.removeAll(engine.rules.findAll {
```
jorg3lopez, line 122: same as above regarding the use of a method to get the rules list.

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRuleEngineIntegrationTest.groovy
```diff
@@ -0,0 +1,153 @@
...
+        bundle.entry[0].getResource().event.code == "O21"
+    }
+
+    def "Testing accuracy of rule: addContactSectionToPatientResource"() {
+        given:
+        def untouchedBundle = FhirBundleHelper.createMessageBundle(messageTypeCode: "OML_O21")
+        untouchedBundle.addEntry(new Bundle.BundleEntryComponent().setResource(new Patient()))
+        def bundle = untouchedBundle.copy()
+        engine.ensureRulesLoaded()
+        engine.rules.removeAll(engine.rules.findAll {
```
jorg3lopez, line 140: same as above regarding the use of a method to get the rules list.

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRuleEngineTest.groovy
```diff
@@ -0,0 +1,152 @@
...
+    def "ensureRulesLoaded happy path"() {
+        given:
+        mockRuleLoader.loadRules(_ as InputStream, _ as TypeReference) >> [mockRule]
+
+        when:
+        ruleEngine.ensureRulesLoaded()
+
+        then:
+        1 * mockRuleLoader.loadRules(_ as InputStream, _ as TypeReference) >> [mockRule]
+        ruleEngine.rules.size() == 1
```
jorg3lopez, line 39: same comment regarding the use of a method to get the rules list.

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/ExamplesHelper.groovy
```diff
@@ -0,0 +1,26 @@
+package gov.hhs.cdc.trustedintermediary
+
+import gov.hhs.cdc.trustedintermediary.context.ApplicationContext
+import gov.hhs.cdc.trustedintermediary.external.hapi.HapiFhirImplementation
+import gov.hhs.cdc.trustedintermediary.external.hapi.HapiFhirResource
+import org.hl7.fhir.r4.model.Bundle
+
+import java.nio.file.Files
+
+class ExamplesHelper {
```
jorg3lopez, line 10: Java docs

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/FhirBundleHelper.groovy
```diff
@@ -0,0 +1,33 @@
+package gov.hhs.cdc.trustedintermediary
+
+import org.hl7.fhir.r4.model.Bundle
+import org.hl7.fhir.r4.model.Coding
+import org.hl7.fhir.r4.model.MessageHeader
+import org.hl7.fhir.r4.model.Organization
+import org.hl7.fhir.r4.model.Reference
+
+class FhirBundleHelper {
```
jorg3lopez, line 9: java docs

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRule.java
```diff
@@ -0,0 +1,86 @@
...
+
+    static Class<?> loadClassFromCache(String className) throws ClassNotFoundException {
+        return classCache.computeIfAbsent(className, TransformationRule::loadClassByName);
+    }
+
+    static void executeCustomTransformationMethod(
+            Class<?> clazz, FhirResource<?> resource, Map<String, String> args)
+            throws NoSuchMethodException, InvocationTargetException, InstantiationException,
+                    IllegalAccessException {
+        Method method = clazz.getDeclaredMethod("transform", FhirResource.class, Map.class);
```
jorg3lopez, line 49: Thoughts on using an interface and concrete implementations for each type of transformation instead of reflection?

basiliskus: Good point. We already have an interface so I changed it to use that

halprin (commented):
Some initial comments.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/SendOrderUseCase.java
```diff
@@ -42,18 +42,9 @@ public void convertAndSend(final Order<?> order, String receivedSubmissionId)
...
-        metadata.put(order.getFhirResourceId(), EtorMetadataStep.ORDER_CONVERTED_TO_OML);
+        transformationEngine.runRules(order);
 
-        omlOrder = converter.addContactSectionToPatientResource(omlOrder);
-        metadata.put(order.getFhirResourceId(), EtorMetadataStep.CONTACT_SECTION_ADDED_TO_PATIENT);
-
-        omlOrder = converter.addEtorProcessingTag(omlOrder);
-        metadata.put(
-                order.getFhirResourceId(),
-                EtorMetadataStep.ETOR_PROCESSING_TAG_ADDED_TO_MESSAGE_HEADER);
```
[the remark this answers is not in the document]
basiliskus: These were moved to the custom functions. You can find this etor processing one in `etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/addEtorProcessingTag.java`

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/results/SendResultUseCase.java
```diff
@@ -45,12 +43,9 @@ public void convertAndSend(Result<?> result, String receivedSubmissionId)
 
         sendMessageHelper.savePartnerMetadataForReceivedMessage(partnerMetadata);
 
-        var convertedResult = converter.addEtorProcessingTag(result);
-        metadata.put(
-                result.getFhirResourceId(),
-                EtorMetadataStep.ETOR_PROCESSING_TAG_ADDED_TO_MESSAGE_HEADER);
```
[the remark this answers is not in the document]
basiliskus: Same as above

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
```diff
@@ -10,4 +10,6 @@ public interface RuleEngine {
     void ensureRulesLoaded() throws RuleLoaderException;
 
     void runRules(FhirResource<?> resource);
+
+    Rule<?> getRuleByName(String ruleName);
```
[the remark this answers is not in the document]
basiliskus: Sure

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/addSendingFacilityToMessageHeader.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class addSendingFacilityToMessageHeader implements CustomFhirTransformation {
+
+    private final MetricMetadata metadata =
+            ApplicationContext.getImplementation(MetricMetadata.class);
+
+    @Override
+    public void transform(FhirResource<?> resource, Map<String, String> args) {
+        Bundle bundle = (Bundle) resource.getUnderlyingResource();
+        HapiOrderConverter.addSendingFacilityToMessageHeader(
```
basiliskus, line 20: @luis-pabon-tf  I was thinking we can add new transformations in a new PR as this one is big enough

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiOrderConverter.java
```diff
@@ -26,8 +25,7 @@
  * Order <Bundle>}). Also converts an order to identify as an HL7v2 OML in the {@link
  * MessageHeader}.
  */
-public class HapiOrderConverter implements OrderConverter {
-    private static final HapiOrderConverter INSTANCE = new HapiOrderConverter();
+public class HapiOrderConverter {
```
halprin, line 28: You changed these converters to be unable to be `@Injected` and pure helper static classes.  What is your reasoning behind this?

halprin (commented):
More comments.  Weee!

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/ExamplesHelper.groovy
halprin: This is a unit test file, so while we don't require them, if something is strange and can benefit explaining "why", then please add JavaDocs.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRule.java
```diff
@@ -0,0 +1,85 @@
...
+        return classCache.computeIfAbsent(className, TransformationRule::loadClassByName);
+    }
+
+    private static Class<?> loadClassByName(String className) {
+        String fullClassName = getFullClassName(className);
+        try {
+            return Class.forName(fullClassName);
+        } catch (ClassNotFoundException e) {
+            throw new RuntimeException(e);
+        }
```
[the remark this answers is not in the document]
basiliskus: I think @jorg3lopez added this one. Thoughts Jorge?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiOrderConverter.java
basiliskus: This implementation changed now that we're using the transformation engine. Methods in this class are only called in custom transformations. We could actually move most of this code into the custom methods but I thought we could leave them as a helper class that could be reuse by multiple transformations

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRule.java
```diff
@@ -0,0 +1,85 @@
...
+                "gov.hhs.cdc.trustedintermediary.etor.ruleengine.transformation.custom";
+        return packageName + "." + className;
+    }
+
+    static void executeCustomTransformationMethod(
+            Class<?> clazz, FhirResource<?> resource, Map<String, String> args)
+            throws NoSuchMethodException, InvocationTargetException, InstantiationException,
+                    IllegalAccessException {
+        CustomFhirTransformation transformation =
+                (CustomFhirTransformation) clazz.getDeclaredConstructor().newInstance();
```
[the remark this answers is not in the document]
basiliskus: Great point. @jorg3lopez added the caching here. What do you think?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRule.java
```diff
@@ -0,0 +1,85 @@
...
+        logger.logInfo("Applying transformation: ", name);
+
+        try {
+            Class<?> clazz = loadClassFromCache(name);
+            executeCustomTransformationMethod(clazz, resource, args);
+        } catch (NoSuchMethodException
+                | IllegalAccessException
+                | InvocationTargetException
+                | InstantiationException e) {
+            logger.logError("Error invoking method: " + name + ", due to: " + e.getMessage(), e);
```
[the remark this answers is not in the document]
basiliskus: Good point. We don't want to change the current behavior with this PR, including error handling. I'll look into this

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/RuleEngine.java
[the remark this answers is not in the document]
basiliskus: Removed

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRule.java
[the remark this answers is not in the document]
basiliskus: Added a new RuleException to throw here, bubble up, and then interrupt the flow when sending message

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRule.java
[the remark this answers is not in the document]
basiliskus: I refactored it to cache the instance

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRule.java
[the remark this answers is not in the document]
jorg3lopez: I wonder if we want to stop the flow or have it continue? Either way, we will need a way of handling the exception and notifying the client

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRuleEngine.java
jorg3lopez: This might work for the first instance if there is only one thread and one settings file. The problem might be more apparent when we start to load individual settings files for each client. I'm picturing a scenario where two threads are loading different settings files to the singleton instance. Do we want to address this now or do it when each client have their own settings file?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRule.java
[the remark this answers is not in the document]
jorg3lopez: This code has change, I don't see this method any more. And yes, I agree with throwing the custom exception.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/validation/ValidationRuleEngine.java
```diff
@@ -36,18 +35,17 @@ public void unloadRules() {
...
-                                    getClass()
-                                            .getClassLoader()
-                                            .getResource(ruleDefinitionsFileName)
-                                            .getPath());
-
+                    InputStream resourceStream =
+                            getClass()
+                                    .getClassLoader()
+                                    .getResourceAsStream(ruleDefinitionsFileName);
+                    assert resourceStream != null;
```
jorg3lopez, line 46: I think assertions run the risk of being disabled through JVM flags. Thoughts on using an if statement to check if an object is null?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRuleEngine.java
```diff
@@ -0,0 +1,71 @@
...
+    public void ensureRulesLoaded() throws RuleLoaderException {
+        // Double-checked locking - needed to protect from excessive sync locks
+        if (rules.isEmpty()) {
+            synchronized (this) {
+                if (rules.isEmpty()) {
+                    InputStream resourceStream =
+                            getClass()
+                                    .getClassLoader()
+                                    .getResourceAsStream(ruleDefinitionsFileName);
+                    assert resourceStream != null;
```
jorg3lopez, line 47: I think assertions run the risk of being disabled through JVM flags. Thoughts on using an if statement to check if an object is null?

shared/src/main/java/gov/hhs/cdc/trustedintermediary/context/ApplicationContext.java
```diff
@@ -164,4 +166,12 @@ public static String getProperty(String key, String defaultValue) {
     public static String getEnvironment() {
         return getProperty("ENV", "local");
     }
+
+    public static Path getRootPath() {
+        return Paths.get(System.getProperty("user.dir")).getParent();
```
jorg3lopez, line 171: Nice enhancement!

shared/src/main/java/gov/hhs/cdc/trustedintermediary/context/ApplicationContext.java
```diff
@@ -164,4 +166,12 @@ public static String getProperty(String key, String defaultValue) {
...
+    }
+
+    public static Path getExamplesPath() {
+        return getRootPath().resolve("examples");
```
jorg3lopez, line 175: This one is a good idea as well.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRule.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    private static CustomFhirTransformation createTransformationInstance(
+            String transformationName) {
+        String fullClassName = getFullClassName(transformationName);
+        try {
+            Class<?> clazz = Class.forName(fullClassName);
+            return (CustomFhirTransformation) clazz.getDeclaredConstructor().newInstance();
+        } catch (ClassNotFoundException
+                | InstantiationException
```
jcrichlake, line 69: Would it be better to catch an exception here instead of trying to catch the many different errors that could come through?

e2e/src/test/groovy/gov/hhs/cdc/trustedintermediary/e2e/DemographicsTest.groovy
```diff
@@ -29,24 +29,25 @@ class DemographicsTest extends Specification {
...
-        parsedSentPayload.entry[0].resource.resourceType == "MessageHeader"
-        parsedSentPayload.entry[2].resource.resourceType == "ServiceRequest"
-
-        parsedSentPayload.entry[1].resource.resourceType == "Patient"
-        parsedSentPayload.entry[1].resource.id == "infant-twin-1"
-
-        parsedSentPayload.entry[1].resource.identifier[1].value == parsedResponseBody.patientId  //the second (index 1) identifier so happens to be the MRN
-        parsedSentPayload.resourceType + "/" + parsedSentPayload.id == parsedResponseBody.fhirResourceId
-    }
+    //    todo: ignoring while figuring out how to filter the demographics example
```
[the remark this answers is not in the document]
basiliskus: Good idea. Done

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRule.java
basiliskus: We'd loose more specific information about the cause of the error if we just catch one Exception

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRuleEngine.java
basiliskus: Good point! I'll update it to use an if statement

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRule.java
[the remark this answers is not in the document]
basiliskus: For now I think we want to stop the flow to keep the current behavior. That's a good question though

e2e/src/test/groovy/gov/hhs/cdc/trustedintermediary/e2e/DemographicsTest.groovy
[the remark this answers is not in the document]
jorg3lopez: I think the demographics code will get deleted. I like the idea of using the ignore annotation until then.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/OrderConverter.java
```diff
@@ -1,10 +0,0 @@
-package gov.hhs.cdc.trustedintermediary.etor.orders;
-
-/** Interface for converting things to orders and things in orders. */
-public interface OrderConverter {
```
jorg3lopez, line 4: All three methods from this interface belong to the general flow of the code. Are we going to create a section in the Transformation Engine that will be for our default transformations and a section for custom ones? I'm thinking on a way of separating the default behavior with the custom one.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/results/ResultConverter.java
```diff
@@ -1,6 +0,0 @@
-package gov.hhs.cdc.trustedintermediary.etor.results;
-
-/** Interface for converting things to results and things in results. */
-public interface ResultConverter {
```
jorg3lopez, line 4: Same comment as the one for OrderConverter

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRule.java
jorg3lopez: I wonder if we would benefit from having a separate log entry for each kind of exception. It might be robust but it will be more detailed.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/convertToOmlOrder.java
```diff
@@ -0,0 +1,24 @@
...
+/** Custom transformation to convert an ORM order to OML. */
+public class convertToOmlOrder implements CustomFhirTransformation {
+
+    private final MetricMetadata metadata =
+            ApplicationContext.getImplementation(MetricMetadata.class);
+
+    @Override
+    public void transform(FhirResource<?> resource, Map<String, String> args) {
+        Bundle bundle = (Bundle) resource.getUnderlyingResource();
+        HapiOrderConverterHelper.convertToOmlOrder(bundle);
```
jorg3lopez, line 21: I was thinking that the logic was going to be part of the transformation object.  But instead, it calls a the helper class that has the logic there. What would be the difference if we have the logic in here, so if we need to change  the logic, we go to`convertToOmlOrder.java` instead of the helper class, and the helper class can be a wrapper to that calls all the transformation objects?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/addEtorProcessingTag.java
```diff
@@ -0,0 +1,24 @@
...
+/** Custom transformation to add an ETOR tag to the message header of a FHIR bundle. */
+public class addEtorProcessingTag implements CustomFhirTransformation {
+
+    private final MetricMetadata metadata =
+            ApplicationContext.getImplementation(MetricMetadata.class);
+
+    @Override
+    public void transform(FhirResource<?> resource, Map<String, String> args) {
+        Bundle bundle = (Bundle) resource.getUnderlyingResource();
+        HapiMessageConverterHelper.addEtorTagToBundle(bundle);
```
jorg3lopez, line 21: Same as the transformation for `convertToOmlOrder`

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/addContactSectionToPatientResource.java
```diff
@@ -0,0 +1,24 @@
...
+/** Custom transformation to add a contact section to a patient resource. */
+public class addContactSectionToPatientResource implements CustomFhirTransformation {
+
+    private final MetricMetadata metadata =
+            ApplicationContext.getImplementation(MetricMetadata.class);
+
+    @Override
+    public void transform(FhirResource<?> resource, Map<String, String> args) {
+        Bundle bundle = (Bundle) resource.getUnderlyingResource();
+        HapiOrderConverterHelper.addContactSectionToPatientResource(bundle);
```
jorg3lopez, line 21: Same as the transformation for `convertToOmlOrder`

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiOrderConverterHelper.java
```diff
@@ -0,0 +1,46 @@
+package gov.hhs.cdc.trustedintermediary.external.hapi;
+
+import java.util.List;
+import org.hl7.fhir.r4.model.Bundle;
+import org.hl7.fhir.r4.model.Coding;
+import org.hl7.fhir.r4.model.Patient;
+
+/** Helper class with transformation methods that take an order Bundle and modifies it */
+public class HapiOrderConverterHelper {
```
jorg3lopez, line 9: Check the comments for the transformation objects. Thoughts on having the transformation logic as part of each object and having this class be a wrapper that couples with all the transformation objects?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRule.java
jcrichlake: It would still print the specific exception type

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiOrderConverterHelper.java
```diff
@@ -0,0 +1,46 @@
...
+                    "http://terminology.hl7.org/CodeSystem/v2-0003",
+                    "O21",
+                    "OML - Laboratory order");
+
+    private static final List<Coding> CODING_LIST =
+            List.of(
+                    new Coding(
+                            "http://terminology.hl7.org/CodeSystem/v3-RoleCode", "MTH", "mother"));
+
+    public static void convertToOmlOrder(Bundle order) {
```
jcrichlake, line 22: Could we make these methods more generic? Like ConvertToOrderType

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/orders/OrderConverter.java
basiliskus: This PR replaces the transformation in OrderConverter and ResultConverter. There's no distinction between default or custom behavior anymore. Is there a benefit to have them separated? If we want, we could add a "default" flag in the json file in a future PR. We could discuss this in eng. block

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/convertToOmlOrder.java
basiliskus: We could have the logic here. The idea of having the helper is to have functions that could be used by multiple transformations. I don't think we have that right now but I'd propose to leave this refactoring to a separate PR

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiOrderConverterHelper.java
basiliskus: That's a great idea. As I wrote on a response to Jorge, I think we should refactor this helper in a separate PR

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiOrderConverterHelper.java
basiliskus: Agreed that we need refactoring and move logic between helper and transformations. As mentioned above, I propose to do it in a separate PR

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRule.java
basiliskus: In that case, I'll collapse it into one exception. I don't see the benefit of logging each exception separately unless we're adding information that is not already in the exception itself

basiliskus: Done

luis-pabon-tf:
We're bypassing the SonarCloud duplication notice because we are saving refactoring of Rule Engines for a future PR.

luis-pabon-tf (approved):
This is a very thoroughly worked on PR, with good coverage and tested functionality. Code refactoring is needed, but due to the size of the PR it will be better off done separately.

PR #1055, opened 2024-04-30
Adding logging to save others in the future

jcrichlake:
# Adding log

Adding warning logging to alert folks about the randomly generated UUID in case they are trying to send data to report stream and are confused when the logged ID doesn't show up in the RS database.


basiliskus (approved):

jorg3lopez (approved):
Looks good.

jorg3lopez:
I approved it. Thoughts on adding logging to the other method calls to add similar acknowledgment?

PR #1061, opened 2024-05-03
Stream1/1044/e2e tests

jcrichlake:
# Add a PR title

## Issue

## Checklist


saquino0827:
I like the simplifications added to this, however this may only end up being a temporary solution. I **believe** we have another story later on down to include a more complex `OperationalOutcome`. But I think we can worry about that when we do the work on it.

saquino0827 (approved):

PR #1069, opened 2024-05-07
Add new transformation rules

basiliskus:
# Add new transformation rules

- Added UCSD baseline transformations
  - Update Receiving Application's Namespace Id (MSH-5.1) to 'EPIC'. Also remove Universal Id (MSH-5.2) and Universal Id Type (MSH-5.3) 
  - Remove Message Structure (MSH-9.3) from the Message Type (MSH-9)
  - Swap Placer Order Number (ORC-2) and Placer Group Number (ORC-4)
  - Remove Assigning Authority (PID-3.4) and Identifier Type Code (PID-3.5) from Patient Identifier List (PID-3)
  - Remove Name Type Code (PID-5.7) from Patient Name (PID-5)
- Refactored and added new methods to `HapiHelper` 

## Issue

#1024 

## Checklist


shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -53,16 +57,66 @@ public static void addMetaTag(
...
+        var header = findOrCreateMessageHeader(bundle);
+        var org = new Organization();
+        org.setName(name);
+        var destination = new MessageHeader.MessageDestinationComponent();
+        destination.setReceiver(new Reference(org));
+        header.setDestination(Collections.singletonList(destination));
+        bundle.addEntry(new Bundle.BundleEntryComponent().setResource(org));
+    }
+
+    public static void swapOBRsRenameThis(Bundle bundle) {
```
basiliskus, line 90: @luis-pabon-tf  I think this code should go in the custom transform method as it's specific to a transformation. The HapiHelper library should have more generic functions that can be used by multiple transformations

jorg3lopez:
What are your thoughts on having generic methods that will do similar tasks? For example, replacing a value with a hard coded value. This method can take the fhir path and the value. It can work for replacing the value of the sending facility, the receiving facility, some of the MSH fields, etc. The only thing that it could get enhance to take a map<K, V> pair. We can apply the same logic to fields that need to swap their values. SwapAwithB that will take a map<K , Map<K, V>>. This way, we only need to create one generic method for similar tasks of replacing a value and for swapping values. At least, if we can get away with making one transformation class for that, instead of creating multiple classes that will perform a similar function.

jorg3lopez:
Closed by accident. PR is open again.

basiliskus:
> What are your thoughts on having generic methods that will do similar tasks? For example, replacing a value with a hard coded value. This method can take the fhir path and the value. It can work for replacing the value of the sending facility, the receiving facility, some of the MSH fields, etc. The only thing that it could get enhance to take a map<K, V> pair. We can apply the same logic to fields that need to swap their values. SwapAwithB that will take a map<Map<K, V> , Map<K, V>>. This way, we only need to create one generic method for similar tasks of replacing a value and for swapping values. At least, if we can get away with making one transformation class for that, instead of creating multiple classes that will perform a similar function.

I agree, that would be ideal. We need to figure out how to set a value given the fhirpath with the Hapi library. So far we don't have a method for this (we only have one to get the value), but we should definitely look into that

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/addReceivingFacilityToMessageHeader.java
```diff
@@ -0,0 +1,33 @@
...
+import gov.hhs.cdc.trustedintermediary.external.hapi.HapiHelper;
+import gov.hhs.cdc.trustedintermediary.wrappers.MetricMetadata;
+import java.util.Collections;
+import java.util.Map;
+import org.hl7.fhir.r4.model.Bundle;
+import org.hl7.fhir.r4.model.MessageHeader;
+import org.hl7.fhir.r4.model.Organization;
+import org.hl7.fhir.r4.model.Reference;
+
+public class addReceivingFacilityToMessageHeader implements CustomFhirTransformation {
```
basiliskus, line 16: I think this should be Receiving Application, not facility. Is the organization name the value we want to update here? Receiving Application doesn't have a name in HL7, it has namespace id, a universal id and universal id type. All those are in the organization identifier. Do we know which one do we need to replace? Do we need to collapse them into one?

basiliskus: Looking at the spreadsheet, it seems we're collapsing them into MSH.5-1, which is the namespace (`identifier.where(extension.url = 'https://reportstream.cdc.gov/fhir/StructureDefinition/hl7v2Field' and extension.value = 'HD.1').value`)

luis-pabon-tf: Pushed a rename calling it `ReceivingApplication` instead across the board.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemoveMessageTypeStructure.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Override
+    public void transform(FhirResource<?> resource, Map<String, String> args)
+            throws RuleExecutionException {
+        try {
+            Bundle bundle = (Bundle) resource.getUnderlyingResource();
+            Coding coding = HapiHelper.getMessageTypeCoding(bundle);
+            String display = coding.getDisplay();
+            String delimiter = "^";
+            String[] displayArray = display.split("\\" + delimiter);
```
jcrichlake, line 24: 1. Does display need a null check?

2. If you use something like the below you can take display and get the ArrayList you want back in one line.  This would allow you to remove/refactor some of the following lines 

```suggestion
                        ArrayList displayArray = Arrays.stream(display.split("\\" + delimiter)).collect(Collectors.toCollection(ArrayList::new));
```

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemoveMessageTypeStructure.java
```diff
@@ -0,0 +1,38 @@
...
+            Coding coding = HapiHelper.getMessageTypeCoding(bundle);
+            String display = coding.getDisplay();
+            String delimiter = "^";
+            ArrayList<String> displayList =
+                    Arrays.stream(display.split("\\" + delimiter))
+                            .collect(Collectors.toCollection(ArrayList::new));
+            if (displayList.size() < 3) {
+                return;
+            }
+            displayList.remove(2);
```
somesylvie, line 31: This is just a contextual question - it seems like the other transforms are operating on entire FHIR fields, but this one's changing a substring of the display value of a `coding`.  Does the original MSH-9 not fill out the other parts of the `coding` (like the `system` and `code`) during the transition to FHIR inside ReportStream? How does this `display` field end up mapping back to MSH-9 on the other end when it goes back through ReportStream? I'm more familiar with FHIR in a FHIR-only context where the `display` would be like the least important part of a `coding`

basiliskus: Good question. RS uses `code` to determine the message type, which is mapped from the original HL7 MSH-9. But not all the fields are parsed and mapped into separate properties, only `code` is AFAIK. `display` is mapped back to `MSH-9` as is (I think)

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/SwitchPlacerOrderAndGroupNumbers.java
```diff
@@ -0,0 +1,54 @@
...
+                        var orc4_1 = serviceCoding.getCode();
+                        var orc4_2 = serviceCoding.getDisplay();
+
+                        // Switch values between ORC 2.1 and 4.1
+                        serviceIdentifier.setValue(orc4_1);
+                        serviceCoding.setCode(orc2_1);
+
+                        // Switch values between ORC 2.2 and 4.2
+                        serviceNamespaceExtension.setValue(new StringType(orc4_2));
+                        serviceCoding.setDisplay(orc2_2);
```
halprin, line 48: Was it to swap or just overwrite one with the other?

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/HappyPathMockClass.groovy
```diff
@@ -5,7 +5,6 @@ import gov.hhs.cdc.trustedintermediary.etor.ruleengine.transformation.CustomFhir
 import gov.hhs.cdc.trustedintermediary.external.hapi.HapiHelper
 import org.hl7.fhir.r4.model.Bundle
 import org.hl7.fhir.r4.model.Coding
-import org.hl7.fhir.r4.model.MessageHeader
 
 class HappyPathMockClass implements CustomFhirTransformation {
```
halprin, line 9: I know this class wasn't exactly part of this PR, but `HappyPathMockClass` doesn't exactly tell me what it does.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +90,120 @@ public static void addMetaTag(
...
+        destination.setReceiver(organizationReference);
+        messageHeader.setDestination(List.of(destination));
+        bundle.addEntry(new Bundle.BundleEntryComponent().setResource(receivingFacility));
+    }
+
+    // PID.3 - Patient Identifier List
+    public static List<Identifier> getPatientIdentifierList(Bundle bundle) {
+        Patient patient = (Patient) resourceInBundle(bundle, Patient.class);
+        if (patient == null) {
+            throw new NoSuchElementException("Patient not found in the bundle");
```
halprin, line 191: You throw `NoSuchElementException` in a couple of places in this class.  What are your thoughts about instead either returning `null` or an empty collection (for example, returning an empty `List` here)?  I haven't thought through this enough to have an opinion, but I'm curious what your thoughts are.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/ConvertToOmlOrder.java
```diff
@@ -9,8 +9,8 @@
 import java.util.Map;
 import org.hl7.fhir.r4.model.Bundle;
 
-/** Custom transformation to convert an ORM order to OML. */
-public class convertToOmlOrder implements CustomFhirTransformation {
+/** Converts an ORM order to OML. */
+public class ConvertToOmlOrder implements CustomFhirTransformation {
```
halprin, line 13: Should the `transform` method have a `try`/`catch` and thrown a `RuleExecutionException` like the newer transformations?  And then unit tests for this too.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/AddEtorProcessingTag.java
```diff
@@ -9,8 +9,8 @@
 import java.util.Map;
 import org.hl7.fhir.r4.model.Bundle;
 
-/** Custom transformation to add an ETOR tag to the message header of a FHIR bundle. */
-public class addEtorProcessingTag implements CustomFhirTransformation {
+/** Adds the ETOR tag to a message header meta tag for RS processing. */
+public class AddEtorProcessingTag implements CustomFhirTransformation {
```
halprin, line 13: Should the `transform` method have a `try`/`catch` and thrown a `RuleExecutionException` like the newer transformations?  And then unit tests for this too.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/AddContactSectionToPatientResource.java
```diff
@@ -12,8 +12,8 @@
 import org.hl7.fhir.r4.model.Coding;
 import org.hl7.fhir.r4.model.Patient;
 
-/** Custom transformation to add a contact section to a patient resource. */
-public class addContactSectionToPatientResource implements CustomFhirTransformation {
+/** Adds mother's contact section to patient. */
+public class AddContactSectionToPatientResource implements CustomFhirTransformation {
```
halprin, line 16: Should the `transform` method have a `try`/`catch` and thrown a `RuleExecutionException` like the newer transformations?  And then unit tests for this too.

halprin (commented):
Lot's of cool work here!  Mind adding JavaDocs to `RuleExecutionException`?  That can be in a separate PR if you like.

basiliskus:
> Lot's of cool work here! Mind adding JavaDocs to `RuleExecutionException`? That can be in a separate PR if you like.

Sounds good, I'll add that to a separate PR with miscellaneous improvements

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemoveMessageTypeStructure.java
```diff
@@ -0,0 +1,38 @@
...
+    public void transform(FhirResource<?> resource, Map<String, String> args)
+            throws RuleExecutionException {
+        try {
+            Bundle bundle = (Bundle) resource.getUnderlyingResource();
+            Coding coding = HapiHelper.getMessageTypeCoding(bundle);
+            String display = coding.getDisplay();
+            String delimiter = "^";
+            ArrayList<String> displayList =
+                    Arrays.stream(display.split("\\" + delimiter))
+                            .collect(Collectors.toCollection(ArrayList::new));
```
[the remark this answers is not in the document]
basiliskus: Even better, thanks! I'll change it

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/HappyPathMockClass.groovy
basiliskus: @saquino0827 I think you added this one?

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -40,9 +63,24 @@ public static <T extends Resource> Resource resourceInBundle(
         return resourcesInBundle(bundle, resourceType).findFirst().orElse(null);
     }
 
+    // MSH - Message Header
```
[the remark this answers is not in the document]
basiliskus: Done

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/ConvertToOmlOrder.java
basiliskus: Yes, I plan to add them in another PR

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/AddEtorProcessingTag.java
basiliskus: Same

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/AddContactSectionToPatientResource.java
basiliskus: Same

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemovePatientNameTypeCode.java
```diff
@@ -0,0 +1,31 @@
...
+    public void transform(final FhirResource<?> resource, final Map<String, String> args)
+            throws RuleExecutionException {
+        try {
+            Bundle bundle = (Bundle) resource.getUnderlyingResource();
+            Patient patient = (Patient) HapiHelper.resourceInBundle(bundle, Patient.class);
+            List<HumanName> names = patient.getName();
+            names.stream()
+                    .map(name -> name.getExtensionByUrl(HapiHelper.EXTENSION_XPN_HUMAN_NAME))
+                    .findFirst()
+                    .ifPresent(extension -> extension.removeExtension(HapiHelper.EXTENSION_XPN7));
```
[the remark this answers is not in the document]
basiliskus: Changed to remove the extension for all names

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Yeah, agreed. I actually was changing this in another PR. Initially I added it stop the flow but I think it's better to return null or empty collection

somesylvie: I think if it's something that should structurally _always_ be there it makes sense to throw an exception (like if you have a resource without an ID for instance?), but if it's an optional field, then returning null or empty and letting the caller decide if that's bad makes sense

basiliskus: Right. This is why it could go either way. In the context of the transformation that uses it, it's required. But since it's a helper function that could be used by other transformations that may not have that requirement, I think it's better to just return null

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemovePatientNameTypeCode.java
```diff
@@ -0,0 +1,34 @@
...
+    @Override
+    public void transform(final FhirResource<?> resource, final Map<String, String> args)
+            throws RuleExecutionException {
+        try {
+            Bundle bundle = (Bundle) resource.getUnderlyingResource();
+            Patient patient = HapiHelper.resourceInBundle(bundle, Patient.class);
+            for (HumanName name : patient.getName()) {
+                Extension extension =
+                        name.getExtensionByUrl(HapiHelper.EXTENSION_XPN_HUMAN_NAME_URL);
+                if ((extension != null)
```
saquino0827, line 25: Pretty sure I wrote this but I think it could be possible to add a condition in [validation_definitions.json](https://github.com/CDCgov/trusted-intermediary/blob/main/etor/src/main/resources/validation_definitions.json#L6).

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemoveMessageTypeStructure.java
```diff
@@ -0,0 +1,38 @@
...
+    public void transform(FhirResource<?> resource, Map<String, String> args)
+            throws RuleExecutionException {
+        try {
+            Bundle bundle = (Bundle) resource.getUnderlyingResource();
+            Coding coding = HapiHelper.getMessageTypeCoding(bundle);
+            String display = coding.getDisplay();
+            String delimiter = "^";
+            List<String> displayList =
+                    new ArrayList<>(Arrays.asList(display.split("\\" + delimiter)));
+            if (displayList.size() < 3) {
```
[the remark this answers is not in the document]
basiliskus: Even if we add a validation there, it will only log a warning

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemovePatientNameTypeCode.java
basiliskus: Same as above. I don't think we can enforce this conditions with the validation rule engine

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/HappyPathMockClass.groovy
saquino0827: Ah yes, I can rename it. It is a Happy Path class for ensuring the TransformationEngine runs a custom class when it is given a valid class as input.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemoveMessageTypeStructure.java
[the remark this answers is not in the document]
saquino0827: It wouldn't skip this transformation if we add a condition for it?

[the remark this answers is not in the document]
basiliskus: Oh I see what you mean. Yes, we could add it as a condition in transformation_definitions.json. We can discuss this in eng. block

etor/src/test/groovy/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/HappyPathMockClass.groovy
saquino0827: This has been renamed

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemoveMessageTypeStructure.java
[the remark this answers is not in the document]
halprin: Very weird.  I would have thought that this would have returned a modifiable list (as its JavaDocs state), but the unit test fails if we don't wrap it with `new ArrayList<>(...)`.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/SwitchPlacerOrderAndGroupNumbers.java
halprin: Any update on this?

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
halprin: The `getMSHMessageHeader` method still throws `NoSuchElementException`.  Is this on purpose?  Looks like that method is eventually called by our transformations and the exception isn't caught.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -35,14 +65,28 @@ public static <T extends Resource> Stream<T> resourcesInBundle(
...
+    public static MessageHeader getMSHMessageHeader(Bundle bundle) throws NoSuchElementException {
+        MessageHeader messageHeader = resourceInBundle(bundle, MessageHeader.class);
+        if (messageHeader == null) {
+            throw new NoSuchElementException("MessageHeader not found in the bundle");
+        }
+        return messageHeader;
+    }
+
+    public static MessageHeader createMSHMessageHeader(Bundle bundle) {
```
halprin, line 81: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
         messageHeader.setMeta(meta);
     }
 
-    public static void setMessageTypeCoding(Bundle order, Coding coding) {
-        var messageHeader = findOrInitializeMessageHeader(order);
+    // MSH-3 - Sending Application
+    public static MessageHeader.MessageSourceComponent getMSH3MessageSourceComponent(
```
halprin, line 100: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+            Bundle bundle) {
+        MessageHeader messageHeader = getMSHMessageHeader(bundle);
+        return messageHeader.getSource();
+    }
+
+    public static void setMSH3MessageSourceComponent(
```
halprin, line 106: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+            Bundle bundle, MessageHeader.MessageSourceComponent sendingApplication) {
+        MessageHeader messageHeader = getMSHMessageHeader(bundle);
+        messageHeader.setSource(sendingApplication);
+    }
+
+    public static MessageHeader.MessageSourceComponent createMSH3MessageSourceComponent() {
```
halprin, line 112: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+        return source;
+    }
+
+    // MSH-4 - Sending Facility
+    public static Organization getMSH4Organization(Bundle bundle) {
+        MessageHeader messageHeader = getMSHMessageHeader(bundle);
+        return (Organization) messageHeader.getSender().getResource();
+    }
+
+    public static void setMSH4Organization(Bundle bundle, Organization sendingFacility) {
```
halprin, line 124: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+    }
+
+    // MSH-5 - Receiving Application
+    public static MessageHeader.MessageDestinationComponent getMSH5MessageDestinationComponent(
+            Bundle bundle) {
+        MessageHeader messageHeader = getMSHMessageHeader(bundle);
+        return messageHeader.getDestinationFirstRep();
+    }
+
+    public static void setMSH5MessageDestinationComponent(
```
halprin, line 147: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+            Bundle bundle, MessageHeader.MessageDestinationComponent receivingApplication) {
+        MessageHeader messageHeader = getMSHMessageHeader(bundle);
+        messageHeader.setDestination(List.of(receivingApplication));
+    }
+
+    // MSH-6 - Receiving Facility
+    public static Organization getMSH6Organization(Bundle bundle) {
```
halprin, line 154: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+        MessageHeader messageHeader = getMSHMessageHeader(bundle);
+        return (Organization) messageHeader.getDestinationFirstRep().getReceiver().getResource();
+    }
+
+    public static void setMSH6Organization(Bundle bundle, Organization receivingFacility) {
```
halprin, line 159: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+        Coding coding = getMSH9Coding(bundle);
+        coding.setDisplay(value);
+    }
+
+    // PID - Patient
+    public static Patient getPIDPatient(Bundle bundle) {
+        return resourceInBundle(bundle, Patient.class);
+    }
+
+    public static Patient createPIDPatient(Bundle bundle) {
```
halprin, line 195: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+    // PID-3 - Patient Identifier List
+    public static Identifier getPID3Identifier(Bundle bundle) {
+        Patient patient = getPIDPatient(bundle);
+        if (patient == null) {
+            return null;
+        }
+        return patient.getIdentifierFirstRep();
+    }
+
+    public static void setPID3Identifier(Bundle bundle, Identifier identifier) {
```
halprin, line 210: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+    public static Identifier getPID3_4Identifier(Bundle bundle) {
+        Identifier identifier = getPID3Identifier(bundle);
+        if (identifier == null) {
+            return null;
+        }
+        Organization organization = (Organization) identifier.getAssigner().getResource();
+        return organization.getIdentifierFirstRep();
+    }
+
+    public static void setPID3_4Identifier(Bundle bundle, Identifier identifier) {
```
halprin, line 228: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+        if (patientIdentifier == null) {
+            return;
+        }
+        Organization organization = new Organization();
+        organization.setIdentifier(List.of(identifier));
+        Reference orgReference = createOrganizationReference(bundle, organization);
+        patientIdentifier.setAssigner(orgReference);
+    }
+
+    public static String getPID3_4Value(Bundle bundle) {
```
halprin, line 239: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+    // PID-3.5 - Identifier Type Code
+    public static Coding getPID3_5Coding(Bundle bundle) {
+        Identifier identifier = getPID3Identifier(bundle);
+        if (identifier == null) {
+            return null;
+        }
+        return identifier.getType().getCodingFirstRep();
+    }
+
+    public static void setPID3_5Coding(Bundle bundle, Coding coding) {
```
halprin, line 264: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+        Identifier identifier = getPID3Identifier(bundle);
+        if (identifier == null) {
+            return;
+        }
+        identifier.setType(new CodeableConcept().addCoding(coding));
+    }
+
+    public static String getPID3_5Value(Bundle bundle) {
```
halprin, line 272: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+    public static Extension getPID5Extension(Bundle bundle) {
+        Patient patient = getPIDPatient(bundle);
+        if (patient == null) {
+            return null;
+        }
+        HumanName name = patient.getNameFirstRep();
+        return name.getExtensionByUrl(HapiHelper.EXTENSION_XPN_HUMAN_NAME_URL);
+    }
+
+    public static void setPID5Extension(Bundle bundle) {
```
halprin, line 298: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+        Patient patient = getPIDPatient(bundle);
+        if (patient == null) {
+            return;
+        }
+        HumanName name = patient.getNameFirstRep();
+        name.addExtension(new Extension(HapiHelper.EXTENSION_XPN_HUMAN_NAME_URL));
+    }
+
+    // PID-5.7 - Name Type Code
+    public static String getPID5_7Value(Bundle bundle) {
```
halprin, line 308: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+        if (extension == null || !extension.hasExtension(HapiHelper.EXTENSION_XPN7_URL)) {
+            return null;
+        }
+        return extension
+                .getExtensionByUrl(HapiHelper.EXTENSION_XPN7_URL)
+                .getValue()
+                .primitiveValue();
+    }
+
+    public static void setPID5_7Value(Bundle bundle, String value) {
```
halprin, line 319: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+
+    public static void removePID5_7Extension(Bundle bundle) {
+        Extension extension = getPID5Extension(bundle);
+        if (extension != null && extension.hasExtension(HapiHelper.EXTENSION_XPN7_URL)) {
+            extension.removeExtension(HapiHelper.EXTENSION_XPN7_URL);
+        }
+    }
+
+    // ORC - Common Order
+    public static DiagnosticReport getDiagnosticReport(Bundle bundle) {
```
halprin, line 340: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+        return resourceInBundle(bundle, DiagnosticReport.class);
+    }
+
+    public static DiagnosticReport createDiagnosticReport(Bundle bundle) {
```
halprin, line 344: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+        DiagnosticReport diagnosticReport = new DiagnosticReport();
+        bundle.addEntry(new Bundle.BundleEntryComponent().setResource(diagnosticReport));
+        return diagnosticReport;
+    }
+
+    public static ServiceRequest getBasedOnServiceRequest(DiagnosticReport diagnosticReport) {
```
halprin, line 350: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+        return (ServiceRequest) diagnosticReport.getBasedOnFirstRep().getResource();
+    }
+
+    public static ServiceRequest createBasedOnServiceRequest(DiagnosticReport diagnosticReport) {
```
halprin, line 354: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+    }
+
+    public static void setEI2Value(Identifier identifier, String value) {
+        identifier
+                .getExtensionByUrl(EXTENSION_ASSIGNING_AUTHORITY_URL)
+                .getExtensionByUrl(EXTENSION_NAMESPACE_ID_URL)
+                .setValue(new StringType(value));
+    }
+
+    protected static Organization createOrganization() {
```
halprin, line 478: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
```diff
@@ -52,17 +96,419 @@ public static void addMetaTag(
...
+        Organization organization = new Organization();
+        String organizationId = UUID.randomUUID().toString();
+        organization.setId(organizationId);
+        return organization;
+    }
+
+    protected static MessageHeader.MessageDestinationComponent createMessageDestinationComponent() {
```
halprin, line 485: This is only called by test code.  Let's remove it.  If you want this as a helper method to make testing easier, add it to (or create new) a helper class that is just inside the `test/groovy/` area.

halprin (commented):
I have a few outstanding questions that I unresolved from my previous review, and here are some more comments.  Most, if not all, are about moving test-only methods to the test area of the code.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/AddContactSectionToPatientResource.java
jorg3lopez, line 16: I know the try catch was mentioned before. Thoughts on adding logging?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/AddEtorProcessingTag.java
jorg3lopez, line 13: Same for try catch and adding logging for context.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/AddContactSectionToPatientResource.java
halprin: And to clarify, logging should only be used if the exception is being _completely_ handled.  E.g. if you re-throw a new exception from the caught exception, don't log.  Something else will eventually completely handle that exception and log at that point.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/ConvertToOmlOrder.java
```diff
@@ -9,16 +9,16 @@
...
 
     private final MetricMetadata metadata =
             ApplicationContext.getImplementation(MetricMetadata.class);
 
     @Override
     public void transform(FhirResource<?> resource, Map<String, String> args) {
```
jorg3lopez, line 19: Same regarding the try catch and logging for context.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Hmm, good point. I think those creation methods will only be needed by tests

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Well, the message header is absolutely required so something would be very wrong if it's not there. There's another PR that will change how the exception handling is currently done but I'll check that it's handled correctly

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/AddContactSectionToPatientResource.java
basiliskus: Exception handling for transformations was refactored in PR 1104 and we no longer need try/catch here. In terms of adding logging, I think it could be useful to have debug level logging here, but info level logging may create too much noise. Unless there's something specific we need to log? Info logging for each transformation is already happening in the engine

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemoveMessageTypeStructure.java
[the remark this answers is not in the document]
basiliskus: Correct

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/SwitchPlacerOrderAndGroupNumbers.java
basiliskus: This was refactored. What we want is to swap them. Here's the description: "Update ORC-2 with content from ORC-4 in the ORU result message. Replace ORC-2.1 with content of ORC-4.1. Replace ORC-2.2 with content of ORC-4.2. Effectively, we're swapping ORC-2 for ORC-4 and vice versa"

jorg3lopez:
When the resulting message gets ingested by RS, it errors out during the FHR to HL7 translation.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
halprin: I'm willing to be pragmatic by leaving the methods that you think will be used as part of this story even though they aren't used in this specific PR.  But I'd like a follow-on PR at the end of this story to move the unused methods to the tests if they are unused by the real code.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Removed throwing the exception and returning now null for consistency

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Moved to test package

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Moved to test package

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Moved to test package

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Moved to test package

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Moved to test package

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Moved to test package

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Moved to test package

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Moved to test package

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Moved to test package

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Moved to test package

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Moved to test package

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Moved to test package

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Moved to test package

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Moved to test package

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Moved to test package

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Moved to test package

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Moved to test package

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Moved to test package

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelper.java
basiliskus: Actually, moved back to this helper class now that is being used after refactoring to handle null when calling getMSHMessageHeader

shared/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelperTest.groovy
```diff
@@ -60,15 +60,15 @@ class HapiHelperTest extends Specification {
         actualMessageHeader == messageHeader
     }
 
-    def "getMSHMessageHeader throws a NoSuchElementException if the message header does not exist"() {
+    def "getMSHMessageHeader returns null if the message header does not exist"() {
```
halprin, line 63: Since entire methods were moved out of `HapiHelper`, shouldn't entire tests be able to be deleted?

basiliskus: Good catch. Cleaned up the tests

shared/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/hapi/HapiFhirHelper.groovy
```diff
@@ -0,0 +1,198 @@
...
+import org.hl7.fhir.r4.model.HumanName
+import org.hl7.fhir.r4.model.Identifier
+import org.hl7.fhir.r4.model.MessageHeader
+import org.hl7.fhir.r4.model.Organization
+import org.hl7.fhir.r4.model.Patient
+import org.hl7.fhir.r4.model.Reference
+import org.hl7.fhir.r4.model.ServiceRequest
+import org.hl7.fhir.r4.model.StringType
+
+class HapiFhirHelper {
```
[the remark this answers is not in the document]
basiliskus: Moved to a shared location and merged them into one

halprin (approved):
Kaboom!  About that code coverage.  I'll take a swing at that.

jorg3lopez (approved):
Great job. The flow from FHIR to HL7 works!

jorg3lopez:
I think a good idea would be to make classes for each HL7 segment (ex: MSG, PID, OBX, OBR, etc) and add the helper methods for each segment resource. The HapiHelper will be the classes that uses these objects., thoughts? Of course, this would be done during the refactoring ticket.

Issue #1044, opened 2024-04-25
Load Tests Broken


JohnNKing:
# Bug

## Describe the Bug

As identified while reviewing #621, load tests are not working correctly.  They were passing at some point.  Consider walking back in history to see if there is an obvious culprit.  The purpose of this ticket is to fix the load tests, as they are architected today.  Out of scope, for example, is to have the load tests run against a deployed environment.

## Notes

- See [Slack](https://flexion.slack.com/archives/C06CP7E3NAF/p1713987260530329) for some details.
- [x] Fix load tests stream1/1044/e2eTests
  - [x] Fix non-db implementation for metadata
- [x] Fix e2e tests

jcrichlake:
Going through the history this works as of commit 68c767aa45ea4ebc7a16d524d2de7959a9a3b218

PR #1074, opened 2024-05-08
Use a (local) test DB for load tests so it doesn't wipe out the regular local DB

somesylvie:
# Use a (local) test DB for load tests so it doesn't wipe out the regular local DB
Add a `docker-compose.postgres-test.yml` file that sets up an `intermediary-test` DB and have the load tests use that DB - then the regular local `intermediary` DB won't be wiped out by running load tests

## Issue

#1044.

## Checklist

### DIFF
--- a/docker-compose.postgres-test.yml
+++ b/docker-compose.postgres-test.yml
@@ -1,0 +1,23 @@
+# Run the following to migrate...
+# liquibase update --changelog-file ./etor/databaseMigrations/root.yml --url jdbc:postgresql://localhost:5433/intermediary --username intermediary --password 'changeIT!' --label-filter '!azure'
+
+# Run the following to rollback...
+# liquibase rollback-count --changelog-file ./etor/databaseMigrations/root.yml --url jdbc:postgresql://localhost:5433/intermediary --username intermediary --password 'changeIT!' --count 2
+
+version: "3.7"
+
+services:
+  postgresql:
+    image: postgres:16
+    restart: unless-stopped
+    environment:
+      POSTGRES_DB: "intermediary-test"
+      POSTGRES_PASSWORD: "changeIT!" # pragma: allowlist secret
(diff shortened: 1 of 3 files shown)


docker-compose.postgres-test.yml
```diff
@@ -0,0 +1,17 @@
...
+services:
+  postgresql-test:
+    image: postgres:16
+    restart: unless-stopped
+    environment:
+      POSTGRES_DB: "intermediary-test"
+      POSTGRES_PASSWORD: "changeIT!" # pragma: allowlist secret
+      POSTGRES_USER: "intermediary"
+    ports:
+      - 5434:5432
```
basiliskus, line 12: Is it supposed to be `5432:5434`?

load-execute.sh
```diff
@@ -4,8 +4,8 @@ set -e
 start_api() {
     echo 'Starting API'
     export DB_URL=localhost
-    export DB_PORT=5433
-    export DB_NAME=intermediary
+    export DB_PORT=5434
```
basiliskus, line 7: Why was this port changed to 5434?

somesylvie: so that if you have the regular local DB already up and running, this won't conflict with or kill it

docker-compose.postgres-test.yml
somesylvie: The non-test version is 5433:5432, so this is just incrementing the 5433 to match. I think the first number is the Docker host port (so what any services on your machine or a DB client or whatever would connect to) and the second is the container's port inside docker. Since the two DBs (`intermediary` and `intermediary-test`) are running on different containers, it's okay they're both using 5432 (the default postgres port), and then the 5434 is what the load tests connect to

basiliskus (approved):

halprin (approved):

Issue #1001, opened 2024-04-03
DB Staging Connection Pooling Issue


JohnNKing:
# Backlog Task

Staging is timing out when saving metadata. We suspect this is caused by an issue with connection pooling via Hikari Connection Pool.

## Completion Criteria

- [x] We're able to save metadata in staging.

## Tasks
- [x] Check authentication and network settings
  - [x]  Azure managed identity service lifetime is 24 hours.
- [x] ~~Verify Azure Postgres flexible server settings (does that match PG's configuration and what we need?)~~
- [x] Implement new configuration in Hikari
  - [x] Set hikariCP `maxLifetime' to 23 hours and 57 minutes @saquino0827 
- [x] Verify Hikari settings
  - [x] ~~Test in the dev environment (before moving to staging)~~
    - [x]  N/A. Issue is not occurring in the Dev environment.
- [x]  Add logging for HikariCP token retrieval
- [x] Add Azure Log check monitoring for HikariCP invalid tokens
- [x] Rotate database password as new ones become available. - @halprin 
- [x] Verify Alert settings in staging
  - [x] Follow up tickets for any issue that arise.

jorg3lopez:
Do we have any logs for this?

saquino0827:
Since ticket #1001 got merged, I have not seen any alerts in Staging for the db connection issue. I feel we can close this task.

saquino0827:
Spoke with @halprin and we agreed to close this task.

PR #1102, opened 2024-05-16
Story-1001: Azure Log Alert Set up

saquino0827:
# Azure HikariCP token failure log alert

Azure Log Alert Set Up for the connection pooling error.

## Issue

- Adding an azure log search alert for checking a fatal connection error for HikariCP getting an invalid access token into the Postgresql database.
- One outstanding question I have is do we want our alert to auto mitigate or not.
  - Auto mitigate would keep the alert from firing until the alert resolves itself. AKA, doesn't meet the alert condition a set number of times or for a set period of time.
  - No auto mitigate would mean the alert fires every time the log statement in this PR is found, would be a bit more noisy and cost a little more.

[#1001](https://app.zenhub.com/workspaces/cdc-ti-633de70d939f6c00155f3722/issues/gh/cdcgov/trusted-intermediary/1001)

## Checklist


.github/workflows/terraform-deploy_reusable.yml
```diff
@@ -13,10 +13,10 @@ on:
         type: string
         required: false
         default: ""
-      TERRAFORM_APPLY_PARAMETERS:
-        type: string
-        required: false
-        default: ""
+#      TERRAFORM_APPLY_PARAMETERS:
+#        type: string
```
jcrichlake, line 17: Are these commented variables needed?

saquino0827: They are not, nice catch 👍

operations/template/alert.tf
```diff
@@ -0,0 +1,42 @@
...
+    QUERY
+
+  severity    = 3
+  frequency   = 10
+  time_window = 30
+
+  trigger {
+    operator  = "GreaterThan"
+    threshold = 5
+  }
```
halprin, line 41: This is saying that the log needs to show up 5 times within the last 30 minutes?  Or...?

halprin (commented):
Please update the PR description with the `#<ID number for this issue>`.

operations/template/alert.tf
saquino0827: Correct, open to thoughts on this check. Right now, the log is stateless so it will fire an alert  every time this check counts 5 of these logs in the last 30 minutes.

saquino0827: Now updated to fail on 1 log search found.

operations/template/alert.tf
```diff
@@ -0,0 +1,43 @@
...
+  resource_group_name = data.azurerm_resource_group.group.name
+  short_name          = "cdcti-alerts"
+
+  email_receiver {
+    name          = "cdcti-flexion-slack-email-receiver"
+    email_address = var.alert_slack_email
+  }
+}
+
+resource "azurerm_monitor_scheduled_query_rules_alert" "alert" {
```
halprin, line 12: Let's call this some more descriptive than `alert`.  Perhaps `database_token_expired_alert`.
```suggestion
resource "azurerm_monitor_scheduled_query_rules_alert" "database_token_expired_alert" {
```

operations/template/alert.tf
```diff
@@ -0,0 +1,43 @@
+resource "azurerm_monitor_action_group" "monitor" {
```
halprin, line 1: Consider something more descriptive than `monitor`.  Perhaps `notify_slack_email`.
```suggestion
resource "azurerm_monitor_action_group" "notify_slack_email" {
```

halprin (commented):
Looks really good.  I just have two comments on making the Terraform names more descriptive.

saquino0827:
> Looks really good. I just have two comments on making the Terraform names more descriptive.

I appreciate the review, this has been updated.

PR #1119, opened 2024-05-29
1001: Rotate Database Password

halprin:
# Rotate Database Password

The reason we failed to connect to the database after some amount of time is because Azure rotates the password.  Our code would only grab the password at the very beginning and use it forever.  After Azure rotated the password, our code would start failing because we would try to authenticate with an outdated password.

The fix is to grab the latest password whenever Hikari requests the password when it creates a new database connection for its pool.

I also reduced the database connection lifetime to 58 minutes.  The Hikari default is 30 minutes.

_Note: I will be waiting a bit longer before I merge to confirm this fix continues to work in the Internal environment._

## Issue

#1001.

## Checklist


saquino0827 (approved):
Nice work, it looks good!

Issue #1121, opened 2024-05-30
Load Tests Failing


JohnNKing:
# Bug

## Describe the Bug

Locally ran Load tests in main fail our threshold of `> 1%` failures. We need to look into what is causing slow downs and fix them.

## Impact

Potential blocker of stories due to load tests failures.

## To Reproduce

1. Run load tests script in the repo

## Expected Behavior

Load tests pass with  `< 1%` failure rate

## Screenshots

<img width="1376" alt="Main_Load_Tests" src="https://github.com/CDCgov/trusted-intermediary/assets/26700499/3a515fa3-d119-487b-8b18-b09b0a684be7">

## Logs

We will need to add logs to help triage this issue.

## Version

The [git commit ](https://github.com/CDCgov/trusted-intermediary/commit/07845b1d8aeb75ff98a461be42b7e603e63b0f9f) sha of the application that the bug found on.

## Additional Context

Auth, Order and Result testing are the main bulk of the issues.

PR #1130, opened 2024-06-05
Add Containerization To Load Test

jorg3lopez:
# Run API in Docker During Load Test

This PR adds the logic to run the API in a docker container when performing the load test.

## Issue

#1121


README.md
```diff
@@ -104,12 +104,17 @@ The `test` directory contains the tests.  The `main` directory contains our cust
 
 #### Load Testing
 
-Load tests are completed with [Locust.io](https://docs.locust.io/en/stable/installation.html).  Run the load tests by
-running...
+Load tests are completed with [Locust.io](https://docs.locust.io/en/stable/installation.html).
+Please make sure your `/shared/src/main/resoirces/.env` file does not contain the variable `REPORT_STREAM_URL_PREFIX=http://localhost:7071`.
```
basiliskus, line 108: This is not needed if you set the environment variable when running TI in the bash script

docker-load-execute.sh
```diff
@@ -0,0 +1,104 @@
...
+        sleep 5
+    done
+
+    echo 'API is responding'
+}
+
+warm_up_api() {
+    echo 'Warming up API...'
+    pem=$(pwd)/mock_credentials/organization-trusted-intermediary-private-key-local.pem
+    token=$(jwt encode --exp='+5min' --jti $(uuidgen) --alg RS256  --no-iat -S@${pem})
```
basiliskus, line 51: You could read the JWT from the mock_credentials .jwt file to remove the dependency with the `jwt` CLI

docker-load-execute.sh
```diff
@@ -0,0 +1,104 @@
...
+
+    echo ${token}
+
+    echo 'Warming up auth...'
+    tiAuthResponse=$(curl --request POST "http://localhost:8080/v1/auth/token" \
+    --header "Content-Type: application/x-www-form-urlencoded" \
+    --data-urlencode "scope=trusted-intermediary" \
+    --data-urlencode "client_assertion=${token}")
+
+    tiToken=$(echo "${tiAuthResponse}" | jq -r '.access_token')
```
basiliskus, line 61: You could use grep to remove the dependency with `jq`. Otherwise `jq` should be added as a requirement

docker-load-execute.sh
saquino0827: Part of the reason for this is to also warm up the Auth endpoint. I found that individually calling the endpoints helps those endpoints perform faster.

docker-load-execute.sh
saquino0827: Got it working using Grep.

README.md
saquino0827: This has been changed to set `REPORT_STREAM_URL_PREFIX=` as a blank value.

docker-load-execute.sh
basiliskus: What I mean is that you could replace the `jwt` CLI command, for something like this: `cat mock_credentials/trusted-intermediary-valid-token.jwt`

docker-load-execute.sh
```diff
@@ -0,0 +1,109 @@
...
+
+start_database() {
+    echo 'Starting database'
+    docker compose -f docker-compose.postgres-test.yml up -d
+    sleep 2
+    echo "Database started"
+}
+migrate_database() {
+    echo 'Migrating database'
+    liquibase update --changelog-file ./etor/databaseMigrations/root.yml --url jdbc:postgresql://localhost:5434/intermediary-test --username intermediary --password 'changeIT!' --label-filter '!azure'
```
basiliskus, line 28: Would it be worth extracting the port number into a variable and place that at the top of the file? Given the port may change more frequently? Also, this port and `DB_PORT` above are different. Why is that?

docker-load-execute.sh
saquino0827: Ah I see, for some reason I thought the `mock_credentials/trusted-intermediary-valid-token.jwt` was the latter step in auth but that makes sense. I'll change this asap.

basiliskus (approved):

Issue #1133, opened 2024-06-05
CA Result OBX Mapping Issues


JohnNKing:
# Bug

## Describe the Bug

We've identified a few issues with OBX mapping for the CA sample ORU: https://github.com/CDCgov/trusted-intermediary/blob/main/examples/CA/002_CA_ORU_R01.hl7

- ~~OBX.2 of TN is rejected by RS (and if changed to XTN, OBX.5 is lost)~~
  - To confirm with CDPH if this would be used for UCSD
- OBX.2 of CE are losing OBX.3 and OBX.5
  - Confirmed that data is lost when converting HL7 => FHIR 
  - Mappings for OBX.3 and OBX.5 are here: `prime-router/metadata/HL7/catchall/hl7/segments/OBX/Observation.yml`
  - [RS is looking into this issue](https://github.com/CDCgov/prime-reportstream/issues/14685)
- OBX.6: For μmol, 'μ' has an incorrect encoding in the resulting HL7
  - Fixed in PR #1134  

## Impact

This will prevent integration with CA.

## To Reproduce

Submit the sample ORU (https://github.com/CDCgov/trusted-intermediary/blob/main/examples/CA/002_CA_ORU_R01.hl7) to ReportStream.

## Expected Behavior

The content of the input and output HL7 matches

## Screenshots

n/a

## Logs

n/a

## Version

Current main

## Additional Context
Related backlog items:
- #905
- #1071

basiliskus:
I just found out that the CE OBX mapping issue is a known bug in RS. A ticket will be created and someone from RS will take a look next week

basiliskus:
A ticket to address the OBX CE issue was created in RS: https://github.com/CDCgov/prime-reportstream/issues/14685

basiliskus:
Waiting for RS to look into this issue. Tagged as blocked

jorg3lopez:
Findings:
- [HL7-2.5.1](https://hl7-definition.caristix.com/v2/HL7v2.5.1/Tables/0125) supports CE datatype for OBX.2
- [HL7-2.7](https://hl7-definition.caristix.com/v2/HL7v2.7/Tables/0125) does not support CE datatype for OBX.2
[...]

PR #1134, opened 2024-06-06
Fix message encoding when sending HTTP request back to RS

basiliskus:
# Fix message encoding when sending HTTP request back to RS

The HTTP request message was being encoded using `ISO-8859-1` by default, which doesn't encode correctly characters like `µ`. Fixed by explicitly encoding with `UTF-8`

## Issue

#1133


halprin (approved):
Nice find.  Also, do we need to do something similar in our e2e tests in the HTTP client there too?

basiliskus:
> Nice find. Also, do we need to do something similar in our e2e tests in the HTTP client there too?

I just checked the e2e implementation and it already specifies UTF-8

PR #1142, opened 2024-06-12
Add validation and transformation ADRs

basiliskus:
# Add validation and transformation ADRs

Added ADRs for the Validation and Transformation Rules Engine


adr/022-transformation-engine.md
```diff
@@ -0,0 +1,28 @@
+# 22. Transformation Engine
+
+Date: 2024-06-11
+
+## Decision
+
+As with the [Validation Rule Engine](https://github.com/CDCgov/trusted-intermediary/blob/f11891504f5303657cfe1b5c61a515013265f239/adr/021-validation-engine.md), we will use a variation of the [Rules Engine Pattern](https://deviq.com/design-patterns/rules-engine-pattern) to design and implement a transformation rule engine that can transform incoming FHIR messages given a condition. Each transformation will be represented as a rule, specifying conditions that must be met and the corresponding failure message if the transformation fails. Initially, rules will be stored in a JSON file, with considerations on transitioning to a database for scalability.
```
tjohnson7021, line 7: @basiliskus I updated this whole document based on the Validation Rules Engine ADR so I'm assuming it will need corrections/updates

adr/021-validation-engine.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Status
+
+Accepted.
+
+## Context
+
+The intermediary needs to validate FHIR messages based on SME research and partner-specific requirements, which are subject to change. The Rules Engine Pattern enables flexible, scalable, and maintainable rule management.
+
+## Impact
```
tjohnson7021, line 17: I came up with these generically

adr/001-architecture-decision-records.md
```diff
@@ -21,6 +21,21 @@ We want to record our architectural decisions so that...
...
 
+## Impact
+
+_The outcomes of the decision, both positive and negative. This section explains the impact of the decision, such as trade-offs, risks, and what needs to be done to implement it._
+
+**Positive:**
+- Allows us to show scope of considerations at the time of the decision
+
+**Negative:**
+- Time consumption?
```
tjohnson7021, line 32: This is just an example

tjohnson7021:
I mostly added an 'Impact' section to each of these ADRs and then reworded the entries to be more succinct.

adr/021-validation-engine.md
```diff
@@ -0,0 +1,29 @@
+# 21. Validation Engine
+
+Date: 2024-06-11
+
+## Decision
+
+We will implement a Validation Engine using the [Rules Engine Pattern](https://deviq.com/design-patterns/rules-engine-pattern) to validate incoming FHIR messages based on given conditions. Each validation will be represented as a rule, specifying conditions that must be met and the corresponding failure message if validation fails. Initially, rules will be stored in a JSON file, with considerations on transitioning to a database for scalability.
```
somesylvie, line 7: For the validation and transformation engines, it might be useful to call out what RS does vs what we do in TI and how those are different

basiliskus: I think what RS would be outside of the scope of these ADRs as they are not about the transformation or validation themselves, but about the framework that implements them. The engines are meant to be generic and the FHIR that gets validated or transformed could be coming from anywhere (in fact, I think in the future could be a good idea to spin off the engines into their own services)

adr/020-azure-alerts.md
```diff
@@ -1,10 +1,10 @@
-# 1. Architecture Decision Records
+# 20. Azure Alerts
```
halprin, line 1: Good catch.

adr/020-azure-alerts.md
```diff
@@ -14,12 +14,31 @@ Accepted.
...
+
+- **Trigger:** Any logged failures with database connections.
+
+
+- **Configuration:** Alerts are stateful (auto-mitigation); set to `fired` status to reduce noise from frequent or duplicate alerts.
+
+
+- **Notification:** Alerts sent to a Slack channel via email until PagerDuty is operational.
+
+## Consequences:
```
halprin, line 30: Change this to Impact?  Also, there is no need for a `:` because this is a heading and already indicates that it applies to what is to come.

```suggestion
## Impact
```

adr/001-architecture-decision-records.md
```diff
@@ -21,6 +21,21 @@ We want to record our architectural decisions so that...
 - New team members who join can see why we made the decisions we made.
 - The team can revise or revisit decisions with more confidence and context.
 
+## Impact
+
+_The outcomes of the decision, both positive and negative. This section explains the impact of the decision, such as trade-offs, risks, and what needs to be done to implement it._
+
+**Positive:**
```
halprin, line 28: I like how in the validation engine one these headings were turned into legitimate headings by using `###`.  Thoughts on going converting this, and the Azure Alerts one, to using the formal headings for positive, negative, and risk?

```suggestion
### Positive
```

halprin (approved):

PR #1188, opened 2024-07-23
Fix local storage bugs occurring when running in Windows environment

jbiskie:
## Fix local storage bugs occurring when running in Windows environment

This PR:
- Updates folder creation to avoid using Posix attributes when run in Windows 
- Changes the delimiter in the metadata file path from a pipe ("|") to a dash ("-") to work with both *nix and Windows
- Refactors the System.getProperty(...) calls out of the file storage classes into the ApplicationContext

## Issue

#1191

## Checklist

## Testing instructions

- Run unit tests
- Run end to end tests


shared/src/main/java/gov/hhs/cdc/trustedintermediary/context/ApplicationContext.java
```diff
@@ -176,7 +181,24 @@
         return Paths.get(System.getProperty("user.dir")).getParent();
     }
 
+    public static Path getTempPath() {
+        return Paths.get(System.getProperty("java.io.tmpdir"));
```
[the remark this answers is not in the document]
tjohnson7021: Does this matter? How does it get resolved?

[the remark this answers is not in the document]
jorg3lopez: Good question @tjohnson7021. I think we should look into [java.nio.file.Files.createTempDirectory](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#createTempDirectory-java.nio.file.Path-java.lang.String-java.nio.file.attribute.FileAttribute...-). 

example:  

`Path temp = Files.createTempDirectory("random-directory");`

README.md
```diff
@@ -59,6 +59,11 @@ Apply all outstanding migrations:
 liquibase update --changelog-file ./etor/databaseMigrations/root.yml --url jdbc:postgresql://localhost:5433/intermediary --username intermediary --password 'changeIT!' --label-filter '!azure'
 ```
 
+If running in Windows, use double quotes instead:
```
basiliskus, line 62: If OSX and linux work with double quotes, we can just replace the single quotes for double. I can try this in OSX

basiliskus: Hmm I tried it and it doesn't work, so nevermind

jorg3lopez (approved):
Great job. Everything works in Windows environment and in Unix-base environment.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/context/ApplicationContext.java
```diff
@@ -172,11 +176,48 @@ public static String getEnvironment() {
...
+
     public static Path getExamplesPath() {
         return getRootPath().resolve("examples");
     }
+
+    public static Path createTempFile(String fileName) throws IOException {
+        Path tempFilePath = getTempPath().resolve(fileName);
+
+        if (!Files.exists(tempFilePath)) {
+            if (tempFilePath.getFileSystem().supportedFileAttributeViews().contains("posix")) {
```
luis-pabon-tf, line 195: This line and line 211 could be changed to call a method that checked this, so that the method can contain the explanation for this.

PR #1190, opened 2024-07-24
Transformation to update OBR-2 with ORC-4

basiliskus:
# Transformation to update OBR-2 with ORC-4

There was a missing transformation for UCSD that we're adding here: `Update OBR-2 with content from ORC-4 in the ORU result message. Replace OBR-2.1 with original content of ORC-4.1. Replace OBR-2.2 with original content of ORC-4.2`. We are expanding the `SwapPlacerOrderAndGroupNumbers` transformation to include this changes

## Issue

#1194

## Checklist


etor/src/main/java/gov/hhs/cdc/trustedintermediary/plugin/path/Hl7FhirMappingPath.java
```diff
@@ -7,11 +7,16 @@
  * defines the HL7 field names.
  */
 public enum Hl7FhirMappingPath {
-    PLACER_ORDER_NUMBER(
+    PLACER_ORDER_NUMBER_ORC_2(
             "ORC.2",
             """
      Bundle.entry.resource.ofType(ServiceRequest).identifier.where(type.coding.code = 'PLAC').value
      """),
+    PLACER_ORDER_NUMBER_OBR_2(
```
jorg3lopez, line 15: Is this path being used?

tjohnson7021: I've got _no usages_ coming up too; Is this a question of whether we need this now or will need it?

shared/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelperTest.groovy
```diff
@@ -601,6 +601,39 @@ class HapiHelperTest extends Specification {
         HapiHelper.getORC21Value(sr) == null
     }
 
+
+    // OBR-2 - Placer Order Number
+    def "obr-2 methods work as expected"() {
```
jcrichlake, line 606: I think this test could use a better name

etor/src/main/java/gov/hhs/cdc/trustedintermediary/plugin/path/Hl7FhirMappingPath.java
basiliskus: I don't think it is right now, but we're also using this FHIR paths as reference. Even if not used, it's useful for us to document all the HL7 FHIR paths

shared/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/hapi/HapiHelperTest.groovy
basiliskus: This test is covering all the OBR-2 related methods in HapiHelper. Other tests in this file follow similar pattern, so if we change this one we should change the other ones too. Do you have any suggestions?

jcrichlake: Maybe obr-2 sets value to input or something?

basiliskus: It's also testing the getters. What about `obr-2 getter and setters methods work as expected`? The `work as expected` is also a pattern we've been using for naming tests

jcrichlake: Sure, or you can leave it if you think it's fine

etor/src/main/java/gov/hhs/cdc/trustedintermediary/plugin/path/Hl7FhirMappingPath.java
jorg3lopez: I’ll approve the PR, but I would recommend that we add paths as and when they are needed. This way, we can ensure that every documented path is relevant and necessary at the time of its inclusion.

jorg3lopez (approved):
Great work! The swap works as expected, good job.

PR #1201, opened 2024-07-29
Ignore Terraform tags managed by CDC

jbiskie:
# Add a PR title

The CDC automatically adds a set of tags to Terraform resources. This PR adds a set of ignore tags so we do not delete them each time we apply Terraform and CDC re-adds them.

## Issue

N/A


jbiskie:
Closing. Incorrect branches.

Issue #1208, opened 2024-07-31
Reduction in Privileges in Azure


halprin:
# Backlog Task

The CDC unilaterally reduced the permissions of the role assigned to our GitHub deployer and our user accounts.  We need to work with the CDC to understand why this was done, push back if we can, and subsequently modify our Terraform depending on the discussions with the CDC.

## Completion Criteria

- [x] We come to a conclusion of having our permissions restored in some form or fashion or not.
- [x] Our Terraform is updated to remove items from management that we can no longer modify.

## Tasks

- [x] Research what has been impacted by this priviledge change (other than our NSGs). - @halprin 
- [x] Respond to Bob's e-mail. - @halprin 
- [x] Update TI's Terraform and successfully deploy to Dev environment.
- [x] Update TI's Terraform and successfully deploy to PR environment. 
- [x] Remove the same resources from the Internal environment's TF state and successfully deploy.
- [x] Push TI's Terraform updates to Staging and Production.
- [x] Update SFTP Ingestion's Terraform and successfully deploy to Dev environment.
- [x] Push SFTP Ingestion's Terraform updates to Staging.

## Other Notes

- Any other notes to help clarify this task for the team

somesylvie:
CDC wants to own route tables, network security groups, and peering settings

For envs where these resources already exist, we'll need to remove them from terraform state and convert them in our TF from resources to data sources (removing them from state tells TF we're not managing those and it shouldn't delete them). By using the Azure Portal Cloud Shell, we can interact with the Terraform CLI while logged in as ourselves (this will be the easiest way to manage this stuff for the -SU accounts). Clone our repo, CD into our TF directories, terraform init, then run commands to remove the relevant resources

For envs where these resources don't yet exist (prod for ingestion service), we'll need to submit a helpdesk ticket to get the relevant resources created

PR #1223, opened 2024-08-02
#1208 -  Reduction in Privileges in Azure

jherrflexion:
Updated terraform to reflect changes needed to support the privilege changes.

## Issue
#1208


halprin (commented):
I think there are three more things we need to do.

1. Also remove from state (and maybe convert to `data`) the routes used on the route tables.
2. Also remove from state (and maybe convert to `data`) the network security rules used on the network security groups.
3. Perhaps create some of these things in the Flexion domain environments (Internal and PRs).

jcrichlake (approved):

halprin (approved):

Issue #1212, opened 2024-07-31
Fix unreleased resources in TI


jcrichlake:
# DevEx/OpEx

Priority: High

Unreleased resources exist in `TransformationRuleEngine:43`, `validationRuleEngine:43`, and `OpenApiReaderImplementation:30`.  For the InputStream  we need a finally block that closes the stream

## Tasks

- [x] Close streams in reporting files.

## Additional Context

Add any other context or screenshots about the work here.

PR #1234, opened 2024-08-08
Close unreleased InputStreams in the Rule Engines and ApiReader

luis-pabon-tf:
# Close unreleased InputStreams in the Rule Engines and ApiReader

Added closing statements in such a way that output is unchanged for a couple methods.

## Issue
#1212 

## Checklist

### DIFF
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
@@ -47,7 +47,6 @@ import gov.hhs.cdc.trustedintermediary.external.reportstream.ReportStreamSenderH
 import gov.hhs.cdc.trustedintermediary.wrappers.FhirParseException;
 import gov.hhs.cdc.trustedintermediary.wrappers.HapiFhir;
 import gov.hhs.cdc.trustedintermediary.wrappers.Logger;
-import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
@@ -144,8 +143,7 @@ public class EtorDomainRegistration implements DomainConnector {
     @Override
     public String openApiSpecification() throws UnableToReadOpenApiSpecificationException {
         String fileName = "openapi_etor.yaml";
-        return OpenApiReaderImplementation.getInstance()
-                .openAsString(fileName, StandardCharsets.UTF_8);
+        return OpenApiReaderImplementation.getInstance().openAsString(fileName);
     }
 
     DomainResponse handleOrders(DomainRequest request) {
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
@@ -47,6 +47,7 @@ import gov.hhs.cdc.trustedintermediary.external.reportstream.ReportStreamSenderH
 import gov.hhs.cdc.trustedintermediary.wrappers.FhirParseException;
 import gov.hhs.cdc.trustedintermediary.wrappers.HapiFhir;
 import gov.hhs.cdc.trustedintermediary.wrappers.Logger;
+import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
@@ -143,7 +144,8 @@ public class EtorDomainRegistration implements DomainConnector {
     @Override
     public String openApiSpecification() throws UnableToReadOpenApiSpecificationException {
         String fileName = "openapi_etor.yaml";
-        return OpenApiReaderImplementation.getInstance().openAsString(fileName);
+        return OpenApiReaderImplementation.getInstance()
+                .openAsString(fileName, StandardCharsets.UTF_8);
     }
 
     DomainResponse handleOrders(DomainRequest request) {
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRuleEngine.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRuleEngine.java
@@ -11,6 +11,7 @@ import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.List;
 import javax.inject.Inject;
+import org.apache.commons.io.IOUtils;
 
 /** Implements the RuleEngine interface. It represents a rule engine for transformations. */
 public class TransformationRuleEngine implements RuleEngine {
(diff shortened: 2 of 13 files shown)


shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/openapi/OpenApiReaderImplementation.java
```diff
@@ -29,8 +29,11 @@ public static OpenApiReaderImplementation getInstance() {
     public String openAsString(String fileName) throws UnableToReadOpenApiSpecificationException {
         InputStream stream = getClass().getClassLoader().getResourceAsStream(fileName);
         try {
-            return new String(
-                    Objects.requireNonNull(stream).readAllBytes(), StandardCharsets.UTF_8);
+            var out =
+                    new String(
+                            Objects.requireNonNull(stream).readAllBytes(), StandardCharsets.UTF_8);
+            stream.close();
```
basiliskus, line 35: Can you use a try/catch managed scope to close the InputStream? Something like this: `try (InputStream stream = getClass().getClassLoader().getResourceAsStream(fileName)) {`. Also, now that you're here, it may be a good idea to pass the utf8 charset as an argument

basiliskus (commented):
Looks good, I just left a couple of comments/suggestions

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/openapi/OpenApiReaderImplementation.java
luis-pabon-tf: Ok, updated it

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/openapi/OpenApiReaderImplementation.java
```diff
@@ -22,15 +22,17 @@ public static OpenApiReaderImplementation getInstance() {
...
-    public String openAsString(String fileName) throws UnableToReadOpenApiSpecificationException {
-        InputStream stream = getClass().getClassLoader().getResourceAsStream(fileName);
-        try {
-            return new String(
-                    Objects.requireNonNull(stream).readAllBytes(), StandardCharsets.UTF_8);
+    public String openAsString(String fileName, Charset charset)
+            throws UnableToReadOpenApiSpecificationException {
+        try (InputStream stream = getClass().getClassLoader().getResourceAsStream(fileName)) {
+            var out = new String(Objects.requireNonNull(stream).readAllBytes(), charset);
+            stream.close();
```
basiliskus, line 34: Is the `stream.close()` needed now that you're scoping the InputStream? Not sure about Java, but in C# the the resource would be closed when leaving the scope

halprin: I don't believe the `stream.close()` is needed now.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
```diff
@@ -143,7 +144,8 @@ public Map<HttpEndpoint, Function<DomainRequest, DomainResponse>> domainRegistra
     @Override
     public String openApiSpecification() throws UnableToReadOpenApiSpecificationException {
         String fileName = "openapi_etor.yaml";
-        return OpenApiReaderImplementation.getInstance().openAsString(fileName);
+        return OpenApiReaderImplementation.getInstance()
+                .openAsString(fileName, StandardCharsets.UTF_8);
```
halprin, line 148: I wouldn't pass in `StandardCharset.UTF_8`.  There's no reason for `.openAsString(...)` to support other charsets.  Instead, hardcode `StandardCharsets.UTF_8` in `OpenApiReaderImplementation#openAsString`.

This applies to the other places where we call `OpenApiReaderImplementation.getInstance()
                 .openAsString(...)`.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/openapi/OpenApiReaderImplementation.java
```diff
@@ -22,15 +22,17 @@ public static OpenApiReaderImplementation getInstance() {
...
     @Override
-    public String openAsString(String fileName) throws UnableToReadOpenApiSpecificationException {
-        InputStream stream = getClass().getClassLoader().getResourceAsStream(fileName);
-        try {
-            return new String(
-                    Objects.requireNonNull(stream).readAllBytes(), StandardCharsets.UTF_8);
+    public String openAsString(String fileName, Charset charset)
+            throws UnableToReadOpenApiSpecificationException {
+        try (InputStream stream = getClass().getClassLoader().getResourceAsStream(fileName)) {
+            var out = new String(Objects.requireNonNull(stream).readAllBytes(), charset);
```
halprin, line 33: I don't think we need the `Objects.requireNonNull()`.  `stream` will be not be `null` if we get to this point.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/TransformationRuleEngine.java
```diff
@@ -53,6 +54,7 @@ public void ensureRulesLoaded() throws RuleLoaderException {
                             ruleLoader.loadRules(resourceStream, new TypeReference<>() {});
                     rules.addAll(parsedRules);
                     rulesLoaded = true;
+                    IOUtils.closeQuietly(resourceStream);
```
halprin, line 57: I wouldn't use this.  Use the "try with resources" mechanism.  I.e. the same thing that @basiliskus mentioned [below](https://github.com/CDCgov/trusted-intermediary/pull/1234/files#r1711680750).

halprin: For the `catch`, I would catch `IOException | NullPointerException` and throw a `throw new RuleLoaderException("File not found: " + ruleDefinitionsFileName, new FileNotFoundException());` like above.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/openapi/OpenApiReaderImplementation.java
luis-pabon-tf: Updated - got rid of that closing call

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
luis-pabon-tf: I had made this change for this PR. Reverted the change, the diff should be a bit smaller now too

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/openapi/OpenApiReaderImplementation.java
luis-pabon-tf: I have that there to silence a warning on the IDE, but since we're catching the NullPointerException that it's worried about it's probably a false positive. Will do!

halprin (approved):

basiliskus (approved):

Issue #1216, opened 2024-08-01
Constructor Invokes Overridable Function


jcrichlake:
# DevEx/OpEx

Priority: Medium

The `OrderResponse.java` code has constructors invoking an overridable function.  Fix it

## Additional Context

Add any other context or screenshots about the work here.

PR #1245, opened 2024-08-14
Fix `Constructor Invokes Overridable Function` Fortify finding

basiliskus:
# Fix `Constructor Invokes Overridable Function` Fortify finding

Refactored `OrderResponse` to make it immutable, and along with `ResultResponse` which was already immutable, turned them into records

## Issue

#1216 

## Checklist


luis-pabon-tf (approved):

jbiskie (approved):

basiliskus:
> Looks good. This is a simple immutable class, would a record construct work here?

Yes, I think that'd be a good idea. I'm not sure why we made into a class in the first place. I think we used to have methods aside from getters and setters, but now we probably don't need a class

halprin:
Kick ass on making these records!

Issue #1220, opened 2024-08-01
Implement Azure Deployment Slots


saquino0827:
# Story
As California, so that our results are not lost, I need Azure deployment slots to ensure Zero Downtime Deployment(ZDD).

## Pre-conditions
- [x] N/A

## Acceptance Criteria
- [x] We have Zero Downtime Deployments for our TI application code.
- [x] We have Zero Downtime Deployments for our RS SFTP Ingestion application code.
- [x] We have tested that we have minimal traffic loss during a zero downtime deployment.

### Research
- [x] *Research work needed to complete the story*
- [x] Foundational: *High-level research that will support this and future efforts*

###  Engineering
- [x] Implement Azure Deployment Slots
  - [x] For TI
  - [x] For ingestion service web app
- [x] Update GitHub actions to swap slots after deploy to lower slot
  - [x] For TI
  - [x] For ingestion service web app
- [x] Confirm Deployment Slots Swap and Function properly
  - [X] For TI
  - [X] For Ingestion Service Web App

Notes for letting pre-live be healthy but not handle traffic/events (having pre-live be healthy enables quickly rolling back by swapping slots manually):
- For TI, web traffic drives the app. Post-swap, the pre-live slot is unhealthy because it's missing DB creds. We should give both slots all the same app settings - either by putting them in a var that both reference, or just copy/pasta the values
- For Ingestion Service app, without the queues, the pre-live service is unhealthy after swap (`main.go` completes, and then restarts later). To keep it healthy, we could either:
  - Share all the settings like with TI, but have a different `env` value and check the `env` before reading queues (still need some kind of infinite loop to keep the app running)
  - Not share any more settings than we currently are, but add an infinite loop to `main` to keep the app up even when we don't have queues to use

## Definition of Done
- [x] Documentation tasks completed
  - [x] Documentation and diagrams created or updated
    - [x] ADRs (`/adr` folder)
    - [x] Main `README.md`
    - [x] Other READMEs in the repo
  - [x] [Threat model](https://lucid.app/lucidchart/693ff5f9-c289-4585-9e74-a7f5abb36d69/edit?invitationId=inv_24d83a2b-b0fe-436d-8bd2-027b17beeb33) updated
- [x] Code quality tasks completed
  - [x] Code refactored for clarity and no design/technical debt
  - [x] Adhere to separation of concerns; code is not tightly coupled, especially to 3rd party dependencies
[...]

PR #1264, opened 2024-08-22
first attempt at copying the slot/deployment settings from ingestion service

somesylvie:
# Add a PR title

## Issue

## Checklist

### DIFF
--- a/.github/workflows/deploy_reusable.yml
+++ b/.github/workflows/deploy_reusable.yml
@@ -113,7 +113,7 @@ jobs:
       uses: azure/webapps-deploy@v3
       with:
         app-name: ${{ inputs.APP }}
-        slot-name: production
+        slot-name: pre-live
         images: '${{ inputs.REGISTRY }}/${{ inputs.REPO }}:${{ github.sha }}'
 
     - name: Azure Swap Slots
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy_reusable.yml
```diff
@@ -116,6 +116,12 @@ jobs:
         slot-name: production
```
jherrflexion, line 116: Should this be pre-live?

.github/workflows/deploy_reusable.yml
```diff
@@ -116,6 +116,12 @@ jobs:
...
         images: '${{ inputs.REGISTRY }}/${{ inputs.REPO }}:${{ github.sha }}'
 
+    - name: Azure Swap Slots
+      uses: azure/CLI@v2
+      with:
+        inlineScript: |
+          az webapp deployment slot swap -n ${{ inputs.APP }} --slot pre-live --resource-group csels-rsti-${{ inputs.ENVIRONMENT }}${{ github.event.number }}-moderate-rg
```
jherrflexion, line 123: Do we want auto swap here? Otherwise I believe we need to do --target-slot(?)

somesylvie: `--target-slot` defaults to `production` when not specified, and `auto swap` unfortunately doesn't work for linux

.github/workflows/deploy_reusable.yml
somesylvie: yep, good catch

jcrichlake (approved):

Issue #1211, opened 2024-07-31
Dockerfiles using default root user


jcrichlake:
# DevEx/OpEx

Priority: High

Need to specify a user for the dockerfiles in TI, the TI implementation guide, and the SFTP service

## Tasks

- [x] Play around with SFTP Ingestion Docker image. - @halprin 
- [x] Fix TI Docker image.
- [x] Fix RS SFTP Ingestion Service Docker image.

## Additional Context

Add any other context or screenshots about the work here.

PR #1275, opened 2024-08-27
Low Privilege User for Running the Application Take 2

halprin:
# Description

Updated our `Dockerfile` to use a lower privilege user for running the application.

Java cannot find the jar file if it resides in the lower-permission user's home directory in the Docker container while running on our Azure Web App.  Java sees the jar file just fine when running the Docker container locally.  This is what bit us in the previous PR #1261.

So, now we keep the work directory in the home folder, but we put the jar file in a location where any user can access (`/usr/local/bin/`).  This mimics the RS SFTP Ingestion service's `Dockerfile`.

This was tested in the Internal environment successfully.

## Issue

#1211.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,9 @@
 FROM amazoncorretto:17.0.12-alpine
 
+# Uppdate dependencies and clear the dependency cache.
 RUN apk update && apk -U upgrade && rm -rf /var/cache/apk/*
 
+# Create and use a lower permission (non-root) user.
 RUN adduser -S myLowPrivilegeUser
 USER myLowPrivilegeUser
 
@@ -13,6 +15,8 @@ WORKDIR /home/myLowPrivilegeUser/app/
 # If we put the jar file into the myLowPrivilegeUser's home directly, the container fails to run in Azure.
 COPY --chown=myLowPrivilegeUser ./app/build/libs/app-all.jar /usr/local/bin/app.jar
 
+# Run the service.
 CMD ["java", "-jar", "/usr/local/bin/app.jar"]
 
+# Inform Docker that this container listens on the specified port.
 EXPOSE 8080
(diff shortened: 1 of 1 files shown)


jcrichlake (approved):

Dockerfile
```diff
@@ -1,18 +1,18 @@
-# Use Linux-Alpine image
```
pluckyswan, line 1: Why remove this comment? Is it causing an issue or outdated info?

Dockerfile
```diff
@@ -1,18 +1,18 @@
...
+# which is in the myLowPrivilegeUser home folder because we are running as that user instead of root.
+WORKDIR /home/myLowPrivilegeUser/app/
 
-# Add application JAR to created folder
-COPY ${JAR_LIB_FILE} app.jar
+# Copy the jar file into /usr/local/bin/ because it seemingly needs to go to a location that any user can access.
+# If we put the jar file into the myLowPrivilegeUser's home directly, the container fails to run in Azure.
+COPY --chown=myLowPrivilegeUser ./app/build/libs/app-all.jar /usr/local/bin/app.jar
 
-# Run the api
```
pluckyswan, line 14: Same as previous comment, including line 17.

Dockerfile
halprin: It wasn't causing an issue nor was it outdated, but we can see we're using Alpine Linux due to it being called out directly in the line below.  The `-alpine` tag is a very common way to dictate that we're using Alpine Linux.

Dockerfile
halprin: I thought it was obvious what was happening.  Perhaps not?  Do you think otherwise, and I should put it back?

pluckyswan: I prefer having step by step because I am unfamiliar and it helps having plain context. I was mainly curious since they were there before

halprin: Sure, I'll add them back.

halprin: Done.

Issue #1214, opened 2024-07-31
Fortify: Use Customer Managed Keys


halprin:
# DevEx/OpEx

Fortify is warning about us not using a customer managed key for our storage accounts and container registry.  We should look into what it takes to create and manage a customer managed key.  If it is like AWS KMS, I'm all for it.  If it takes additional work and management, I don't think it is worth the price of admission.

## Tasks

- [x] Figure out what it takes to make and create a customer managed key.
- [x] Do this for TI.
  - [x] Storage accounts.
  - [x] Container registries.
- [x] Do this for RS SFTP Ingestion.
  - [x] Storage accounts.
  - [x] Container registries.

## Additional Context

Category: Azure Terraform Misconfiguration: Storage Account Missing Infrastructure Encryption
Primary Location: storage.tf: 1, app.tf: 2

- https://learn.microsoft.com/en-us/azure/storage/common/customer-managed-keys-overview
- https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/key_vault_key

rajeev-flexion:
We can use azure key vault for this
1) creating a key vault and key
2) Associating the key with storage account

https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account_customer_managed_key

PR #1276, opened 2024-08-27
Setting up keyvault settings and adding permissions to app

jcrichlake:
# Description

We are changing settings for our azure storage accounts so that they utilize Azure Customer managed keys.  

Run something like `az storage account update --identity-type SystemAssigned --name cdctipr1276 --resource-group csels-rsti-pr1276-moderate-rg` in all the environments shortly before merging.

## Issue

[Fortify card](https://github.com/CDCgov/trusted-intermediary/issues/1214)

### DIFF
--- a/operations/template/docs.tf
+++ b/operations/template/docs.tf
@@ -44,4 +44,5 @@ resource "azurerm_key_vault_managed_storage_account" "key_vault_storage_account"
   storage_account_id           = azurerm_storage_account.docs.id
   storage_account_key          = azurerm_key_vault_key.customer_managed_key.name
   regenerate_key_automatically = false
+  regeneration_period          = "P360D"
 }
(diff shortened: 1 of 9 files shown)


operations/template/docs.tf
```diff
@@ -27,7 +27,22 @@ resource "azurerm_storage_account" "docs" {
...
     ]
   }
+
+  identity {
+    type = "SystemAssigned"
+  }
+}
+
+
+resource "azurerm_key_vault_managed_storage_account" "key_vault_storage_account" {
```
somesylvie, line 41: it looks like this resource isn't referenced by anything else - is that expected?

halprin: Yes, this is expected.  There can be plenty of "leaf" nodes (or resources) that aren't referenced.

pluckyswan (approved):

saquino0827 (approved):

PR #1284, opened 2024-08-30
Keep pre-live deployment slot healthy

somesylvie:
# Description
This PR duplicates the main app_settings in the pre-live slot. When some of these values are not set, the pre-live slot becomes unhealthy and cannot be swapped back. Since TI is driven by http traffic, it's safe to have all the values set in both the production and pre-live slots, since the slot status itself will control traffic.

## Issue

### DIFF
--- a/operations/template/app.tf
+++ b/operations/template/app.tf
@@ -119,8 +119,7 @@ resource "azurerm_linux_web_app" "api" {
     }
   }
 
-  #   When adding new settings that are needed for the live app but shouldn't be used in the pre-live
-  #   slot, add them to `sticky_settings` as well as `app_settings` for the main app resource
+  # New settings may also need to be added to the pre-live slot's app_settings
   app_settings = {
     DOCKER_REGISTRY_SERVER_URL    = "https://${azurerm_container_registry.registry.login_server}"
     ENV                           = var.environment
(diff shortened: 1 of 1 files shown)


halprin:
I'm going to merge `main` into this branch because Jeff and I [merged a change](https://github.com/CDCgov/trusted-intermediary/pull/1285) that makes deploys better.

operations/template/app.tf
```diff
@@ -137,8 +137,7 @@ resource "azurerm_linux_web_app" "api" {
     }
   }
 
-  #   When adding new settings that are needed for the live app but shouldn't be used in the pre-live
-  #   slot, add them to `sticky_settings` as well as `app_settings` for the main app resource
+  # New settings may also need to be added to the pre-live slot's app_settings
```
jcrichlake, line 140: Can we expand this comment to include when to add to both as opposed to just one.

operations/template/app.tf
```diff
@@ -227,9 +221,18 @@ resource "azurerm_linux_web_app_slot" "pre_live" {
   }
 
   app_settings = {
-    DOCKER_REGISTRY_SERVER_URL = "https://${azurerm_container_registry.registry.login_server}"
-
-    ENV = var.environment
+    DOCKER_REGISTRY_SERVER_URL    = "https://${azurerm_container_registry.registry.login_server}"
+    ENV                           = var.environment
+    REPORT_STREAM_URL_PREFIX      = "https://${local.rs_domain_prefix}prime.cdc.gov"
+    KEY_VAULT_NAME                = azurerm_key_vault.key_storage.name
```
jcrichlake, line 227: Are we not using sticky settings now at all?

somesylvie: When we didn't share all the settings, we were getting errors in `pre-live` after swapping (which would prevent swapping back). Since TI is driven off of http calls, it seemed simpler to share all the settings than to figure out which ones are required for the app to be healthy, which makes the sticky settings unneeded for TI

operations/template/app.tf
somesylvie: We updated the comment to just say to always add them to both - for TI I think this is cleaner

jcrichlake (approved):

PR #1288, opened 2024-09-02
Transformation: Modified OBR-4.3 codingSystem

tjohnson7021:
# CA Transform: OBR-4.3 Modification

- Modified file so that ORUs passed through ReportStream will be modified to always have 'CDPHGSPEAP' on OBR-4.3.
- Updated transform and added a unit test to retain value of OBR-4.4 when an alternateId is not specified in the configuration.
- Minor refactoring to the `UpdateUniversalServiceIdentifier` transform. Added constants for the hardcoded string values and updated the private `updateAlternateCodingId` method to use a string parameter rather than a Map since we only use one value.

## Checklist


etor/src/main/resources/transformation_definitions.json
```diff
@@ -181,8 +181,8 @@
           "name": "UpdateUniversalServiceIdentifier",
           "args": {
             "checkValue": "54089-8",
-            "codingSystem": "LN",
-            "alternateId": "CDPHGSPEAP"
+            "codingSystem": "CDPHGSPEAP",
+            "alternateId": ""
```
tjohnson7021, line 185: I wasn't sure if this should remain or not because it was the same value, so I left it here and in the tests.

luis-pabon-tf: It's only used on `UpdateUniversalServiceIdentifier`, so let's see if we can take it out safely.

jbiskie:
Do we want the alternate identifier to be emptied out if it is not specified in the transform definition? I ran a message through with this OBR:

`OBR|1|7241234515^FormNumber||54089-8^NB Screen Panel Patient AHIC^OBR43VALUE^OBR44VALUE^OBR45VALUE^OBR46VALUE|||202407111346|||||||||^HUBBARD^EUSTRATIA||||||20240711034913|||F`

After the transform, the final result looks like this:

`OBR|1|189609160^HospOrdNumber||54089-8^NB Screen Panel Patient AHIC^CDPHGSPEAP^^OBR45VALUE^OBR46VALUE|||202407111346|||||||||^HUBBARD^EUSTRATIA||||||20240711034913|||F`

Note that OBR-4.4 is now empty.

CA is only sending OBR-4.1 and OBR-4.2, so it shouldn't matter for the current requirement, but it could make a difference if we receive messages that use the alternate in the future.

@dbgolson

jbiskie:
> Do we want the alternate identifier to be emptied out if it is not specified in the transform definition? I ran a message through with this OBR:
> 
> `OBR|1|7241234515^FormNumber||54089-8^NB Screen Panel Patient AHIC^OBR43VALUE^OBR44VALUE^OBR45VALUE^OBR46VALUE|||202407111346|||||||||^HUBBARD^EUSTRATIA||||||20240711034913|||F`
> 
> After the transform, the final result looks like this:
> 
> `OBR|1|189609160^HospOrdNumber||54089-8^NB Screen Panel Patient AHIC^CDPHGSPEAP^^OBR45VALUE^OBR46VALUE|||202407111346|||||||||^HUBBARD^EUSTRATIA||||||20240711034913|||F`
> 
> Note that OBR-4.4 is now empty.
> 
> CA is only sending OBR-4.1 and OBR-4.2, so it shouldn't matter for the current requirement, but it could make a difference if we receive messages that use the alternate in the future.
> 
> @dbgolson

Discussed with @dbgolson and @luis-pabon-tf on 9/6/2024 and updated the task. We agreed that the transform should retain the original values for subfields OBR-4.4 through 4.6.

luis-pabon-tf:
As it is now, all tests pass with full coverage on the new code. Approval pending verifying the original values...

jbiskie:
The commit I just made adds a new unit test for the empty OBR-4.4 case and updates the transform. We still need to run a new end to end test. I will do that in the morning.

luis-pabon-tf (approved):

PR #1289, opened 2024-09-03
ADR for deployment slots

somesylvie:
# Description
This PR is for the ADR for deployment slots

## Issue

### DIFF
--- a/adr/023-deployment-slots.md
+++ b/adr/023-deployment-slots.md
@@ -1,0 +1,40 @@
+# 23. Deployment Slots for Zero Downtime Deploys
+
+Date: 2024-09-03
+
+## Decision
+We will use Azure Web App Deployment Slots to facilitate zero-downtime deploys of the TI app
+
+## Status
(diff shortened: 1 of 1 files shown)


adr/023-deployment-slots.md
```diff
@@ -0,0 +1,40 @@
+# 23. Deployment Slots for Zero Downtime Deploys
```
sfradkin, line 1: I'd like to see a reason why we are using deployment slots even though there are some significant negative impacts and risks.

sfradkin (changes requested):

adr/023-deployment-slots.md
somesylvie: Basically the other options were even worse - Kubernetes, roll our own, or switch to AWS (or abandon zero-downtime deploys). We can add a mention of the alternatives

sfradkin (approved):
excellent, thanks!

Issue #1326, opened 2024-09-13
Do not cache bad keys.


scleary1cs:
# DevEx/OpEx

Make DevEx/OpEx backlog item for bad keys not to be cached.

## Tasks

- [x] Update TI service's caching system to remove cache'd token when an Invalid Token is caught. @saquino0827 

## Additional Context

Root cause for session 1 during game day exercises.

saquino0827:
A PR is open here: https://github.com/CDCgov/trusted-intermediary/pull/1333/files

saquino0827:
The PR has been merged.

PR #1333, opened 2024-09-13
Remove Invalid Tokens from cache when found

saquino0827:
# Description

- Uncache our TI token if an InvalidToken is caught and found.

## Issue
#1326 

## Checklist

### DIFF
--- a/app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthRequestValidator.java
+++ b/app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthRequestValidator.java
@@ -14,7 +14,8 @@ import javax.inject.Inject;
 /**
  * This class is used to check the validity of a http request. It has methods that extract the
  * bearer token, check if the token is empty or null, and if the token is valid. For example,
- * expired tokens, empty tokens, or tokens not signed by our private key, will be invalid.
+ * expired tokens, empty tokens, or tokens not signed by our private key, will be invalid. Tokens
+ * are cached on first use, and removed if invalid.
  */
 public class AuthRequestValidator {
 
--- a/app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthRequestValidator.java
+++ b/app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthRequestValidator.java
@@ -21,7 +21,7 @@ public class AuthRequestValidator {
     private static final AuthRequestValidator INSTANCE = new AuthRequestValidator();
 
     @Inject private AuthEngine jwtEngine;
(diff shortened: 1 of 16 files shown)


app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthRequestValidator.java
```diff
@@ -21,10 +21,12 @@ public class AuthRequestValidator {
     private static final AuthRequestValidator INSTANCE = new AuthRequestValidator();
 
     @Inject private AuthEngine jwtEngine;
-    @Inject private Cache keyCache;
+    @Inject Cache keyCache;
     @Inject private Secrets secrets;
     @Inject private Logger logger;
 
+    String ourPublicKey = "trusted-intermediary-public-key-" + ApplicationContext.getEnvironment();
```
jcrichlake, line 28: Do we want this public key name hardcoded here? And if so do we need a #pragma allow secret

saquino0827: It previously existed, but I moved it to a higher level so it's accessible in tests. I can add the #pragma allow secrets if we want but I also don't think it got flagged before as a secret.

jcrichlake: Cool

jcrichlake (approved):

app/src/main/java/gov/hhs/cdc/trustedintermediary/auth/AuthRequestValidator.java
```diff
@@ -21,10 +21,12 @@ public class AuthRequestValidator {
     private static final AuthRequestValidator INSTANCE = new AuthRequestValidator();
 
     @Inject private AuthEngine jwtEngine;
-    @Inject private Cache keyCache;
+    @Inject Cache keyCache;
```
luis-pabon-tf, line 24: Seems this is the only inject we actually use in this class. Would a small doc be good?

halprin: Are you sure?  I see `jwtEngine`, `secrets`, and `logger` being used below.

saquino0827: I added some extra comments on the class. I think all the injects are used unless I'm misunderstanding what you mean.

pluckyswan (commented):
Checklist needs to be updated (remove items or check them off)

Issue #1322, opened 2024-09-13
Update hurl scripts. 


scleary1cs:
# DevEx/OpEx

Update hurl scripts to automatically update the MSH headers to be a value such that we wouldn't really direct messages to a real partner.

## Tasks

- [x] Update hurl script.
  - [x] Figure out how to scramble correct fields.
  - [x] Figure out how to not commit scrambles.
  - [x] Scramble.

## Additional Context

Add any other context or screenshots about the work here.

pluckyswan:
Asked Jorge for info. on handling file path separators in TI.

pluckyswan:
PR is out for approval.

pluckyswan:
Based on feedback, we don't want to scramble it but want to make it a Flexion receiver.

pluckyswan:
Will work Basillo on Mon to continue working on issue.

PR #1342, opened 2024-09-20
Updated RS Hurl Script to Replace MSH Header When Not Toggled

pluckyswan:
# Replace MSH Header When Not Toggled

Added so by default when hurl script for ReportStream is ran, MSH header is replaced to prevent out bound to a real receiver. Added a flag to allow out bound in command line. 

## Issue

https://github.com/CDCgov/trusted-intermediary/issues/1322

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,5 +35,5 @@ localfileorder.json
 
 logback.log
 
-# Ignore folder created when substituting receiver fields in MSH Header
-substituted_receivers
+# Ignore folder created when substituting header in MSH Header
+substituted_header
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,6 @@ __pycache__
 localfileorder.json
 
 logback.log
+
+# Ignore folder created when substituting receiver fields in MSH Header
+substituted_receivers
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,3 @@ __pycache__
 localfileorder.json
 
 logback.log
-
-# Ignore folder created when substituting header in MSH Header
-substituted_header
--- a/scripts/hurl/rs/hrl
+++ b/scripts/hurl/rs/hrl
@@ -120,25 +120,20 @@ if [ -z "$secret" ]; then
 fi
 
 
-if [ "$allow_outbound" = true ]; then
-  file_name=$(basename $fpath)
-  echo "$fpath"
-  mkdir -p "$CDCTI_HOME/substituted_receivers/$fpath" && touch -f "$CDCTI_HOME/substituted_receivers/$fpath/$file_name"
-  echo "$file_name"
-  receiverFieldsToSubstitute=$(head -n1 "$CDCTI_HOME/$fpath" | cut -d "|" -f 5,6)
-  echo "$receiverFieldsToSubstitute"
-  substitutedReceiverFields=$(sed "s/$receiverFieldsToSubstitute/!!!!!!!**********substituted|ToPreventFileOutput**********!!!!!!!/" "$CDCTI_HOME/$fpath" > "$CDCTI_HOME/substituted_receivers/$fpath/$file_name")
-  echo "$substitutedReceiverFields"
-  fpath=$CDCTI_HOME/substituted_receivers/$fpath
-
-  echo "$fpath"
-  exit 1
-fi
-
 if [ "$allow_outbound" = false ]; then
-  echo "$allow_outbound"
-  exit 1
+  file_path=$(dirname "$fpath")
+  substitutedFilePathPrefix="$CDCTI_HOME/substituted_receivers"
+#  Make temporary directory to store scrambled file
+  mkdir -p "$substitutedFilePathPrefix/$file_path" && touch -f "$substitutedFilePathPrefix/$fpath"
+#  Grab MSH Header receiver fields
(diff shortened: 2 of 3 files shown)


scripts/hurl/rs/hrl
```diff
@@ -29,6 +30,7 @@ show_help() {
     echo "    -i <SUBMISSION_ID>    The submissionId to call the history API with (Required for history API)"
     echo "    -v                    Verbose mode"
     echo "    -h                    Display this help and exit"
+    echo "    -p                    Produce a real message, receiver will not be Flexion specific"
```
basiliskus, line 33: It's not clear to me what this flag is doing given the help message. Is the flag just scrambling the receiver facility details?

scripts/hurl/rs/hrl
```diff
@@ -114,6 +119,21 @@ if [ -z "$secret" ]; then
...
 fi
 
+
+if [ "$allow_outbound" = false ]; then
+  file_path=$(dirname "$fpath")
+  substitutedFilePathPrefix="$CDCTI_HOME/substituted_receivers"
+#  Make temporary directory to store scrambled file
+  mkdir -p "$substitutedFilePathPrefix/$file_path" && touch -f "$substitutedFilePathPrefix/$fpath"
+#  Grab MSH Header receiver fields
+  receiverFieldsToSubstitute=$(head -n1 "$CDCTI_HOME/$fpath" | cut -d "|" -f 5,6)
```
basiliskus, line 129: Instead of replacing MSH-5/6, we can replace the whole MSH segment. That will also handle some other edge cases. For orders, we can replace it with: `MSH|^~\&|Sender Application^sender.test.com^DNS|Sender Facility^0.0.0.0.0.0.0.0^ISO|Receiver Application^0.0.0.0.0.0.0.0^ISO|Receiver Facility^simulated-lab-id^DNS|20230101010000-0000||ORM^O01^ORM_O01|111111|T|2.5.1||||||||||` and for results:  `MSH|^~\&|Sender Application^sender.test.com^DNS|Sender Facility^0.0.0.0.0.0.0.0^ISO|Receiver Application^0.0.0.0.0.0.0.0^ISO|Receiver Facility^simulated-hospital-id^DNS|20230101010000-0000||ORU^R01^ORU_R01|111111|T|2.5.1||||||||||`

scripts/hurl/rs/hrl
pluckyswan: Description updated to be more concise. Thoughts?

basiliskus: The line above also needs updating. I think the update is good, but I'd remove the `Produce a real message` part which can create confusion. I'd also add something explaining the default behavior. It could be something like: `By default, the MSH segment is replaced to contain a Flexion receiver in order to avoid sending messages to partners inadvertently. When using this flag, the MSH segment will not be replaced`

scripts/hurl/rs/hrl
pluckyswan: Implemented suggestions as discussed!

scripts/hurl/rs/hrl
pluckyswan: I added this to the readme and option description.

.gitignore
```diff
@@ -34,3 +34,6 @@ __pycache__
 localfileorder.json
 
 logback.log
+
+# Ignore folder created when substituting receiver fields in MSH Header
+substituted_receivers
```
basiliskus, line 39: Looks like this folder was renamed to `substituted_header`?

basiliskus: Can the temporary file be deleted after calling the hurl command instead?

pluckyswan: I implemented your suggestion

scripts/hurl/rs/hrl
```diff
@@ -114,6 +120,30 @@ if [ -z "$secret" ]; then
...
+  msh_header=$(head -n1 "$CDCTI_HOME/$fpath")
+# Check message type
+  if [[ "$msh_header" = *"^R01"* ]]; then
+    msh_header_replacement="MSH|^~\&|Sender Application^sender.test.com^DNS|Sender Facility^0.0.0.0.0.0.0.0^ISO|Receiver Application^0.0.0.0.0.0.0.0^ISO|Receiver Facility^simulated-hospital-id^DNS|20230101010000-0000||ORU^R01^ORU_R01|111111|T|2.5.1||||||||||"
+  elif [[ "$msh_header" = *"^O01"* ]]; then
+    msh_header_replacement="MSH|^~\&|Sender Application^sender.test.com^DNS|Sender Facility^0.0.0.0.0.0.0.0^ISO|Receiver Application^0.0.0.0.0.0.0.0^ISO|Receiver Facility^simulated-lab-id^DNS|20230101010000-0000||ORM^O01^ORM_O01|111111|T|2.5.1||||||||||"
+  elif [[ "$msh_header" = *"^O21"* ]]; then
+      msh_header_replacement="MSH|^~\&|Sender Application^sender.test.com^DNS|Sender Facility^0.0.0.0.0.0.0.0^ISO|Receiver Application^0.0.0.0.0.0.0.0^ISO|Receiver Facility^simulated-lab-id^DNS|20230101010000-0000||OML^O21^OML_O21|111111|T|2.5.1||||||||||"
+  else
+      echo "File does not contain any valid subfields."
```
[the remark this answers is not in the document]
pluckyswan: Updated it

PR #1306, opened 2024-09-10
CA observation transforms from local code to LOINC or PLT

luis-pabon-tf:
# CA observation transforms from local code to LOINC or PLT

This PR adds a new transform that converts local observation codes in the CA result messages to LOINC or PLT codes. The tranform checks each observation in the message, and the mapped LOINC/PLT codes are added to the observation's coding resource as an additional code. Following the ReportStream conversion from fhir back to HL7, the mapped codes are placed in OBX-3.1/2/3, and the local code remains in OBX-3.4/5/6.

## Issue
#1277 

## Checklist

- [x] I have added tests to cover my changes
- [x] I have added logging where useful (with appropriate log level)
- [x] I have added JavaDocs where required
- [x] I have updated the documentation accordingly

**Note**: You may remove items that are not applicable

PR #1362, opened 2024-09-26
Transformation Engine To Use Map<String,Object> As Args

jorg3lopez:
# Complex Map As Args For Transformation Engine

This PR adds the logic for the trasnformation engine to use a Map<String,Object> intead of a Map<String,String> allowing for more flexibility when passing parameters when using the transformation_definitions.json file.

## Issue
#1366 

## Checklist

- [x] All tests pass

### DIFF
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemoveObservationRequests.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemoveObservationRequests.java
@@ -24,13 +24,8 @@ public class RemoveObservationRequests implements CustomFhirTransformation {
     @Override
     public void transform(FhirResource<?> resource, Map<String, Object> args) {
         Bundle bundle = (Bundle) resource.getUnderlyingResource();
-        String universalServiceIdentifier =
-                args.get("universalServiceIdentifier") instanceof String
-                        ? (String) args.get("universalServiceIdentifier")
-                        : null;
-        if (universalServiceIdentifier == null) {
-            return;
-        }
+        // Let it fail if it is not a String
+        String universalServiceIdentifier = (String) args.get("universalServiceIdentifier");
 
         Set<Resource> resourcesToRemove = new HashSet<>();
         List<Reference> observationReferences = new ArrayList<>();
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemoveObservationRequests.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemoveObservationRequests.java
@@ -22,10 +22,13 @@ import org.hl7.fhir.r4.model.ServiceRequest;
 public class RemoveObservationRequests implements CustomFhirTransformation {
 
     @Override
-    public void transform(FhirResource<?> resource, Map<String, String> args) {
+    public void transform(FhirResource<?> resource, Map<String, Object> args) {
         Bundle bundle = (Bundle) resource.getUnderlyingResource();
+        String universalServiceIdentifier =
+                args.get("universalServiceIdentifier") instanceof String
+                        ? (String) args.get("universalServiceIdentifier")
+                        : "";
 
-        String universalServiceIdentifier = args.get("universalServiceIdentifier");
         Set<Resource> resourcesToRemove = new HashSet<>();
         List<Reference> observationReferences = new ArrayList<>();
         DiagnosticReport singleDiagnosticReport = null;
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemoveObservationRequests.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemoveObservationRequests.java
@@ -27,7 +27,7 @@ public class RemoveObservationRequests implements CustomFhirTransformation {
         String universalServiceIdentifier =
                 args.get("universalServiceIdentifier") instanceof String
                         ? (String) args.get("universalServiceIdentifier")
-                        : "";
+                        : null;
 
         Set<Resource> resourcesToRemove = new HashSet<>();
         List<Reference> observationReferences = new ArrayList<>();
--- a/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemoveObservationRequests.java
+++ b/etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemoveObservationRequests.java
@@ -28,6 +28,9 @@ public class RemoveObservationRequests implements CustomFhirTransformation {
                 args.get("universalServiceIdentifier") instanceof String
                         ? (String) args.get("universalServiceIdentifier")
                         : null;
+        if (universalServiceIdentifier == null) {
+            return;
+        }
 
         Set<Resource> resourcesToRemove = new HashSet<>();
         List<Reference> observationReferences = new ArrayList<>();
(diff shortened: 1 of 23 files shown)


etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemoveObservationRequests.java
```diff
@@ -22,10 +22,13 @@
 public class RemoveObservationRequests implements CustomFhirTransformation {
 
     @Override
-    public void transform(FhirResource<?> resource, Map<String, String> args) {
+    public void transform(FhirResource<?> resource, Map<String, Object> args) {
         Bundle bundle = (Bundle) resource.getUnderlyingResource();
+        String universalServiceIdentifier =
+                args.get("universalServiceIdentifier") instanceof String
+                        ? (String) args.get("universalServiceIdentifier")
+                        : "";
```
jorg3lopez, line 30: I'm not sure if we want this to pass with an empty string or do we want it fail with a nullPointerException so we know that the transformation was not successful? This will not stop the flow of the request.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/UpdateUniversalServiceIdentifier.java
```diff
@@ -23,33 +23,40 @@ public class UpdateUniversalServiceIdentifier implements CustomFhirTransformatio
...
+                    String alternateId = castToString(args.get(ALTERNATE_ID_NAME));
                     if (alternateId != null) {
                         updateAlternateCodingId(allCodings, alternateId);
                     }
                 });
     }
 
+    // Returns null if it is not a String
+    private String castToString(Object obj) {
+        return obj instanceof String ? (String) obj : null;
```
jorg3lopez, line 57: Again, do we want this to skip or pass an empty string ?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemoveObservationRequests.java
jorg3lopez: I ended up using null

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemoveObservationRequests.java
```diff
@@ -22,10 +22,16 @@
 public class RemoveObservationRequests implements CustomFhirTransformation {
 
     @Override
-    public void transform(FhirResource<?> resource, Map<String, String> args) {
+    public void transform(FhirResource<?> resource, Map<String, Object> args) {
         Bundle bundle = (Bundle) resource.getUnderlyingResource();
+        String universalServiceIdentifier =
+                args.get("universalServiceIdentifier") instanceof String
```
basiliskus, line 28: We control the value returned in args.get("universalServiceIdentifier"), so I think it's better to let it fail instead of returning null, as it would be a mistake to have a value that is not a string there

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/UpdateReceivingApplicationNamespace.java
```diff
@@ -13,11 +13,19 @@
 public class UpdateReceivingApplicationNamespace implements CustomFhirTransformation {
 
     @Override
-    public void transform(FhirResource<?> resource, Map<String, String> args) {
+    public void transform(FhirResource<?> resource, Map<String, Object> args) {
         Bundle bundle = (Bundle) resource.getUnderlyingResource();
+        String name = (args.get("name") instanceof String ? (String) args.get("name") : null);
         var receivingApplication = HapiHelper.getMSH5MessageDestinationComponent(bundle);
+        if (name == null) {
```
basiliskus, line 20: Same as my comment above. I don't think we want to return null when an expected value in the args is not there or doesn't have the right type

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/UpdateSendingFacilityNamespace.java
```diff
@@ -15,14 +16,20 @@
...
         Bundle bundle = (Bundle) resource.getUnderlyingResource();
         Identifier namespaceIdentifier = HapiHelper.getMSH4_1Identifier(bundle);
         if (namespaceIdentifier == null) {
             return;
         }
-        namespaceIdentifier.setValue(args.get("name"));
-        HapiHelper.getMSH4Organization(bundle)
+
+        String name = args.get("name") instanceof String ? (String) args.get("name") : null;
+        if (name == null) {
```
basiliskus, line 27: Same as my other comments

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/UpdateUniversalServiceIdentifier.java
basiliskus: Since we control the values passed in args, I think it's safe to assume they will be there and have the correct type. If that's not the case, we need to know so we can fix it, so it's better to let it fail

basiliskus (commented):
I left a few comments. It's actually just one comment but in multiple places. Otherwise looks good to me

jorg3lopez:
Since local code to LOINC or PLT code [PR ](https://github.com/CDCgov/trusted-intermediary/pull/1306)was merged, I had to refactor that transformation class (MapLocalObservationCodes) and it's test cases.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/RemoveObservationRequests.java
jorg3lopez: I agree. The exception will get caught in the RulesEngine when it's cycling through the rules and calling the translate method.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/UpdateReceivingApplicationNamespace.java
jorg3lopez: same

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/UpdateUniversalServiceIdentifier.java
```diff
@@ -23,33 +23,41 @@ public class UpdateUniversalServiceIdentifier implements CustomFhirTransformatio
...
-                    String alternateId = args.get(ALTERNATE_ID_NAME);
+                    String alternateId = castToString(args.get(ALTERNATE_ID_NAME));
                     if (alternateId != null) {
                         updateAlternateCodingId(allCodings, alternateId);
                     }
                 });
     }
 
+    // Returns null if it is not a String
+    private String castToString(Object obj) {
```
basiliskus, line 56: Since the method is just casting to String, it's probably better to remove it to avoid adding overhead

basiliskus (approved):
Left one suggestion, otherwise looks good!

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/UpdateUniversalServiceIdentifier.java
jorg3lopez: That's a good call.

PR #1383, opened 2024-10-04
Improve Slot Swapping

halprin:
# Improve Slot Swapping

TI is having issues with its slot swapping whenever deploying to any environment.  It fails a lot.  What's worse is that it takes 20+ minutes to fail.  RS SFTP Ingest doesn't have nearly the same failure rate.  It succeeds much more often.

One difference between RS SFTP Ingest and TI is us responding with an HTTP `200` status code from the `/` path.  So I tried adding this to TI, and that change is this PR.  I [tested this multiple times](https://github.com/CDCgov/trusted-intermediary/actions/runs/11181877789) in the Internal environment, and the tests always succeeded when previously they would easily fail.

Now, you may be wondering, why do we need to respond to `/` at all?  Our health check points to `/health`, for goodness sake.  And... I have no idea why.  Even with the health check pointing to `/health`, the health check seems to want to hit `/` anyway according to the app service's HTTP logs.  I looked for some other setting that would tell Azure "I mean it, actually use /health dang it", but I couldn't find it.  So instead I made `/` respond to health checks and it seems to work.

## Issue

_None_.

## Checklist

- [x] [Tested four times](https://github.com/CDCgov/trusted-intermediary/actions/runs/11181877789) in the Internal environment.

### DIFF
--- a/app/src/main/java/gov/hhs/cdc/trustedintermediary/external/javalin/App.java
+++ b/app/src/main/java/gov/hhs/cdc/trustedintermediary/external/javalin/App.java
@@ -38,7 +38,8 @@ import java.util.Set;
 public class App {
     private static final long MAX_REQUEST_SIZE = 2 * 1024 * 1024L; // 2 MB in bytes
     private static final int PORT = 8080;
-    static final String HEALTH_API_ENDPOINT = "/health";
+    private static final String HEALTH_API_ENDPOINT = "/health";
+    private static final String ROOT_API_ENDPOINT = "/";
 
     public static void main(String[] args) {
         var app =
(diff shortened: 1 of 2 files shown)


app/src/main/java/gov/hhs/cdc/trustedintermediary/external/javalin/App.java
```diff
@@ -50,6 +51,7 @@ public static void main(String[] args) {
 
         try {
             app.get(HEALTH_API_ENDPOINT, ctx -> ctx.result("Operational"));
+            app.get(ROOT_API_ENDPOINT, ctx -> ctx.result("Operational"));
```
jorg3lopez, line 54: Maybe as a future improvement, this  can serve as the landing endpoint for documentation, or provide other information like API version or build info. I know we already have the swagger endpoint, but it can be a possibility.

Issue #1387, opened 2024-10-04
Use `MSH-11` to identify test messages in staging and production environments


basiliskus:
# DevEx/OpEx

After an internal discussion, we decided `MSH-11` (ProcessingId) is the right candidate to identify two types of test messages: automated/scheduled and manually triggered. Following to the [HL7 specs](http://terminology.hl7.org/CodeSystem/v2-0103), we decided to use `T` (Training) for manual tests and `N` (Non-Production Testing) for automated tests.

## Tasks

- [x] Update TI test files to use either `T`, `D` or `N` in `MSH-11`
  - `T`: test files to be sent to partners and sent manually
  - `D`: test files not to be sent to partners and sent manually
  - `N`: test files not to be sent to partners and sent by scheduled task
- [x] Update RS org settings to route test messages
- [x] Make sure RS accepts the values we plan to use for MSH-11
- [x] Remove workaround to rewrite MSH in hurl script
- [x] Add documentation in readme
- [x] Add documentation in ADR
- [x] Update routing settings in staging
- [x] Ask Ott to update routing settings in prod

## Additional Context

Add any other context or screenshots about the work here.

basiliskus:
I'm waiting for feedback from SMEs to decide on the routing filters for `MSH-11` to use in production

basiliskus:
We'll discuss the `MSH-11` value for internal testing tomorrow in engineering block and I'm expecting to make a final decision there. After that, we'll ask RS to update the org settings in prod

basiliskus:
We've made a decision to use `D` for internal testing and `T` for partner testing. We have [this PR in RS](https://github.com/CDCgov/prime-reportstream/pull/16064) with the latest changes that is up for review. I've also asked Ott to update the settings in production

basiliskus:
The `flexion` org settings have been applied to production

PR #1389, opened 2024-10-07
Use MSH-11 to identify test messages

basiliskus:
# Use MSH-11 to identify test messages

- Updated `MSH-11` to `D` for all sample files in `examples/` (except files in `examples/Test/Automated/`)
- Updated `MSH-11` to `N` for all sample files in `examples/Test/Automated/`
- Removed workaround in `scripts/hurl/rs/hrl` to avoid unintentionally routing test messages to partners
- Added ADR and also documented in `examples/README.md`

## Issue

#1387


pluckyswan:
@CodiumAI-Agent /review

examples/CA/003_CA_ORU_R01_CDPH_produced_0_initial_message.hl7
```diff
@@ -1,4 +1,4 @@
-MSH|^~\&|SISGDSP|SISGDSP|SISHIERECEIVER^11903029^L,M,N|^^L,M,N|20240212103049||ORU^R01^ORU_R01|243408787|T|2.5.1
+MSH|^~\&|SISGDSP|SISGDSP|SISHIERECEIVER^11903029^L,M,N|^^L,M,N|20240212103049||ORU^R01^ORU_R01|243408787|D|2.5.1
```
somesylvie, line 1: Does this change need to be made in the `final` messages too? Like `003_CA_ORU_R01_CDPH_produced_3_hl7_translation_final.hl7` still has `T` in msh-11, and I assume the other `final` messages are like that

examples/README.md
```diff
@@ -15,6 +18,15 @@
   - `3_hl7_translation_final`
 - File extension: either `hl7` or `fhir`
 
+## Routing of these files
```
halprin, line 21: I like this section, but consider _also_ referencing and linking to the ADR.

examples/README.md
```diff
@@ -1,5 +1,8 @@
 # Sample message files
 
+> [!IMPORTANT]
+> Any new files added to the `examples/Test/` folder should be updated to have `D` in `MSH-11`. New files added to the `examples/Test/Automated/` folder should be updated to have `N` in `MSH-11`
```
somesylvie, line 4: All example files should have `D` or `N` in them right, not just the ones in `Test`? That matches the changes above

Also the hl7 files in `examples/Test` that _aren't_ in `examples/Test/Automated/` don't appear to have had their MSH-11s updated

halprin (commented):
This looks good to me, but I am curious about the other HL7 files that weren't updated (e.g. the `final` ones) that @somesylvie mentioned.

examples/Test/Orders/003_AL_ORM_O01_NBS_Fully_Populated_0_initial_message.hl7
```diff
@@ -1,4 +1,4 @@
-MSH|^~\&|BaptistOracle^2.16.840.1.114222.4.1.000000^ISO|BaptistEast^2.16.840.1.114222.4.1.000001^ISO|ALlabNatus^2.16.840.1.114222.4.1.181960.2^ISO|ALlab^simulated-lab-id^ISO|20240224134009||ORM^O01^ORM_O01|Q1960841872T2476960690||2.5.1||||||8859/1
+MSH|^~\&|BaptistOracle^2.16.840.1.114222.4.1.000000^ISO|BaptistEast^2.16.840.1.114222.4.1.000001^ISO|ALlabNatus^2.16.840.1.114222.4.1.181960.2^ISO|ALlab^simulated-lab-id^ISO|20240224134009||ORM^O01^ORM_O01|Q1960841872T2476960690|D|2.5.1||||||8859/1
```
somesylvie, line 1: should the corresponding final message for these also be updated?

examples/README.md
basiliskus: > All example files should have D or N in them right, not just the ones in Test? That matches the changes above

You are right. It should be for `examples/`

> Also the hl7 files in examples/Test that aren't in examples/Test/Automated/ don't appear to have had their MSH-11s updated

They should be updated. Can you point to which one is not?

basiliskus:
> This looks good to me, but I am curious about the other HL7 files that weren't updated (e.g. the `final` ones) that @somesylvie mentioned.

You can check my reply to @somesylvie with my reasoning

examples/Test/Orders/003_AL_ORM_O01_NBS_Fully_Populated_0_initial_message.hl7
basiliskus: See my other response

adr/026-hl7-test-message-routing.md
```diff
@@ -0,0 +1,43 @@
...
+
+After internal discussion, we decided to use `MSH-11` (Processing ID) for this purpose as it is a field already used in HL7 to identify test messages.
+
+We decided to use `D` (Debugging) for internal test messages and `T` (Training/Testing) for partner test messages because `T` is already used by our partners for testing, so we need to stick to that value.
+
+## Impact
+
+### Positive
+
+- We will be able to identify and route test messages in ReportStream without relying on fields that are overwritten by transformations; and as a result, avoid sending test messages to partners by mistake
```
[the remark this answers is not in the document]
basiliskus: Hmm good point. I'll add some clarification on the differences between staging and prod

examples/README.md
basiliskus: Added

examples/README.md
basiliskus: Fixed path

adr/026-hl7-test-message-routing.md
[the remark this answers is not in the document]
basiliskus: Added some more context

examples/CA/003_CA_ORU_R01_CDPH_produced_0_initial_message.hl7
somesylvie: I think it can be okay to have the `final` ones be like...a snapshot in time, as long as it's noted that they are. But I think if we don't update that value, they're not actually the right output for those inputs? Since I don't think MSH-11 changes through the workflow. It might be nice to either note when they were last generated, or have a way to regenerate them automatically. The nice part of having the files more static is there's fewer surprises, but having them more up to date makes them more valid for tests

examples/README.md
somesylvie, line 4: looks like the folder got corrected on line 23 but not here on line 4

basiliskus: Fixed

examples/CA/003_CA_ORU_R01_CDPH_produced_0_initial_message.hl7
basiliskus: I think regenerating them automatically is a good idea. It should be straightforward to create a bash script to do that using hurl. Though we'd need to manually get and overwrite the FHIR files from the local azure storage to update those

basiliskus: I'll create a new PR to add a script that will automate those updates as much as possible

saquino0827 (approved):

Issue #1386, opened 2024-10-04
🐞 - Intermittent permission failures retrieving secrets from KeyVault


basiliskus:
# Bug

## Describe the Bug

We're getting an intermittent exception in TI staging when RS sends a message to TI and TI tries to authenticate the token received with its public key. We don't know why sometimes it happens, and sometimes not.

Here's the stacktrace for one that happened on 2024-10-04T18:40:50.4613617Z UTC:
```
gov.hhs.cdc.trustedintermediary.wrappers.SecretRetrievalException: Not able to retrieve secret trusted-intermediary-public-key-stg from Azure
at gov.hhs.cdc.trustedintermediary.external.azure.AzureSecrets.getKey(AzureSecrets.java:50)
at gov.hhs.cdc.trustedintermediary.auth.AuthRequestValidator.retrievePublicKey(AuthRequestValidator.java:66)
at gov.hhs.cdc.trustedintermediary.auth.AuthRequestValidator.isValidAuthenticatedRequest(AuthRequestValidator.java:50)
at gov.hhs.cdc.trustedintermediary.external.javalin.DomainsRegistration.authenticateRequest(DomainsRegistration.java:158)
at gov.hhs.cdc.trustedintermediary.external.javalin.DomainsRegistration.processRequest(DomainsRegistration.java:140)
at gov.hhs.cdc.trustedintermediary.external.javalin.DomainsRegistration.lambda$createHandler$4(DomainsRegistration.java:128)
at io.javalin.router.Endpoint.handle(Endpoint.kt:52)
at io.javalin.router.ParsedEndpoint.handle(ParsedEndpoint.kt:15)
at io.javalin.http.servlet.DefaultTasks.HTTP$lambda$9$lambda$7$lambda$6(DefaultTasks.kt:52)
at io.javalin.http.servlet.JavalinServlet.handleTask(JavalinServlet.kt:99)
at io.javalin.http.servlet.JavalinServlet.handleSync(JavalinServlet.kt:64)
at io.javalin.http.servlet.JavalinServlet.handle(JavalinServlet.kt:50)
at io.javalin.http.servlet.JavalinServlet.service(JavalinServlet.kt:30)
at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:587)
at io.javalin.jetty.JavalinJettyServlet.service(JavalinJettyServlet.kt:52)
at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:587)
[...]

PR #1390, opened 2024-10-07
revert keyvault upgrade

somesylvie:
# Add a PR title
Reverting https://github.com/CDCgov/trusted-intermediary/pull/1369 - we started seeing errors retrieving secrets from keyvault after this change. 

## Issue
https://github.com/CDCgov/trusted-intermediary/issues/1386


somesylvie:
We no longer think this is related to the issue we're seeing

PR #1413, opened 2024-10-11
Scripts to submit messages to RS locally and update snapshots in /examples folder

basiliskus:
# Scripts to submit messages to RS locally and update snapshots in /examples folder

Added bash scripts to automate local testing. The scripts will 
- make sure the pre-requisites are installed (`hurl`, `jq` and `az`) and the TI and RS services are running
- detect the message type and use the right receiver
- with the submission id, query the RS history API to check the submission status and
  - get the inbound submission id
  - get the name of the first snapshot in azurite to download
  - copy the snapshot to the same folder where the original message is, adding the suffix `_1_hl7_translation`
- with the inbound submission id, query TI's metadata endpoint and
  - get the outbound submission id 
  - get the name for the next snapshot in azurite to download
  - copy the snapshot to the same folder where the original message is, adding the suffix `_2_fhir_transformation`
- with the outbound submission id, query the RS history API to check the status for the second leg of the message and
  - get the name for the final snapshot in azurite to download
  - copy the snapshot to the same folder where the original message is, adding the suffix `_3_hl7_translation_final`

Documentation on how to run the scripts here: `scripts/hurl/README.md`

## Issue

#1436

## Checklist


scripts/hurl/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+- [ReportStream](./rs/): scripts to send requests to ReportStream's endpoints
+- [CDC Intermediary](./ti/): scripts to send requests to the CDC Intermediary's endpoints
+- [Epic/UCSD](./epic/): scripts to send requests to Epic endpoints for UCSD
+
+## Local Submission Scripts
+
+- `submit_message.sh`: sends a HL7 message to a locally running RS instance. It also grabs the snapshots of the file in azurite after converting to FHIR, after applying transformations in TI, and after converting back to HL7. It copies these files to the same folder where the submitted file is
+    ```
+    Usage: ./submit_message.sh /path/to/message.hl7
```
halprin, line 20: This puts `Usage:` in monospaced font too.  Did you mean that?

halprin (approved):
Wow, that's some slick bash.

scripts/hurl/README.md
basiliskus: Yeah, it's the output of the `help` message, but now that you mention it, it looks weird because it doesn't have anything else aside from usage. I'll remove the `Usage:` part

Issue #1437, opened 2024-10-16
Vulnerability - Javalin jetty-server


jorg3lopez:
# Security Bug

## Javalin jetty-server 

Snyk notified us that the version of the jetty-server that comes with javalin version 6.3.0, has vulnerabilities. 

## Impact

This can impact the flow of data. For more information regarding the vulnerability, contact @jorg3lopez 

## Expected Behavior

To maintain all processes running from  the application.

## Screenshots

![Image](https://github.com/user-attachments/assets/54b4fb0f-c629-49c9-829f-95b09a3a4286)

PR #1440, opened 2024-10-16
Bug - Javalin Security Vulnerabilities

jorg3lopez:
# Javalin Vulnerabilities

Javalin uses Jetty-server and Snyk notified us of vulnerabilities related to the jetty-server verstion. The dependencies flagged were:
jetty-server and jetty-http

## Issue
#1437 

## Checklist

- [x] All tests pass

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -27,6 +27,7 @@ dependencies {
     // Import the Jetty BOM to pull consistent Jetty versions
     implementation 'org.eclipse.jetty:jetty-bom:12.0.14'
 
+    // Version 6.3.0 has jetty vulnerabilities. We are pulling the jetty directly to by pass this.
     implementation 'io.javalin:javalin:6.3.0'
 
     testImplementation 'org.apache.groovy:groovy:4.0.23'
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -23,6 +23,12 @@ dependencies {
     implementation project(':etor')
     testImplementation testFixtures(project(':shared'))
 
+    // Patch: Add the fixed Jetty version
+    implementation('org.eclipse.jetty:jetty-server:12.0.12') {
+        // Exclude the vulnerable version brought by Javalin
+        exclude group: 'org.eclipse.jetty', module: 'jetty-server'
+    }
+
     implementation 'io.javalin:javalin:6.3.0'
 
     testImplementation 'org.apache.groovy:groovy:4.0.23'
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -24,10 +24,8 @@ dependencies {
     testImplementation testFixtures(project(':shared'))
 
     // Patch: Add the fixed Jetty version
-    implementation('org.eclipse.jetty:jetty-server:12.0.12') {
-        // Exclude the vulnerable version brought by Javalin
-        exclude group: 'org.eclipse.jetty', module: 'jetty-server'
-    }
+    // Import the Jetty BOM to pull consistent Jetty versions
+    implementation 'org.eclipse.jetty:jetty-bom:12.0.14'
 
     implementation 'io.javalin:javalin:6.3.0'
 


app/build.gradle
```diff
@@ -23,6 +23,11 @@ dependencies {
     implementation project(':etor')
     testImplementation testFixtures(project(':shared'))
 
+    // Patch: Add the fixed Jetty version
+    // Import the Jetty BOM to pull consistent Jetty versions
+    implementation 'org.eclipse.jetty:jetty-bom:12.0.14'
```
jorg3lopez, line 28: Javalin pulls a bill of materials so I'm doing the same in order to have all dependencies working correctly

luis-pabon-tf (approved):

app/build.gradle
somesylvie: Just a contextual question since I'm not super familiar with Gradle - how do you test this? Like if you build locally, can you see what version of Jetty is getting used?

basiliskus: Can we remove `org.eclipse.jetty:jetty-bom:12.0.14` once javalin is updated with the patch? Or is the intention to leave it?

halprin (changes requested):
There are a couple of solutions to this problem.

1. Force the Jetty dependency to a newer version without the vulnerability.  This is what is basically happening in this PR.
2. Ignore this vulnerability in Snyk.  Say what?!

I propose we pursue solution 2, but I'm curious what others think.

For the rest of this review, I'll be using two terms: immediate dependency and transitive dependency.  In this specific case, our immediate dependency is Javalin and the transitive dependency is Jetty.  This is because we directly depend on Javalin (our immediate dependency), while Javalin subsequently depends on Jetty (our transitive dependency).

## 1. Force upgrade the transitive dependency

Again, this is what we're doing in this PR.

This is normally a valid solution when a vulnerability is disclosed and our immediate dependency hasn't upgraded yet.  This is true of our situation: Javalin hasn't released a new version that requires Jetty to be version 11.0.24 or newer.  Why 11.0.24?  Snyk tells us this vulnerability was fixed in 11.0.24.

When doing this, you need to check whether the immediate dependency works with the force upgraded transitive dependency.  That generally is checked by whether a compile works and by a battery of unit and integration tests passing.  You can also check the immediate dependency's documentation to see if it mentions anything about compatibility.

In this specific case, Javalin is currently on Jetty 11.  They have GitHub issues to explore updating to Jetty versions 12.  Because Javalin isn't already on Jetty 12, I suggest that this PR change to force Jetty to a non-vulnerable version of Jetty _11_.

Regardless, if we are to pursue this solution, we should put a DevEx/OpEx ticket into the backlog, and prioritize it for when Javalin upgrades to a non-vulnerable version of Jetty.  We do not want forced upgrade Gradle lines hanging about any longer than necessary because that can eventually lead to hard-to-debug bugs when Javalin gets out of sync with our forced Jetty version.

## 2. Ignore the vulnerability

This solution doesn't always apply.  The only reason this is acceptable in our case is because Javalin doesn't use the vulnerable class in question.  Snyk documents the vulnerable class as `ThreadLimitHandler`.  You can search in IntelliJ in a specific way to see that Javalin doesn't use it.  You can also [search Javalin's own repository](https://github.com/search?q=repo%3Ajavalin%2Fjavalin+ThreadLimitHandler&type=code) and see it isn't used.  Sometimes it isn't this easy to check to see whether our direct dependency is using the vulnerable aspect of the transitive dependency.

## Therefore

Therefore, I propose we just ignore this non-applicable vulnerability in Snyk, but I'm curious what others think.  There's less code changes, and we don't need to track a DevEx/OpEx ticket to revert this change.  I say ignore the vulnerability.

app/build.gradle
halprin: > Just a contextual question since I'm not super familiar with Gradle - how do you test this? Like if you build locally, can you see what version of Jetty is getting used?

You can run the following to see how Gradle resolved all the dependencies versions.
`./gradlew dependencies`

> Can we remove `org.eclipse.jetty:jetty-bom:12.0.14` once javalin is updated with the patch? Or is the intention to leave it?

Yes, normally you would remove it.

basiliskus:
> Therefore, I propose we just ignore this non-applicable vulnerability in Snyk, but I'm curious what others think. There's less code changes, and we don't need to track a DevEx/OpEx ticket to revert this change. I say ignore the vulnerability.

It makes sense to ignore the vulnerability if Javalin is not impacted by it. We'd avoid tracking another issue and having additional future work to revert this change

jorg3lopez:
Closing PR as we have decided to ignore the vulnerabilities at this time.

Issue #1452, opened 2024-10-17
CA Transform: Additional observation local code to LOINC


JohnNKing:
# Story

As UCSD, so that I can ingest an NBS result message, I need observations provided by CDPH to use codes that I support.

## Context

During round trip testing for CA, a scenario's result message failed to be processed by UCSD. We found this was caused by an OBX which, unexpectedly, didn't include a LOINC code. Only a local code was provided -- one not mapped as part of earlier efforts (#1277).

Draft mappings: https://docs.google.com/spreadsheets/d/1YwWB5Q7A68z1BTPEkcADq2Z_ZxA6xHcY/edit?gid=1378330894#gid=1378330894

## Acceptance Criteria
- [x] The appropriate LOINC code/description/code-system ("76030-6", "IDS gene full mutation analysis in Blood or Tissue by Sequencing", "LN")  is added to OBX-3.1/2/3 when local codes in OBX-3.4 is "99717-49"
- [x] The values of OBX-3.4/5/6 are retained
- [x] If there is no LOINC/PLT code mapped, the transformation should not happen (OBX-3 should be left as-is) 
- [x] A failed mapping lookup should be logged

###  Engineering
- [x] Add new mapping to transformation #1453
- [x] Refactor hardcoded mappings from transformation class to `transformation_definitions.json` #1453
- [x] Add RS E2E tests for the mapping transformation

## Definition of Done
- [x] Documentation tasks completed
  - [x] Documentation and diagrams created or updated
    - [x] ADRs (`/adr` folder)
    - [x] Main `README.md`
    - [x] Other READMEs in the repo
    - [x] If applicable, update the `ReportStream Setup` section in `README.md`
  - [x] [Threat model](https://lucid.app/lucidchart/8c6e8d37-2612-42a8-ac57-150c83f8e29e/edit) updated
  - [x] API documentation updated
- [x] Code quality tasks completed
  - [x] Code refactored for clarity and no design/technical debt
  - [x] Adhere to separation of concerns; code is not tightly coupled, especially to 3rd party dependencies
- [x] Testing tasks completed
  - [x] Load tests passed
  - [x] Additional e2e tests created
  - [x] Additional RS e2e assertions created in the `rs-e2e` project for any new transformations.  Includes improvements to the assertion code required to make the new assertions
- [x] Build & Deploy tasks completed
  - [x] Build process updated
  - [x] API(s) are versioned
[...]

PR #1453, opened 2024-10-17
Add 99717-49 Local Code to 76030-6 LOINC mapping

jbiskie:
# Adds 99717-49 Local Code to 76030-6 LOINC mapping

This PR:
- Adds a new local to LOINC mapping
- Refactors the hardcoded mapping in the transform class into a configurable arguments in `transformation_definitions.json`
- Adds additional unit tests to cover cases where the codingMap json argument is not structured as expected

## Issue

#1452 

This PR covers engineering tasks 1 and 2. Task 3 (Add RS E2E tests for the mapping transformation) will be addressed in another PR.

## Checklist


jbiskie:
> ## PR Code Suggestions ✨
> 
> Explore these optional code suggestions:
> Category	Suggestion                                                                                                                                    	Score
> **Possible issue** 	
> Prevent potential overwriting of existing keys in the map
> 
> 	7

Non-issue. IntelliJ flags duplicates.

halprin:
We should also update our RS E2E tests to test this.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/MapLocalObservationCodes.java
```diff
@@ -23,52 +23,68 @@
...
+        if (field == null || field.isBlank()) {
+            throw new IllegalArgumentException("missing or empty required field " + fieldName);
         }
+        return field;
     }
 
-    private void logUnmappedLocalCode(Bundle bundle, Coding coding) {
-        var msh41Identifier = HapiHelper.getMSH4_1Identifier(bundle);
-        var msh41Value = msh41Identifier != null ? msh41Identifier.getValue() : null;
+    private void logUnmappedLocalCode(Coding coding, String msh41Identifier, String messageId) {
```
basiliskus, line 81: Looks like this is only used once, it has a single line and the content is pretty straightforward. Does it need to extracted into its own function?

jbiskie:
Manual RS end to end testing was successful. Used file 020 and added an extra OBX for 99717-49:

`OBX|12|TX|^^^99717-49^New local code added^L|1|Negative|||N|||F|||20240711034913`

transformed:

`OBX|113|TX|76030-6^IDS gene full mutation analysis in Blood or Tissue by Sequencing^LN^99717-49^New local code added^L|1|Negative|||N|||F|||20240711034913`

basiliskus (commented):
Seems like there are some extracted functions that have a single line of code and are used only once. Are they needed? There's a tradeoff when refactoring to extract code to functions and adding too many can create unnecessary overhead. If they are added to help understand what the code does, you could consider adding a comment instead

luis-pabon-tf (approved):

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/ruleengine/transformation/custom/MapLocalObservationCodes.java
jorg3lopez: Good suggestion. I'll inline the extracting method and keep the logging one as is to keep the logging logic in one area.

basiliskus: What I meant is that we probably don't need `logUnmappedLocalCode` as it's adding more abstraction without the benefits. You could directly call `logger.logWarning` without putting it into its own function

jorg3lopez: Made the logging inline. Thanks for the suggestion, I think it makes sense.

basiliskus (approved):

Issue #1397, opened 2024-10-08
Azure Alerts for Errors - Azure is down


scleary1cs:
# Story
As a developer, I need to know if Azure is down, so that we can begin an incident.

## Acceptance Criteria
- [x] Alert is working in Slack

## Tasks
- [x] Create alert in TI alert.tf

## Definition of Done
- [x] Documentation tasks completed
  - [x] Documentation and diagrams created or updated
    - [x] ADRs (`/adr` folder)
    - [x] Main `README.md`
    - [x] Other READMEs in the repo
    - [x] If applicable, update the `ReportStream Setup` section in `README.md`
  - [x] [Threat model](https://lucid.app/lucidchart/8c6e8d37-2612-42a8-ac57-150c83f8e29e/edit) updated
  - [x] API documentation updated
- [x] Code quality tasks completed
  - [x] Code refactored for clarity and no design/technical debt
  - [x] Adhere to separation of concerns; code is not tightly coupled, especially to 3rd party dependencies
- [x] Testing tasks completed
  - [x] Load tests passed
  - [x] Additional e2e tests created
  - [x] Additional RS e2e assertions created in the `rs-e2e` project for any new transformations.  Includes improvements to the assertion code required to make the new assertions
- [x] Build & Deploy tasks completed
  - [x] Build process updated
  - [x] API(s) are versioned
  - [x] Feature toggles created and/or deleted.  Document the feature toggle
  - [x] Source code is merged to the main branch

## Notes
- This story came out of #1379, and #1380.

pluckyswan:
Created a Service Health > Health alerts in internal, currently disabled.

jherrflexion:
Curious if this would be too noisy if it is looking at every Azure service? Will attempt to test this and convert to Terraform today.

halprin:
> Curious if this would be too noisy if it is looking at every Azure service? Will attempt to test this and convert to Terraform today.

Yeah, we definitely don't want to be looking at services we don't use.  And yes, we want all these alert stories done via Terraform.

jherrflexion:
azure-outage-alert branch created. Internal clickops enabled for testing.

jherrflexion:
Currently blocked by an error in Terraform "CreateOrUpdate" on the test PR. Attempted to rerun the job a few times and draft another new PR and still received the error.

pluckyswan:
PR is out.

jherrflexion:
Addressing PR comments

[...]

PR #1464, opened 2024-10-21
Setup Azure Alert for Azure is down

jherrflexion:
# Setup Azure Alert for Azure is down

Added terraform for creating Azure Alert for when Azure is down. 

## Issue

https://github.com/CDCgov/trusted-intermediary/issues/1397

## Checklist

### DIFF
--- a/operations/template/alert.tf
+++ b/operations/template/alert.tf
@@ -31,7 +31,7 @@ resource "azurerm_monitor_action_group" "notify_slack_email" {
 resource "azurerm_monitor_activity_log_alert" "azure_service_health_alert" {
   count               = local.non_pr_environment ? 1 : 0
   name                = "cdcti-${var.environment}-azure-status-alert"
-  location            = data.azurerm_resource_group.group.location
+  location            = var.service_health_locations
   resource_group_name = data.azurerm_resource_group.group.name
   scopes              = [azurerm_container_registry.registry.id]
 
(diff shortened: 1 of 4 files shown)


operations/template/variables.tf
```diff
@@ -19,3 +19,8 @@ variable "alert_slack_email" {
   nullable  = false
   sensitive = true
 }
+
+variable "service_health_locations" {
+  type    = string
+  default = "global"
+}
```
halprin, line 26: Do we need this?  Do we actually customize this to anything?  I feel we could just hardcode `"global"` in the `azurerm_monitor_activity_log_alert`.

pluckyswan: Removed variable for now as its only used in this file. If more instances pop up, will refactor later.

operations/template/alert.tf
```diff
@@ -28,6 +28,47 @@ resource "azurerm_monitor_action_group" "notify_slack_email" {
   }
 }
 
+resource "azurerm_monitor_activity_log_alert" "azure_service_health_alert" {
+  count               = local.non_pr_environment ? 1 : 0
+  name                = "cdcti-${var.environment}-azure-status-alert"
+  location            = "global"
```
halprin, line 34: This should be `data.azurerm_resource_group.group.location`?  For every other resource that asks for a `resource_group_name` and `location`, we've put `data.azurerm_resource_group.group.location` for the `location`.

halprin (approved):
The System is Down

Issue #1396, opened 2024-10-08
Azure Alerts for Errors - 4xx status codes


scleary1cs:
# Story
As a developer, I need CDC-TI 4xx status codes to be alerted on after a threshold, so that we can begin an incident.

## Pre-conditions
- [ ] *Assumptions of prior or future work that's out of scope for this story*

## Acceptance Criteria
- [x] 4XX alerts are configured for all non-pr environments and have been tested in the TI application in a deployed environment
- [x] 4XX alerts are configured for all non-pr environments and have been tested in the RS SFTP ingestion service in a deployed environment

## Tasks
- [x] Set up 4xx Azure Alerts in the TI Application in all non PR environments
  - [x] Test 4xx azure alert in TI.
- [x] Set up 4xx Azure Alerts in the RS SFTP Ingestion Service in all non PR environments
  - [x] Test 4xx azure alert in TI.
  - [x] Make sure to remove **test400** endpoint from the **RS SFTP Ingestion Service**

## Definition of Done
- [x] Documentation tasks completed
  - [x] Documentation and diagrams created or updated
    - [x] ADRs (`/adr` folder)
    - [x] Main `README.md`
    - [x] Other READMEs in the repo
    - [x] If applicable, update the `ReportStream Setup` section in `README.md`
  - [x] [Threat model](https://lucid.app/lucidchart/8c6e8d37-2612-42a8-ac57-150c83f8e29e/edit) updated
  - [x] API documentation updated
- [x] Code quality tasks completed
  - [x] Code refactored for clarity and no design/technical debt
  - [x] Adhere to separation of concerns; code is not tightly coupled, especially to 3rd party dependencies
- [x] Testing tasks completed
  - [x] Load tests passed
  - [x] Additional e2e tests created
  - [x] Additional RS e2e assertions created in the `rs-e2e` project for any new transformations.  Includes improvements to the assertion code required to make the new assertions
  - [x] Tested alert triggers and resolutions in non production environments.
- [x] Build & Deploy tasks completed
  - [x] Build process updated
  - [x] API(s) are versioned
  - [x] Feature toggles created and/or deleted.  Document the feature toggle
  - [x] Source code is merged to the main branch

## Research Questions
- *Optional: Any initial questions for research*

## Decisions
- *Optional: Any decisions we've made while working on this story*

## Notes
- This story came out of #1379, and #1380.

saquino0827:
For this alert I believe we need the following:
[...]

PR #1465, opened 2024-10-21
Add 4xx Http Status Alerting

saquino0827:
# Describe

- Added an alert for http status code 4xx.
- Made the alert stateful with a frequency check of 1 minute and a look back period of 5 minutes.
  -  I am looking for opinions on frequency and look back periods.
  - There is a comment in the code where this is set and I will remove that comment when we have a decision on the timings.

## Issue

[1396](https://github.com/CDCgov/trusted-intermediary/issues/1396)

## Checklist


operations/template/alert.tf
```diff
@@ -61,21 +81,40 @@ resource "azurerm_monitor_scheduled_query_rules_alert" "database_token_expired_a
...
-      tags["funding_source"],
-      tags["pii_data"],
-      tags["security_compliance"],
-      tags["security_steward"],
-      tags["support_group"],
-      tags["system"],
-      tags["technical_steward"],
-      tags["zone"]
+      # Ignore changes to tags because the CDC sets these automagically
+      tags,
```
halprin, line 87: The reason the tags were enumerated separately was because we only wanted to ignore the automatically added tags from the CDC, not any tags added by us if we ever were to.

operations/template/alert.tf
```diff
@@ -61,21 +81,40 @@ resource "azurerm_monitor_scheduled_query_rules_alert" "database_token_expired_a
...
+
+resource "azurerm_monitor_metric_alert" "azure_4XX_alert" {
+  count               = local.non_pr_environment ? 1 : 0
+  name                = "cdcti-${var.environment}-azure-http-4XX-alert"
+  resource_group_name = data.azurerm_resource_group.group.name
+  scopes              = [data.azurerm_resource_group.group.id]
+  description         = "Action will be triggered when Http Status Code 4XX is greater than or equal to 1"
+  frequency           = "PT1M" // Checks every 1 minute
+  window_size         = "PT5M" // Every Check, looks back 5 minutes in history
+  //TBD: How frequent do we want this alert and how far do we want it to look back.
```
halprin, line 100: The real question to me is how many 4XXs in so many minutes do we want to initiate an incident?  Like, if one of us does a test in production, and we accidentally hit the wrong URL that generates a 404, is that something worth generating an incident over?  Or what about if ReportStream calls us, they have accidentally lost their secrets for us, and so they start getting 401s?

So, to answer the question, I'm not sure how big the `window_size` should be, but it is impacted by the `threshold` below too.  I feel `threshold` should be greater than `1` at least.

halprin: Thinking about this some more, perhaps the threshold should be 3 4XXs within an hour?  So `threshold` will be `3`, `frequency` can be `PT1M` (every minute) and `window_size` is `PT1H` (or `60M` if `1H` isn't valid?).  Thoughts?

operations/template/alert.tf
```diff
@@ -102,8 +103,49 @@ resource "azurerm_monitor_scheduled_query_rules_alert" "database_token_expired_a
...
+      tags["zone"]
+    ]
+  }
+}
+
+resource "azurerm_monitor_metric_alert" "azure_4XX_alert" {
+  count               = local.non_pr_environment ? 1 : 0
+  name                = "cdcti-${var.environment}-azure-http-4XX-alert"
+  resource_group_name = data.azurerm_resource_group.group.name
+  scopes              = [data.azurerm_resource_group.group.id]
```
halprin, line 129: We don't want the entire resource group checked, right?  Since this is in the TI repo, we want this specific to TI, so should this be perhaps `azurerm_linux_web_app.api.id`?

operations/template/alert.tf
saquino0827: See: https://github.com/CDCgov/trusted-intermediary/issues/1396#issuecomment-2429878961

We reverted back to the previous style of `ignore_changes`

operations/template/alert.tf
saquino0827: I'd have to check what are valid values but if I can I will set it to 1H.

saquino0827: This has been updated to the recommended.

operations/template/alert.tf
```diff
@@ -9,21 +9,41 @@
...
-      tags["escid"],
-      tags["funding_source"],
-      tags["pii_data"],
-      tags["security_compliance"],
-      tags["security_steward"],
-      tags["support_group"],
-      tags["system"],
-      tags["technical_steward"],
-      tags["zone"]
+      # Ignore changes to tags because the CDC sets these automagically
```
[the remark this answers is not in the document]
saquino0827: This is outdated due to https://github.com/CDCgov/trusted-intermediary/issues/1396#issuecomment-2429878961

operations/template/alert.tf
halprin: @saquino0827, I think all I'm waiting for is addressing this feedback (or having a conversation around this if you feel this isn't right).  After that, I'm ready to approve.

halprin: Oh, and also probably want to make sure this actually works in some environment.

saquino0827: @halprin this has been updated. Currently testing in internal and waiting for the alert that was fired to reach resolved status

halprin (approved):
Looks like it works.  Nice.

Issue #1394, opened 2024-10-08
Azure Alerts for Errors - 5xx status codes


scleary1cs:
# Story
As a developer, I need CDC-TI (TI & SFTP) 5xx status codes alerted on, so that we can begin an incident.

## Pre-conditions
- [ ] *Assumptions of prior or future work that's out of scope for this story*

## Acceptance Criteria
- [x] 5xx alerts are configured and have been tested for all non-pr env in TI.
- [x] 5xx alerts are configured and have been tested for all non-pr env in RS SFTP.

## Tasks
- [x] Azure Alert in TI
  - [x] Add terraform
  - [x] Test 
- [x] Azure Alert in RS SFTP
  - [x] Add terraform
  - [x] Test  

### Research
- [ ] *Research work needed to complete the story*
- [ ] Foundational: *High-level research that will support this and future efforts*

###  Engineering
- [ ] *Engineering work needed to complete the story*
- [ ] Foundational: *Technical runway work to support this and future efforts*

## Definition of Done
- [x] Documentation tasks completed
  - [x] Documentation and diagrams created or updated
    - [x] ADRs (`/adr` folder)
    - [x] Main `README.md`
    - [x] Other READMEs in the repo
    - [x] If applicable, update the `ReportStream Setup` section in `README.md`
  - [x] [Threat model](https://lucid.app/lucidchart/8c6e8d37-2612-42a8-ac57-150c83f8e29e/edit) updated
  - [x] API documentation updated
- [x] Code quality tasks completed
  - [x] Code refactored for clarity and no design/technical debt
  - [x] Adhere to separation of concerns; code is not tightly coupled, especially to 3rd party dependencies
  - [x] Code is reviewed or developed by pair; 1 approval is needed but consider requiring an outside-the-pair reviewer
  - [x] Code quality checks passed
- [x] Security & Privacy tasks completed
  - [x] Security & privacy gates passed
- [x] Testing tasks completed
  - [x] Load tests passed
  - [x] Unit test coverage of our code >= 90%
- [x] Build & Deploy tasks completed
  - [x] Build process updated
  - [x] API(s) are versioned
  - [x] Feature toggles created and/or deleted.  Document the feature toggle
  - [x] Source code is merged to the main branch

## Research Questions
- *Optional: Any initial questions for research*

## Decisions
- *Optional: Any decisions we've made while working on this story*

## Notes
- This story came out of #1379, and #1380.

scleary1cs:
[...]

PR #1466, opened 2024-10-21
5xx error alerts

pluckyswan:
# 5xx error alerts

Added terraform for 5xx error Azure alerts. Because it was deemed as higher priority than 4xx errors, the threshold is lower and more frequent. It was tested in internal by creating and hitting a temporary bad health check that returns a 500. Alert appeared in Slack when activated and deactivated. 

## Issue

https://github.com/CDCgov/trusted-intermediary/issues/1394

## Checklist


operations/template/alert.tf
```diff
@@ -61,21 +81,70 @@ resource "azurerm_monitor_scheduled_query_rules_alert" "database_token_expired_a
...
+
+resource "azurerm_monitor_metric_alert" "azure_5XX_alert" {
+  count               = local.non_pr_environment ? 1 : 0
+  name                = "cdcti-${var.environment}-azure-http-5XX-alert"
+  resource_group_name = data.azurerm_resource_group.group.name
+  scopes              = [data.azurerm_resource_group.group.id]
+  description         = "Action will be triggered when Http Status Code 5XX is greater than or equal to 1"
+  frequency           = "PT1M" // Checks every 1 minute
+  window_size         = "PT5M" // Every Check, looks back 5 minutes in history
+  //TBD: How frequent do we want this alert and how far do we want it to look back.
```
halprin, line 130: This makes sense to have a `threshold` of `1` and checking every minute.  And sure, why not `5` minutes ago even though we only last checked 1 minute ago.

halprin (commented):
I see these PRs build on each other.  I'd apply the feedback about the `tags` in the previous PR to this PR too.

operations/template/alert.tf
```diff
@@ -61,21 +81,70 @@ resource "azurerm_monitor_scheduled_query_rules_alert" "database_token_expired_a
...
+
+  criteria {
+    metric_namespace = "Microsoft.Web/sites"
+    metric_name      = "Http5xx"
```
pluckyswan, line 134: I don't think this should work since it's not a metric option but I'm unsure how to verify. Maybe use HttpServerErrors or a log alert instead if it doesn't.

halprin: Hmmm, we could inject a crash into our code so we have a guaranteed 500.  Then deploy to the internal environment.

halprin:
Since this PR seems to have been branched off of our 4xx alert branch, I would wait until #1465 gets merged before this is merged.

operations/template/alert.tf
```diff
@@ -120,6 +161,47 @@ resource "azurerm_monitor_scheduled_query_rules_alert" "database_token_expired_a
...
 }
+
+resource "azurerm_monitor_metric_alert" "azure_5XX_alert" {
+  count               = local.non_pr_environment ? 1 : 0
+  name                = "cdcti-${var.environment}-azure-http-5XX-alert"
+  resource_group_name = data.azurerm_resource_group.group.name
+  scopes              = [azurerm_linux_web_app.api.id]
+  description         = "Action will be triggered when Http Status Code 5XX is greater than or equal to 1"
+  frequency           = "PT1M" // Checks every 1 min
+  window_size         = "PT5M" // Every Check looks back 5 min for 4xx errors
```
halprin, line 172:
```suggestion
  window_size         = "PT5M" // Every Check looks back 5 min for 5xx errors
```

halprin: Besides this one thing, this alert looks good.

halprin (approved):
Kick ass!

Issue #1395, opened 2024-10-08
Azure Alerts for Errors - Log Errors


scleary1cs:
# Story
As a developer, I need CDC-TI (TI & SFTP) Log Errors alerted on (threshold?), so that we can begin an incident.

## Pre-conditions
- [ ] *Assumptions of prior or future work that's out of scope for this story*

## Acceptance Criteria
- [x] TI: Log Alerts for errors are set up and tested in a non production environment
- [x] SFTP: Log Alerts for errors are set up and tested in a non production environment

## Tasks
- [x] Create alert in TI
- [x] Create alert in SFTP

## Definition of Done
- [x] Documentation tasks completed
  - [x] Documentation and diagrams created or updated
    - [x] ADRs (`/adr` folder)
    - [x] Main `README.md`
    - [x] Other READMEs in the repo
    - [x] If applicable, update the `ReportStream Setup` section in `README.md`
  - [x] [Threat model](https://lucid.app/lucidchart/8c6e8d37-2612-42a8-ac57-150c83f8e29e/edit) updated
  - [x] API documentation updated
- [x] Code quality tasks completed
  - [x] Code refactored for clarity and no design/technical debt
  - [x] Adhere to separation of concerns; code is not tightly coupled, especially to 3rd party dependencies
- [x] Testing tasks completed
  - [x] Load tests passed
  - [x] Additional e2e tests created
  - [x] Additional RS e2e assertions created in the `rs-e2e` project for any new transformations.  Includes improvements to the assertion code required to make the new assertions
- [x] Build & Deploy tasks completed
  - [x] Build process updated
  - [x] API(s) are versioned
  - [x] Feature toggles created and/or deleted.  Document the feature toggle
  - [x] Source code is merged to the main branch

## Research Questions
- *Optional: Any initial questions for research*

## Decisions
- *Optional: Any decisions we've made while working on this story*

## Notes
- This story came out of #1379, and #1380.

jherrflexion:
Created branch: implement-and-configure-detailed-azure-log-alerts-with-comprehensive-monitoring-and-notifications-for-optimizing-azure-log-alerts-management-and-tracking-in-some-environments

jherrflexion:
Merged for TI. Waiting to see a successful alert somewhere.

halprin:
I think we needed to add one more additional check to the SFTP Ingestion Service log query to not alert for logs that come in from the pre-live slot.  I think Sylvie has the details on this.

[...]

PR #1469, opened 2024-10-22
Azure Log Errors Alert

jherrflexion:
## Issue

#1395

### DIFF
--- a/operations/template/alert.tf
+++ b/operations/template/alert.tf
@@ -129,7 +129,7 @@ resource "azurerm_monitor_scheduled_query_rules_alert" "ti-log-errors-alert" {
 
   action {
     action_group  = [azurerm_monitor_action_group.notify_slack_email[count.index].id]
-    email_subject = "TI log errors detected!"
+    email_subject = "${var.environment}: TI log errors detected!"
   }
 
   data_source_id = azurerm_linux_web_app.api.id
--- a/operations/template/alert.tf
+++ b/operations/template/alert.tf
@@ -120,3 +120,55 @@ resource "azurerm_monitor_scheduled_query_rules_alert" "database_token_expired_a
     ]
   }
 }
+
+resource "azurerm_monitor_scheduled_query_rules_alert" "ti-log-errors-alert" {
+  count               = local.non_pr_environment ? 1 : 0
+  name                = "cdcti-${var.environment}-log-errors-alert"
+  location            = data.azurerm_resource_group.group.location
+  resource_group_name = data.azurerm_resource_group.group.name
+
+  action {
+    action_group  = [azurerm_monitor_action_group.notify_slack_email[count.index].id]
+    email_subject = "TI log errors detected!"
+  }
+
+  data_source_id = azurerm_linux_web_app.api.id
+  description    = "Alert when total errors cross threshold"
+  enabled        = true
+
+  query = <<-QUERY
+      AppServiceConsoleLogs
+      | where Level == "Error"
+      and TimeGenerated >= ago(30m)
+      and TimeGenerated <= now()
+      | summarize count()
+    QUERY
+
+  severity                = 3
+  frequency               = 10
+  time_window             = 30
+  auto_mitigation_enabled = true
+
+  trigger {
+    operator  = "GreaterThan"
+    threshold = 1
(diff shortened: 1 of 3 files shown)


operations/template/alert.tf
```diff
@@ -120,3 +120,55 @@ resource "azurerm_monitor_scheduled_query_rules_alert" "database_token_expired_a
...
+  data_source_id = azurerm_linux_web_app.api.id
+  description    = "Alert when total errors cross threshold"
+  enabled        = true
+
+  query = <<-QUERY
+      AppServiceConsoleLogs
+      | where Level == "Error"
+      and TimeGenerated >= ago(30m)
+      and TimeGenerated <= now()
+      | summarize count()
```
halprin, line 144: I think the `Level` is not our level in this case.  Looking at `operations/template/logs.tf` and the `azurerm_log_analytics_query_pack_query` resources, we need to dig into the `ResultDescription`.  Taking inspiration from that, I think the query should be something like...

```suggestion
      AppServiceConsoleLogs
      | project JsonResult = parse_json(ResultDescription) | evaluate bag_unpack(JsonResult)
      | where level == 'ERROR'
      and @timestamp >= ago(30m)
      and @timestamp <= now()
      | summarize count()
```

Also, do we need the timestamp checks given that right below the query there's a `time_window`?

operations/template/alert.tf
```diff
@@ -120,3 +120,55 @@ resource "azurerm_monitor_scheduled_query_rules_alert" "database_token_expired_a
...
+    QUERY
+
+  severity                = 3
+  frequency               = 10
+  time_window             = 30
+  auto_mitigation_enabled = true
+
+  trigger {
+    operator  = "GreaterThan"
+    threshold = 1
```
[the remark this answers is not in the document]
halprin: Stop it, you crazy bot.  Not _everything_ needs to be parameterized.  lol

operations/template/alert.tf
```diff
@@ -120,3 +120,55 @@ resource "azurerm_monitor_scheduled_query_rules_alert" "database_token_expired_a
...
+
+resource "azurerm_monitor_scheduled_query_rules_alert" "ti-log-errors-alert" {
+  count               = local.non_pr_environment ? 1 : 0
+  name                = "cdcti-${var.environment}-log-errors-alert"
+  location            = data.azurerm_resource_group.group.location
+  resource_group_name = data.azurerm_resource_group.group.name
+
+  action {
+    action_group  = [azurerm_monitor_action_group.notify_slack_email[count.index].id]
+    email_subject = "TI log errors detected!"
```
halprin, line 132: I believe the e-mail subject is easily seen in Slack.  To help differentiate between production and any of the other environments, I recommend putting the environment in the subject line.  Maybe something like...?

```suggestion
    email_subject = "${var.environment}: TI log errors detected!"
```

operations/template/alert.tf
```diff
@@ -120,3 +120,55 @@ resource "azurerm_monitor_scheduled_query_rules_alert" "database_token_expired_a
...
+  }
+
+  data_source_id = azurerm_linux_web_app.api.id
+  description    = "Alert when total errors cross threshold"
+  enabled        = true
+
+  query = <<-QUERY
+      AppServiceConsoleLogs
+      | where Level == "Error"
```
basiliskus, line 141: The actual `Level` that you need to filter with is not this one. This one will always return `Informational` AFAIK. You can get the correct Level like this:
```
AppServiceConsoleLogs
| project JsonResult = parse_json(ResultDescription)
| evaluate bag_unpack(JsonResult)
| where level == 'ERROR'
```

operations/template/alert.tf
jherrflexion: Great question, what do you think @saquino0827 ?

jherrflexion: Imo, we need both. 

The timestamp is the range of the query and the time_window is how frequently it is ran. If anything, we could refactor out the 30 so there are no gaps between the range and frequency.

halprin (approved):
Assuming this works, I approve.

jherrflexion:
Looks like an internal merge broke this. Going to create a new one.

Issue #1398, opened 2024-10-08
Azure Alerts for Errors - CDC TI services are down/offline


scleary1cs:
# Story
As a developer, I need to know if our services are down/offline, so that we can begin an incident.

## Pre-conditions
- [ ] *Assumptions of prior or future work that's out of scope for this story*

## Acceptance Criteria
- [x] Alerts trigger when the TI and/or Ingestion Service are down. After discussion, we decided to base this on whether there are an appropriate number of healthy instances - if instances can't stay up/healthy, the service is not working

###  Engineering
- [x] Alerts are set up that trigger if TI is down
- [x] Alerts are set up that trigger if Ingestion Service is down

## Definition of Done
- [x] Documentation tasks completed
  - [x] Documentation and diagrams created or updated
    - [x] ADRs (`/adr` folder)
    - [x] Main `README.md`
    - [x] Other READMEs in the repo
    - [x] If applicable, update the `ReportStream Setup` section in `README.md`
  - [x] [TI Threat model](https://lucid.app/lucidchart/8c6e8d37-2612-42a8-ac57-150c83f8e29e/edit) or [Ingestion Service threat model](https://lucid.app/lucidchart/693ff5f9-c289-4585-9e74-a7f5abb36d69/edit) updated
  - [x] API documentation updated
- [x] Code quality tasks completed
  - [x] Code refactored for clarity and no design/technical debt
  - [x] Adhere to separation of concerns; code is not tightly coupled, especially to 3rd party dependencies
- [x] Testing tasks completed
  - [x] Load tests passed
  - [x] Additional e2e tests created
  - [x] Additional RS e2e assertions created in the `rs-e2e` project for any new transformations.  Includes improvements to the assertion code required to make the new assertions
- [x] Build & Deploy tasks completed
  - [x] Build process updated
  - [x] API(s) are versioned
  - [x] Feature toggles created and/or deleted.  Document the feature toggle
  - [x] Source code is merged to the main branch

## Research Questions
- *Optional: Any initial questions for research*

## Decisions
- *Optional: Any decisions we've made while working on this story*

## Notes
- This story came out of #1379, and #1380.

pluckyswan:
I believe this is covered by #1397

JohnNKing:
Already complete?

pluckyswan:
[...]

PR #1472, opened 2024-10-22
alert when instance count is low

somesylvie:
# Description
Alert when Azure instance count is low, fix logic for identifying non-PR environments

## Issue
https://github.com/CDCgov/trusted-intermediary/issues/1398

### DIFF
--- a/operations/template/alert.tf
+++ b/operations/template/alert.tf
@@ -126,6 +126,7 @@ resource "azurerm_monitor_metric_alert" "low_instance_count_alert" {
   resource_group_name = data.azurerm_resource_group.group.name
   scopes              = [azurerm_monitor_autoscale_setting.api_autoscale.id]
   description         = "Action will be triggered when the instance count is too low"
+  severity            = 2       // warning
   frequency           = "PT1M"  // Checks every 1 minute
   window_size         = "PT15M" // Every Check, looks back 15 minutes in history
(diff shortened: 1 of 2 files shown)


operations/template/alert.tf
```diff
@@ -120,3 +120,45 @@ resource "azurerm_monitor_scheduled_query_rules_alert" "database_token_expired_a
...
+  window_size         = "PT15M" // Every Check, looks back 15 minutes in history
+
+  criteria {
+    metric_namespace = "Microsoft.Insights/autoscalesettings"
+    metric_name      = "ObservedCapacity"
+    aggregation      = "Average"
+    operator         = "LessThanOrEqual"
+    // This threshold is based on the autoscale settings in app.tf
+    // How should we tune these numbers if we've scaled up higher than the initial count of 3/1?
+    threshold = local.higher_environment_level ? 2.5 : 0.5
```
halprin, line 140: We might be able to do something like this...

```suggestion
    threshold = azurerm_monitor_autoscale_setting.api_autoscale.profile.capacity.default - 0.5
```

Or we might need to do something like this...

```suggestion
    threshold = azurerm_monitor_autoscale_setting.api_autoscale.profile[0].capacity.default - 0.5
```

halprin (approved):

halprin (approved):
Re-approved.  Looks right.

Issue #1454, opened 2024-10-17
Leading zeros in OBX-5 for NM data type cause exception in RS when converting to FHIR


GilmoreA6:
# Bug

## Describe the Bug

When an OBX segment has a data type of "NM" in OBX-2 and the result value in OBX-5 contains leading zeros, the resulting fhir bundle contains an invalid value for "Bundle.entry.resource.ofType(Observation).valueQuantity.value".

example HL7 segment: 

OBX|1|NM|38473-5^17-Hydroxyprogesterone^LN|1|065.88|nmol/L|<85|N|||F|||20241015055726

example converted FHIR Resource:

`{
            "fullUrl": "Observation/1729104155244834642.39eac40b-c44b-4d07-b4bb-8b96ade62411",
            "resource": {
                "resourceType": "Observation",
                "id": "1729104155244834642.39eac40b-c44b-4d07-b4bb-8b96ade62411",
                "extension": [
                    {
                        "url": "https://reportstream.cdc.gov/fhir/StructureDefinition/sub-id",
                        "valueString": "1"
                    },
                    {
                        "url": "https://reportstream.cdc.gov/fhir/StructureDefinition/obx-observation",
                        "extension": [
                            {
                                "url": "OBX.2",
                                "valueId": "NM"
                            },
                            {
                                "url": "OBX.6",
                                "valueCodeableConcept": {
                                    "coding": [
                                        {
                                            "extension": [
                                                {
                                                    "url": "https://reportstream.cdc.gov/fhir/StructureDefinition/cwe-coding",
                                                    "valueString": "coding"
                                                }
                                            ],
                                            "code": "nmol/L"
                                        }
                                    ]
                                }
                            },
                            {
                                "url": "OBX.11",
                                "valueString": "F"
                            }
                        ]
                    }
                ],
                "status": "final",
                "code": {
                    "coding": [
                        {
                            "extension": [
                                {
                                    "url": "https://reportstream.cdc.gov/fhir/StructureDefinition/cwe-coding",
                                    "valueString": "coding"
                                },
                                {
                                    "url": "https://reportstream.cdc.gov/fhir/StructureDefinition/cwe-coding-system",
                                    "valueString": "LN"
                                }
                            ],
                            "system": "http://loinc.org",
                            "code": "38473-5",
                            "display": "17-Hydroxyprogesterone"
                        }
                    ]
                },
                "subject": {
[...]

PR #1487, opened 2024-10-24
added rs e2e test files and assertions for truncation logic

GilmoreA6:
# Add a PR title
Added two new files and relevant assertions to test leading zero truncation logic in RS e2e tests.

## Issue
https://github.com/CDCgov/trusted-intermediary/issues/1454

## Checklist


halprin (approved):
Looks kick-ass to me.  @basiliskus, mind taking a look?

GilmoreA6:
> Looks good! It's a different approach in terms of setting the conditions to test but it completely makes sense to me for this kind of use case

@basiliskus ya, I considered adding a new rule type that would check a regex and see if it matched and then we could apply it broadly but decided that could be a future enhancement and it made more sense to have specific test cases for right now

PR #1490, opened 2024-10-25
Update env variable name for automated test sender

basiliskus:
Now that we are using `simulated-sender` for the automated test task, updated to use the corresponding key stored in a new github env variable


somesylvie (approved):
Confirmed that this key exists and is spelled the same as in the PR, so it looks good!

Issue #1491, opened 2024-10-25
Split prod and non-prod alerts into separate slack channels


somesylvie:
# DevEx/OpEx

As an engineer so that I can rapidly identify prod issues and possible incidents, I would like prod and non-prod alerts to go to separate slack channels

## Tasks

- [x] Create new slack channel
- [x] Create/find email address for slack channel
- [x] Update Notion docs that reference alerts channels
- [x] Add team to slack channel
- [x] Add new email address as a secret to TI Github secrets
- [x] Update TI to pick right email based on environment - should just be a change in `prod-deploy.yml`
  - [x] Maybe also add a comment to the `azurerm_monitor_action_group`
- [x] Update TI alerts ADR
- [x] Create ingestion service alerts ADR
- [x] Create new non-prod secret
- [x] Update TF to reference new non-prod secret
- [x] Make a plan to delete the old secret: https://github.com/CDCgov/trusted-intermediary/issues/1516
- [x] Update VPN IP allocation because it's breaking things: https://github.com/CDCgov/trusted-intermediary/actions/runs/11525644212/job/32088318915

## Additional Context

Add any other context or screenshots about the work here.

PR #1499, opened 2024-10-28
Update VPN IP allocation

somesylvie:
# Description
Change to Standard SKU because basic stopped working, and static allocation because standard SKU requires it

We started seeing this error in all new envs: https://github.com/CDCgov/trusted-intermediary/actions/runs/11525644212/job/32088318915

Azure requires moving away from the Basic SKU next year anyway, so we'll have to make this change at some point: https://azure.microsoft.com/en-us/updates/upgrade-to-standard-sku-public-ip-addresses-in-azure-by-30-september-2025-basic-sku-will-be-retired/

## Issue
https://github.com/CDCgov/trusted-intermediary/issues/1491


halprin (approved):
Nice.

PR #1501, opened 2024-10-28
Added prelive filter to our error log alert

jherrflexion:
Added pre-live filtering to an alert

## Issue
#1395


halprin (approved):
Looks good to me.  Don't forget to do this also in SFTP Ingestion service.

PR #1510, opened 2024-10-28
Revert "Update VPN IP allocation"

somesylvie:
Reverts CDCgov/trusted-intermediary#1499

This is causing issues in already-existing environments


halprin (approved):

somesylvie:
The revert didn't work so we're going to fix the issue instead https://github.com/CDCgov/trusted-intermediary/pull/1507

PR #1513, opened 2024-10-29
Fix for Automated Test error

basiliskus:
Automated test workflow is failing: https://github.com/CDCgov/trusted-intermediary/actions/runs/11566385627

Fixed by removing extra commas that are making json parsing fail


GilmoreA6 (approved):
good catch

Issue #1517, opened 2024-10-30
Triage 4xx Errors in Staging


saquino0827:
# DevEx/OpEx

- Our 4xx alerts are being triggered. We suspect it's part of ReportStream's Healthcheck

## Tasks

- [x] Look into what is triggering our 4xx alerts in staging. - @halprin, @GilmoreA6, @JeremyIR 
- [x] Test fix in TI Internal. - @halprin, @GilmoreA6, @JeremyIR
- [x] Merge PR with final fix.

## Additional Context

Add any other context or screenshots about the work here.

PR #1519, opened 2024-10-30
Update 4XX Alert to Ignore RS Receiver Status Check

halprin:
# Description

Our 4XX alert was constantly triggering due to RS (and other things) calling non-existent endpoints.  Changed the alert to be log based and filtering out the pre-live slot, the RS receiver status check, and Microsoft pinging non-existent endpoints.

## Issue

#1517

## Checklist

- [x] Tested ignore works correctly in Internal.
- [x] Tested it activates correctly still in Internal.


GilmoreA6 (approved):
LGTM! testing in internal followed expectations,

PR #1522, opened 2024-11-01
Hurl scripts refactor

basiliskus:
- Renamed variables:
  - `client` => `sender`
  - `client_id` => `sender_org`
  - `client_sender` => `sender_name`
- Renamed `epic` => `ucsd`
- Refactored `rs.sh`, `ti.sh` and `ucsd.sh` scripts to reduce code duplication and for reusability

## Issue

#1566


scripts/hurl/epic/results.hurl
```diff
@@ -1,5 +1,5 @@
-# curl -H 'Content-Type: application/x-www-form-urlencoded' --data 'client_assertion=$(jwt encode --exp=+5min --jti $(uuidgen) --alg RS256 -k $client -i $client -s $client -a https://epicproxy-np.et0502.epichosted.com/FhirProxy/oauth2/token --no-iat -S @/path/to/ucsd-epic-private-key.pem)' --data 'client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-bearer' --data 'grant_type=client_credentials' 'https://epicproxy-np.et0502.epichosted.com/FhirProxy/oauth2/token'
-POST https://epicproxy-np.et0502.epichosted.com/FhirProxy/oauth2/token
+# curl -H 'Content-Type: application/x-www-form-urlencoded' --data 'client_assertion=$(jwt encode --exp=+5min --jti $(uuidgen) --alg RS256 -k $sender -i $sender -s $sender -a https://epicproxy-np.et0502.epichosted.com/FhirProxy/oauth2/token --no-iat -S @/path/to/ucsd-epic-private-key.pem)' --data 'client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-bearer' --data 'grant_type=client_credentials' 'https://epicproxy-np.et0502.epichosted.com/FhirProxy/oauth2/token'
```
pluckyswan, line 1: Does `client_assertions` need to be renamed too

pluckyswan: Same applies to any instances of client

basiliskus: No, those are HTTP header keys that won't change and refer to the client in the HTTP request

GilmoreA6 (changes requested):
Couple of overall comments, all of them nits. 

Right now there is a script called epic.sh where the description says "Submit requests to Epic API endpoints" but what it is really doing is sending a result to UCSD's staging Epic environment specifically. Since most hospitals in the US use Epic this could get confusing with more partners. I would say we should consider renaming this and updating the description to be more specific

Similarly we have a script called "submit.sh" and one called "rs.sh" that both send files to ReportStream with the difference being that submit.sh also does some more stuff. Maybe consider a more descriptive name for the submit script?

basiliskus:
> Right now there is a script called epic.sh where the description says "Submit requests to Epic API endpoints" but what it is really doing is sending a result to UCSD's staging Epic environment specifically. Since most hospitals in the US use Epic this could get confusing with more partners. I would say we should consider renaming this and updating the description to be more specific

Right, that script was made while testing connection to UCSD a few months ago. It's intended more as an example to test connections to partners in general. That's a good point though. I'll rename it

> Similarly we have a script called "submit.sh" and one called "rs.sh" that both send files to ReportStream with the difference being that submit.sh also does some more stuff. Maybe consider a more descriptive name for the submit script?

Yeah, submit.sh calls rs.sh and ti.sh. It's a higher level script. It submits a message to RS but its intention is not primarily to to do that, it's basically to run and track the whole ETOR flow. It actually had a longer name before but I shorten on purpose because that's a script I use all the time and didn't want to have it named something really long and descriptive. IMO, as long as the description of the script is clear in the Readme, it should be ok not to have a long descriptive name for it. To be able to set up and use the script you'd need to go through the documentation anyway, so that's a tradeoff that makes sense to me

GilmoreA6:
> eah, submit.sh calls rs.sh and ti.sh. It's a higher level script. It submits a message to RS but its intention is not primarily to to do that, it's basically to run and track the whole ETOR flow. It actually had a longer name before but I shorten on purpose because that's a script I use all the time and didn't want to have it named something really long and descriptive. IMO, as long as the description of the script is clear in the Readme, it should be ok not to have a long descriptive name for it. To be able to set up and use the script you'd need to go through the documentation anyway, so that's a tradeoff that makes sense to me

Cool, yeah if we are only ever expecting people to use submit.sh then I think it's fine the way it is

basiliskus:
@GilmoreA6 I renamed the epic scripts

GilmoreA6 (approved):

pluckyswan (approved):
LGTM!

Issue #1536, opened 2024-11-01
Make the automated test fail if any of the input or output files are not parsed correctly


basiliskus:
# DevEx/OpEx

This could just mean that in `HapiHL7FileMatcher.mapMessageByControlId` we need to bubble up the exception instead of just logging an error and continuing

## Tasks

- [x] Bubble up exceptions, add context, and let automated test fail
- [x] Throw exception if failed to find matching input or output
- [x] Add custom exception `HapiHL7FileMatcherException`

## Additional Context

https://flexion.slack.com/archives/C055XTF22B0/p1727717357268009

PR #1543, opened 2024-11-01
Avoid ignoring exceptions in automated tests

basiliskus:
# Avoid ignoring exceptions in automated tests

- Bubble up exceptions, add context, and let automated test fail
- Throw exception if failed to find matching input or output
- Refactored to simplify logic in `matchFiles`
- Added custom exception `HapiHL7FileMatcherException`

## Issue

#1536 

## Checklist


pluckyswan (approved):
Maybe add context on why we want this instead of logging in the description but besides that me likey

Issue #1528, opened 2024-11-01
Notify the engineering team when the `Automated Staging Test - Run integration tests` action fails


basiliskus:
# DevEx/OpEx

We need the engineering team to be notified when the `Automated Staging Test - Run integration tests` action fails. The workflow is in `.github/workflows/automated-staging-test-run.yml`

## Tasks

- [x] Look into adding people to notify using `CODEOWNERS` file
- [x] Look into adding a Slack Webhook to use in the workflow and notify the `cdcti-alerts-nonprod` slack channel
- [x] Loo into sending an email on test failure in the workflow
- [x] Decided to go with the Slack Webhook. Need to ask someone with Flexion Slack admin access to create it

## Additional Context

https://flexion.slack.com/archives/C055XTF22B0/p1728079599129939

PR #1585, opened 2024-11-14
Add workflow step to notify on Automated Test failure

basiliskus:
# Description

So the eng. team is notified when the `Automated Staging Test - Run integration tests` action fails, we're adding a Slack incoming webhook that will send a notification to the `cdcti-alerts-nonprod` slack channel on test failure. I've tested it and it's working

## Issue

#1528


pluckyswan (approved):
Nice

halprin (approved):
Nice dude!

Issue #1525, opened 2024-11-01
Evaluate whether all rules in `assertion_definitions.json` have been run in Automated Test


basiliskus:
# DevEx/OpEx

We may decide to log and error or throw an exception and make the test fail

## Tasks

- [x] Something.

## Additional Context

https://flexion.slack.com/archives/C055XTF22B0/p1727717357268009

PR #1586, opened 2024-11-14
Make sure all assertion rules run at least once

basiliskus:
# Description

Validate that all assertion rules in `assertion_definitions.json` are used at least once int the Automated Test

## Issue

#1525


rs-e2e/src/test/groovy/gov/hhs/cdc/trustedintermediary/rse2e/AutomatedTest.groovy
```diff
@@ -50,16 +52,19 @@ class AutomatedTest extends Specification  {
...
+        def toRunRules = engine.getRules()
         def matchedFiles = fileMatcher.matchFiles(azureFiles, localFiles)
 
         when:
         for (messagePair in matchedFiles) {
             Message inputMessage = messagePair.getKey() as Message
             Message outputMessage = messagePair.getValue() as Message
-            engine.runRules(outputMessage, inputMessage)
+            def runRules = engine.runRules(outputMessage, inputMessage)
+            toRunRules.removeAll(runRules)
```
somesylvie, line 63: I feel like the naming here might be confusing to come back to in the future since the `run` in `runRules` can either be a verb (like 'go run the rules') or an adjective ('run rules are rules that already ran'). Maybe they could be `evaluatedRules` and `rulesToEvaluate`, or `rulesThatRan` and `rulesToRun`, or some other pair that's slightly more distinct from the verb-based method names?

It might also be worth adding a small comment here, like `//Check whether all the rules in the assertions file have been run`

somesylvie (approved):
Just a small comment on naming clarity and a request for a comment, otherwise looks good

GilmoreA6 (approved):
LGTM! No notes.

rs-e2e/src/test/groovy/gov/hhs/cdc/trustedintermediary/rse2e/AutomatedTest.groovy
basiliskus: Good point, I'll change the name

Issue #824, opened 2024-01-30
Update Metadata Flow to use secure more Hash


jcrichlake:
# DevEx/OpEx

The metadata flow is currently using the built in Java hashing, but this can be reversed and isn't a secure way to have the message stored at rest.  Need to make a small code update to implement a secure, non-reversible hash

## Tasks

- [x] Update code to use a secure hash (SHA-512).
- [x] Add tests

## Additional Context

Add any other context or screenshots about the work here.

basiliskus:
Is this for the order hash? It was left like that on purpose because we don't have the original order hash and the one we're creating can't be use to verify the message integrity. So it's more of a placeholder for now until we can get the actual order has from RS (created from the original order coming in)

halprin:
Today, in our order and result usecase, we call `order.hashCode()` or `result.hashCode()`.  This is not a cryptographically secure hash.  Instead we should use SHA2-256, SHA2-512, or even better, a SHA3-256 or SHA3-512 hashing algorithms.

Unsure if there is built-in support to generate these hashes in Java.  If not, then we'll need to make a "humble object" that wraps the third-party library.

jcrichlake:
Looks like there is support in Java since 11

https://docs.oracle.com/en/java/javase/11/docs/specs/security/standard-names.html#messagedigest-algorithms

basiliskus:
I added a comment on Jan 31st about why it doesn't make sense to make this has code more cryptographically secure. How are we planning to use this hash? Is RS going to use it to verify the message coming from us?

PR #1594, opened 2024-11-19
Use SHA3-512

pluckyswan:
# Description

Generate a SHA3-512 hash for orders and results instead of using hashCode. Added unit tests for the new method generateHash. 

## Issue

https://github.com/CDCgov/trusted-intermediary/issues/824

## Checklist


etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/utils/security/HashHelper.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class HashHelper implements SecureHash {
+    @Inject Logger logger;
+
+    public <T> String generateHash(T input) {
+        try {
+            MessageDigest digest = MessageDigest.getInstance("SHA3-512");
+            byte[] objBytes = input.toString().getBytes(StandardCharsets.UTF_8);
+            byte[] hashBytes = digest.digest(objBytes);
+            return Hex.encodeHexString(hashBytes);
```
halprin, line 18: I believe you don't need a third-party library.  Try `HexFormat` which is built into Java.

`return HexFormat.of().formatHex(hashBytes);`

Caveat, I've never used `HexFormat` before.

pluckyswan: We swapped out the third-party library.

pluckyswan:
@halprin Two questions: 1. Is MessageDigest safe for multi-threading ([i](https://www.baeldung.com/sha-256-hashing-java#message-digest), [ii](https://stackoverflow.com/questions/17554998/need-thread-safe-messagedigest-in-java))? 2. There are instances of hashCode elsewhere in the repo that are overridden. Do these need to be swapped out too?

halprin:
> @halprin Two questions: 1. Is MessageDigest safe for multi-threading ([i](https://www.baeldung.com/sha-256-hashing-java#message-digest), [ii](https://stackoverflow.com/questions/17554998/need-thread-safe-messagedigest-in-java))? 2. There are instances of hashCode elsewhere in the repo that are overridden. Do these need to be swapped out too?

Correct, an instance of `MessageDigest` is not thread safe.  This means the object you get back from calling `MessageDigest.getInstance("whatever")` should not be used in multiple threads.  And you aren't using it in multiple threads.  Each call to `HashHelper#generateHash` could be on a separate thread but it creates its own instance of `MessageDigest`, so there is no shared instance of `MessageDigest` and we're all good.

Can you give me examples of what you're talking about?  Is it something like `HttpEndpoint.java` where we override `public int hashCode()`?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/utils/security/HashHelper.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.HexFormat;
+import javax.inject.Inject;
+
+public class HashHelper implements SecureHash {
+    @Inject Logger logger;
+
+    @Override
+    public <T> String generateHash(T input) {
+        try {
+            MessageDigest digest = MessageDigest.getInstance("SHA-512");
```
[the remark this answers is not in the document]
pluckyswan: When we had SHA3-512, Github's security bot got angry at us. I believe that SHA2 (which is what that is) is supposed to be more secure. @jcrichlake correct me if I'm wrong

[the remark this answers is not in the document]
pluckyswan: Screenshot 2024-11-21 at 9 16 03 AM

pluckyswan:
> > @halprin Two questions: 1. Is MessageDigest safe for multi-threading ([i](https://www.baeldung.com/sha-256-hashing-java#message-digest), [ii](https://stackoverflow.com/questions/17554998/need-thread-safe-messagedigest-in-java))? 2. There are instances of hashCode elsewhere in the repo that are overridden. Do these need to be swapped out too?
> 
> An instance of `MessageDigest` is not thread safe. This means the object you get back from calling `MessageDigest.getInstance("whatever")` should not be used in multiple threads. And you aren't using it in multiple threads. Each call to `HashHelper#generateHash` could be on a separate thread but it creates its own instance of `MessageDigest`, so there is no shared instance of `MessageDigest` and we're all good.
> 
> Can you give me examples of what you're talking about? Is it something like `HttpEndpoint.java` where we override `public int hashCode()`?

MessageLink and HttpEndpoint both have hashCode() that is overrode. Just wanted to double check it those don't need altering too.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/utils/security/HashHelper.java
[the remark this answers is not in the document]
halprin: Very weird.  And I see it highlighted it again.  I'm pretty dang sure that it's more than A-OK.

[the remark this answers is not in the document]
halprin: Looks like CodeQL has created a [fix for this](https://github.com/github/codeql/pull/18084).  I would use `SHA3-512`.

Thanks for making an issue for them to address!  Thanks to this work, you all are improving security for all open-source!

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/utils/security/HashHelper.java
```diff
@@ -0,0 +1,25 @@
+package gov.hhs.cdc.trustedintermediary.etor.utils.security;
+
+import gov.hhs.cdc.trustedintermediary.wrappers.Logger;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.HexFormat;
+import javax.inject.Inject;
+
+public class HashHelper implements SecureHash {
```
halprin, line 10: I think this application could be a singleton class or be purely static.  For example, `public static String generateHash(...)`.  I'll add more to this a bit later.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/utils/security/HashHelper.java
```diff
@@ -0,0 +1,25 @@
...
+    @Inject Logger logger;
+
+    @Override
+    public <T> String generateHash(T input) {
```
halprin, line 14: I don't think we need to use generics here.  I.e. the `T` stuff.  You could probably just use `Object` since we aren't really using the functionality of generics in this method.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/utils/security/HashHelper.java
halprin: And if you are going to make this a singleton like class (like most of our other classes, e.g. `ApacheClient` and its `getInstance()` method), you could even register this with the `ApplicationContext` and use `@Inject` to bring it into `SendOrderUsecase` (and result one too).  In fact, I think I recommend this.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/utils/security/HashHelper.java
pluckyswan: I'll defer to that

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/utils/security/SecureHash.java
```diff
@@ -0,0 +1,7 @@
+package gov.hhs.cdc.trustedintermediary.etor.utils.security;
+
+import java.security.NoSuchAlgorithmException;
+
+public interface SecureHash {
```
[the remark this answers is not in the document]
basiliskus: I agree with Peter. In general, I think it's better to avoid adding code preemptively as there's always a cost to adding more code (adds complexity, which increases potential for bugs, mental overhead, etc); and we can always refactor to add an interface if needed in the future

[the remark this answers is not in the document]
pluckyswan: That makes sense. I removed it

GilmoreA6 (approved):

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/utils/security/HashHelper.java
```diff
@@ -0,0 +1,26 @@
...
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.HexFormat;
+
+public class HashHelper {
+
+    private static final HashHelper INSTANCE = new HashHelper();
+
+    public static HashHelper getInstance() {
```
halprin, line 12: Mind making a follow-up PR to add a private constructor?  I'm happy to collaborate with you on that.

halprin (commented):
I know this has been merged, but let's make one more improvement, @pluckyswan.

Issue #1404, opened 2024-10-09
Rename sentSubmissionId  and receivedSubmissionId  to inboundMessageId  and outboundMessageId


basiliskus:
# DevEx/OpEx

High level description of what what needs to be improved.

## Tasks

- [x] Rename sentSubmissionId to outboundReportId
  - [x] Rename tests
  - [x] Rename variables
  - [x] Rename functions
- [x] Rename receivedSubmissionId to inboundReportId
  - [x] Rename tests
  - [x] Rename variables
  - [x] Rename functions
- [x] Add comment clarifying `recordId` is actual a reportId
- [x] Possibly add an ADR or clarifying comments on first instance of variables
- [x] Confirm cleanup

## Additional Context

https://flexion.slack.com/archives/C055XTF22B0/p1704911117020269

PR #1601, opened 2024-11-20
Rename reportId variables

pluckyswan:
# Description

Rename variables, like (sent/received)SubmissionId and (sent/received)MessageId to be (inbound/outbound)ReportId, and any related functions. The naming, inbound and outbound, is from the perspective of TI. The id that we were receiving from ReportStream is now the inbound id while the id we were sending to ReportStream is now the outbound id. 

## Issue

https://github.com/CDCgov/trusted-intermediary/issues/1404

## Checklist


etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messages/MessageRequestHandler.java
```diff
@@ -12,10 +12,10 @@ public interface MessageRequestHandler<T> {
     /**
      * Parses the request, converts and sends the message
      *
-     * @param receivedSubmissionId the ID for the submission returned from ReportStream
```
basiliskus, line 15: We could rephrase this for clarity. Something like: the report id created by ReportStream and sent to us in the request

pluckyswan: I used your suggestion.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messages/SendMessageHelper.java
```diff
@@ -21,67 +21,66 @@ public static SendMessageHelper getInstance() {
...
+                    inboundReportId, outboundReportId);
         } catch (PartnerMetadataException e) {
             logger.logError(
-                    "Unable to update metadata for received submissionId "
-                            + receivedSubmissionId
-                            + " and sent submissionId "
-                            + sentSubmissionId,
+                    "Unable to update metadata for inbound reportId "
+                            + inboundReportId
+                            + " and sent outbound reportId "
```
basiliskus, line 54: You can rename `sent outbound reportId` => `outboundReportId`

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messages/SendMessageHelper.java
```diff
@@ -21,67 +21,66 @@ public static SendMessageHelper getInstance() {
 
     private SendMessageHelper() {}
 
-    public void savePartnerMetadataForReceivedMessage(PartnerMetadata partnerMetadata) {
-        if (partnerMetadata.receivedSubmissionId() == null) {
+    public void savePartnerMetadataForInboundMessage(PartnerMetadata partnerMetadata) {
+        if (partnerMetadata.inboundReportId() == null) {
             logger.logWarning(
-                    "Received submissionId is null so not saving metadata for received message");
+                    "Inbound reportId is null so not saving metadata for received message");
```
basiliskus, line 27: There are few places with `Inbound reportId` or `Outbound reportId`. I think it's better to just rename them as `InboundReportId`/`OutboundReportId`, without the whitespace. That way it'd be easier search

basiliskus (commented):
I reviewed a few of the changes and added a couple of comments. I'll continue reviewing later

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messages/SendMessageHelper.java
pluckyswan: Good point. I swapped them out

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messages/SendMessageHelper.java
pluckyswan: True. Removed that

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/results/SendResultUseCase.java
```diff
@@ -41,15 +41,15 @@ public void convertAndSend(Result<?> result, String receivedSubmissionId)
...
                         result.getPlacerOrderNumber());
 
-        sendMessageHelper.savePartnerMetadataForReceivedMessage(partnerMetadata);
+        sendMessageHelper.savePartnerMetadataForInboundMessage(partnerMetadata);
 
         transformationEngine.runRules(result);
 
         String outboundReportId = sender.send(result).orElse(null);
-        logger.logInfo("Sent result reportId: {}", outboundReportId);
+        logger.logInfo("Outbound result outboundReportId: {}", outboundReportId);
```
basiliskus, line 49: I think we can keep "Sent result outboundReportId" here because it happens right after sending the result to RS

basiliskus (commented):
Left one comment. Otherwise looking great!

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/results/SendResultUseCase.java
pluckyswan: That makes sense

pluckyswan:
> Left one comment. Otherwise looking great!

Did we want to change `recordId` to `reportId`? Or is that a different card?

basiliskus:
> Did we want to change `recordId` to `reportId`? Or is that a different card?

That would be a different card. It involves changes in RS, so we need to coordinate to make those changes in TI and RS at the same time

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messages/SendMessageHelper.java
```diff
@@ -21,67 +21,66 @@ public static SendMessageHelper getInstance() {
...
-                            + partnerMetadata.receivedSubmissionId(),
+                    "Unable to save metadata for inboundReportId "
+                            + partnerMetadata.inboundReportId(),
                     e);
         }
     }
 
-    public void saveSentMessageSubmissionId(String receivedSubmissionId, String sentSubmissionId) {
-        if (sentSubmissionId == null || receivedSubmissionId == null) {
+    public void saveReportIds(String inboundReportId, String outboundReportId) {
```
halprin, line 40: I would rename this method as `saveOutboundReportId` because we are not saving the inbound report ID here.  Yes, we're passing in the inbound report ID, but that's only to get the correct metadata so we can save the sent report ID.

halprin (commented):
Wow, this is great work.  If you're going keep the modifications in `DatabasePartnerMetadataStorage.java`, then you'll need to also add a database migration in this PR, so that everything works when this PR is merged.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messages/SendMessageHelper.java
pluckyswan: In the function above this one, should the naming match this one `saveInboundReportId` or should the method you called out match `savePartnerMetadataForInboundMessage`

halprin: I'd keep it as is, `savePartnerMetadataForInboundMessage`, because it saves the entire partner metadata object.  This method that this thread of comments is on only updates the outbound report ID on an existing.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/DatabasePartnerMetadataStorage.java
```diff
@@ -27,7 +27,7 @@ public class DatabasePartnerMetadataStorage implements PartnerMetadataStorage {
     private static final DatabasePartnerMetadataStorage INSTANCE =
             new DatabasePartnerMetadataStorage();
 
-    private static final String METADATA_TABLE_RECEIVED_MESSAGE_ID = "received_message_id";
+    private static final String METADATA_TABLE_INBOUND_REPORT_ID = "inbound_report_id";
```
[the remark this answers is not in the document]
pluckyswan: Neither it is.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/DatabasePartnerMetadataStorage.java
```diff
@@ -163,8 +161,8 @@ PartnerMetadata partnerMetadataFromResultSet(ResultSet resultSet) {
             }
 
             return new PartnerMetadata(
-                    resultSet.getString(METADATA_TABLE_RECEIVED_MESSAGE_ID),
-                    resultSet.getString("sent_message_id"),
+                    resultSet.getString("inbound_report_id"),
```
halprin, line 164: You reverted some but not everything in this file.  This is still `"inbound_report_id"` and should be reverted to `"received_message_id"`.  There are also some send/outbound that haven't been reverted either.

I noticed this [conversation](https://github.com/CDCgov/trusted-intermediary/pull/1601#discussion_r1857181887) from last week.  I think the decision should go the other way with a constant for both.  This mix-up with only half of the columns being reverted would be harder to come by if there was just one constant value to change.

halprin: Granted, there are other strings in this file that didn't use the original `METADATA_TABLE_RECEIVED_MESSAGE_ID` but could have.  They also should be updated to use a constant because they could benefit from one of the positives of having a private constant in the a Java class: the ability to keep values in sync through a single constant.

pluckyswan: From my understanding, only the DB related variables or strings should be kept as the old naming convention as that has been kicked to another story regarding database migration. Anything non-DB related has the new naming convention. The reason the constant was removable was for readability purposes. I think it would be better to reintroduce that constant once the DB changes are made, if we do choose to have it again.

halprin: Hmmm, let's look at this together sometime today and chat.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/reportstream/ReportStreamEndpointClient.java
```diff
@@ -109,14 +109,14 @@ public String requestWatersEndpoint(String body, String bearerToken)
     }
 
     @Override
-    public String requestHistoryEndpoint(String submissionId, String bearerToken)
+    public String requestHistoryEndpoint(String outboundReportId, String bearerToken)
```
basiliskus, line 112: Looks good. I think you missed the `reportId` in `requestDeliveryEndpoint` below to be renamed to `inboundReportId`

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/DatabasePartnerMetadataStorage.java
basiliskus: My understanding is these names should not be updated, as they are column names in the DB

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/messages/SendMessageHelper.java
```diff
@@ -21,67 +21,66 @@ public static SendMessageHelper getInstance() {
...
         }
     }
 
-    public void saveSentMessageSubmissionId(String receivedSubmissionId, String sentSubmissionId) {
-        if (sentSubmissionId == null || receivedSubmissionId == null) {
+    public void saveOutboundReportId(String inboundReportId, String outboundReportId) {
+        if (outboundReportId == null || inboundReportId == null) {
             logger.logWarning(
-                    "Received and/or sent submissionId is null so not saving metadata for sent result");
+                    "Inbound and/or outboundReportId is null so not saving metadata for sent result");
```
basiliskus, line 43: I think it's better to use the full variable name here: `inboundReportId`

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/DatabasePartnerMetadataStorage.java
```diff
@@ -28,6 +28,7 @@ public class DatabasePartnerMetadataStorage implements PartnerMetadataStorage {
             new DatabasePartnerMetadataStorage();
 
     private static final String METADATA_TABLE_RECEIVED_MESSAGE_ID = "received_message_id";
+    private static final String METADATA_TABLE_SENT_MESSAGE_ID = "sent_message_id";
```
basiliskus, line 31: I know this has already been discussed, so I'm sorry to bring it up again, but I think it's better not to use a constant for query column names. It makes the code less readable (we need go to the variable instantiation to know the column name) and I don't think we have the benefits. The name of the variable itself has the name of the column, so it kind of defeats the purpose of using a constant because if in the future we have to change the name of the column, we'll also have to change the name of the variable

basiliskus (commented):
Great job @pluckyswan! I just left one comment but otherwise looks great

etor/src/main/java/gov/hhs/cdc/trustedintermediary/external/database/DatabasePartnerMetadataStorage.java
halprin: I don't think the we can really improve these variable names.  I'm not sure I would make them more generic.  Maybe?

It is true, this makes the SQL statements harder to read.

I think the constant solution is slightly better than having to find/replace the strings.  With the constant solution, you can rename the string once and use IDE tools to rename the constant which are more foolproof than find/replace.  _But_, it isn't that much better.

So, I feel this falls within the realm of code organization opinion.  Therefore, I say follow your heart.  Use a constant; don't use a constant.  ¯\_(ツ)_/¯

pluckyswan: I couldn't think of a more suitable constant name that wouldn't require being changed if the value did. I prefer having the readability of the SQL statements. I think having caution when using find/replace is worth the readability.

halprin (approved):
I approve.  I would get at least one other approval, though.

basiliskus (approved):

tjohnson7021 (approved):
This great and very helpful!

Issue #1361, opened 2024-09-25
@Inject for Non-Singleton Classes


jorg3lopez:
## Situation:
At the moment, we use the @Inject annotation for only singleton classes. This is because the ApplicationContext class only registers singleton classes. These classes are later scanned and have the instances injected, of the objects that use the @Inject annotation.

## Proposed Solution:
Have the non-singleton classes be able to use the @Inject annotation to inject the instances of classes that have been registered into the ApplicationContext.

### Reference:
https://flexion.slack.com/archives/C055XTF22B0/p1688582573819939

### Food for Thought:
Create a method in the ApplicationContext that will take in the non-singleton object that has the @Inject annotations. This should be part of the constructor of the non-singleton objecct. It should call the method in the ApplicationContext and pass itself as the parameter.
The ApplicationContext will scan this class and inject the instances for the objects using @Inject.
This is an example of how it can be accomplished.

jorg3lopez:
Draft [RP ](https://github.com/CDCgov/trusted-intermediary/pull/1617)created

JohnNKing:
Status: PR pending review: https://github.com/CDCgov/trusted-intermediary/pull/1617

PR #1617, opened 2024-11-26
Use inject for non singleton class

jorg3lopez:
# Description

This PR adds the logic necessary for non-singleton classes to use the @Inject annotation. 

# Changes
 `ApplicationContext.java`
  - Refactoring of `skipMissingImplementation` logic
  - Renamed `injectRegisteredImplementations` -> `doInjectRegisteredImplementations`
  - Overload `injectIntoField` method
  - Added `injectIntoNonSingleton()` method
  
 `Reflections.java`
  - Added `getFieldsAnnotatedWithInstance`
  
  `ApplicationContextText.groovy`
  - Added test cases
  
  `TestApplicationContext.java`
  - Added `skipMissingImplementation` logic
  
  `HapiFhirResource` 
  - The changes in this file were used for testing, they will get deleted once the PR is approved
  
  `HapiFhirResourceTest`
  - This file was used for testing, it will get deleted once the PR is approved

# Testing
- Choose a non-singleton class to work with. Make sure the class is not dynamically instantiated
- Use @Inject on fields that you want to inject the dependency
- Make sure the class sends itself as a parameter to`ApplicationContext.injectIntoNonSingleton()`, using the constructor
- Build module and make sure all tests pass

## Issue
#1361

## Checklist


shared/src/testFixtures/groovy/gov/hhs/cdc/trustedintermediary/context/TestApplicationContext.groovy
```diff
@@ -19,8 +19,9 @@ class TestApplicationContext extends ApplicationContext {
         TEST_ENV_VARS.clear()
     }
 
-    def static injectRegisteredImplementations() {
-        injectRegisteredImplementations(true)
+    def static injectRegisteredImplementations(def skip = true) {
```
halprin, line 22: We'd never need to set `skip` to `false` in a testing context.  I think we can remove the argument from this method.

The skipping allows us to only need to register implementations that are in use in each specific test.  If we don't allow skipping, each test needs to register every `@Inject` field across the entire application.  For example, this means that a test would need to register an `HttpClient` implementation even if the test isn't anywhere close to the HTTP code.

shared/src/testFixtures/groovy/gov/hhs/cdc/trustedintermediary/context/TestApplicationContext.groovy
```diff
@@ -19,8 +19,9 @@ class TestApplicationContext extends ApplicationContext {
...
+        skipMissingImplementations = skip
```
halprin, line 23: I like this.  But I also have a faint recollection that there's something wrong with this approach.  I'll see if I can't bang my head against the wall enough to remember what it was.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/context/ApplicationContext.java
```diff
@@ -76,15 +101,15 @@ private static void injectIntoField(Field field, boolean skipMissingImplementati
...
         if (fieldImplementation == null) {
             return;
         }
 
         Object declaringClassImplementation =
-                getDeclaringClassImplementation(declaringClassesToTry, skipMissingImplementations);
+                getDeclaringClassImplementation(declaringClassesToTry);
         if (declaringClassImplementation == null) {
-            return;
+            System.err.println("No class implementation found");
```
halprin, line 112: We already have an log statement printed on line 165 of this file.  And if that log statement is being printed, this log statement is being printed.  We don't need both log statements.

If we're going to keep this if statement, I would prefer we keep the `return` so we don't throw an unnecessary `NullPointerException`.  Thoughts?

shared/src/main/java/gov/hhs/cdc/trustedintermediary/context/ApplicationContext.java
```diff
@@ -53,16 +55,39 @@ public static <T> Set<Class<? extends T>> getImplementors(Class<T> interfaze) {
...
+        fields.forEach(ApplicationContext::injectIntoField);
+    }
+
+    public static void injectIntoNonSingleton(Object instance) {
+        var fields = Reflection.getFieldsAnnotatedWithInstance(instance.getClass(), Inject.class);
+
+        fields.forEach(field -> injectIntoField(field, instance));
+    }
+
+    private static void injectIntoField(Field field, Object instance) {
```
halprin, line 73: There is a lot of code duplication between this `injectIntoField` method and the original `injectIntoField` method.  I bet the original method could be refactored to call the new one.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/external/hapi/HapiFhirResource.java
```diff
@@ -1,19 +1,25 @@
...
     private final IBaseResource innerResource;
 
     public HapiFhirResource(IBaseResource innerResource) {
+        ApplicationContext.injectIntoNonSingleton(this);
         this.innerResource = innerResource;
     }
 
     @Override
     public IBaseResource getUnderlyingData() {
+        this.logger.logDebug("testing @Inject");
```
halprin, line 22: This file can now be reverted since it just has test code.

shared/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/hapi/HapiFhirResourceTest.groovy
```diff
@@ -0,0 +1,26 @@
+package gov.hhs.cdc.trustedintermediary.external.hapi
+
+import gov.hhs.cdc.trustedintermediary.context.TestApplicationContext
+import gov.hhs.cdc.trustedintermediary.wrappers.Logger
+import org.hl7.fhir.r4.model.Bundle
+import spock.lang.Specification
+
+class HapiFhirResourceTest extends Specification {
```
halprin, line 8: I guess this test file should be deleted since this is testing test code.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/context/ApplicationContext.java
jorg3lopez: I'm ok with that.

shared/src/test/groovy/gov/hhs/cdc/trustedintermediary/external/hapi/HapiFhirResourceTest.groovy
jorg3lopez: Good catch. I will delete this file once the pr is approved.

shared/src/main/java/gov/hhs/cdc/trustedintermediary/context/ApplicationContext.java
jorg3lopez: Good catch. I will refactor it.

shared/src/test/groovy/gov/hhs/cdc/trustedintermediary/context/ApplicationContextTest.groovy
```diff
@@ -210,17 +210,14 @@ class ApplicationContextTest extends Specification {
         def injectionInstantiation = new InjectionDeclaringClass()
 
         TestApplicationContext.register(List.class, injectionInstantiation)
-        // notice above that I'm registering the injectionInstantiation object as a List class.
-        // injectionInstantiation is of class InjectionDeclaringClass,
-        // and InjectionDeclaringClass doesn't implement List (it only implements AFieldInterface).
```
halprin: Do we no longer want these comments to point out the (on purpose) inconsistency?

halprin (approved):

shared/src/test/groovy/gov/hhs/cdc/trustedintermediary/context/ApplicationContextTest.groovy
jorg3lopez: I'll add it back so it can provide the description

PR #1639, opened 2024-12-03
Application context fix unreachable code

luis-pabon-tf:
# Description

This PR is refactoring work for issue #1361, to be merged into a different branch for the same issue.

### DIFF
--- a/shared/src/main/java/gov/hhs/cdc/trustedintermediary/context/ApplicationContext.java
+++ b/shared/src/main/java/gov/hhs/cdc/trustedintermediary/context/ApplicationContext.java
@@ -108,9 +108,6 @@ public class ApplicationContext {
 
         Object declaringClassImplementation =
                 getDeclaringClassImplementation(declaringClassesToTry);
-        if (declaringClassImplementation == null) {
-            return;
-        }
 
         field.trySetAccessible();
 
--- a/shared/src/test/groovy/gov/hhs/cdc/trustedintermediary/context/ApplicationContextTest.groovy
+++ b/shared/src/test/groovy/gov/hhs/cdc/trustedintermediary/context/ApplicationContextTest.groovy
@@ -1,5 +1,7 @@
 package gov.hhs.cdc.trustedintermediary.context
 
+import gov.hhs.cdc.trustedintermediary.ruleengine.RuleEngine
+import gov.hhs.cdc.trustedintermediary.wrappers.Logger
 import spock.lang.Specification
 
 import javax.inject.Inject
@@ -10,6 +12,12 @@ class ApplicationContextTest extends Specification {
 
     interface TestingInterface {
         void test()
+    }
+
+    class NonSingletonClazz {
+        @Inject
+        Logger logger
+        void test() {}
     }
 
     static class DogCow implements TestingInterface {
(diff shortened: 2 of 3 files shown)


shared/src/main/java/gov/hhs/cdc/trustedintermediary/context/ApplicationContext.java
```diff
@@ -108,9 +108,6 @@ private static void injectIntoField(Field field) {
 
         Object declaringClassImplementation =
                 getDeclaringClassImplementation(declaringClassesToTry);
-        if (declaringClassImplementation == null) {
-            return;
-        }
```
halprin, line 113: Am I reviewing this too soon?

`declaringClassImplementation` can be `null`.  Looking at the `getDeclaringClassImplementation` method, it   can return `null` when `declaringClassImplementation == null` and `skipMissingImplementations == true`.

So, I would keep this if statement.

luis-pabon-tf: I can revert this change, though I really could not find a way it would ever hit. This is the word salad I wrote on my notes while trying to make it happen: 
```
* getImplementation only throws an exception if the expected clazz is not in OBJECT_MAP

* getDeclaringClassImplementation only marks an implementation as null if getImplementation throws an exception

* injectIntoField only reaches getDeclaringClassImplementation if IMPLEMENTATIONS contains the declaring class for the field

* IMPLEMENTATIONS and OBJECT_MAP are both filled at the same time, when a class implementation is registered
```

jorg3lopez: Good catch. After looking into the code with @luis-pabon-tf, we agreed that this if/else statement could never be null. This is because there is another check before, that checks if the implementing class is in the IMPLEMENTATIONS map. If it is not, then it returns.

image

In addition, all the classes will have an interface that they are implementing. If they don't, then wouldn't go through this method, they would go through the non-singleton flow.

halprin: OK, I took another look at this.  There is a very small corner case for that if statement.  Without the if statement, the following `field.set(declaringClassImplementation, fieldImplementation);` code will throw a `NullPointerException`.

This corner case is if you do something like the following in a unit test...

```groovy
class ApplicationContextTest extends Specification {

    interface AFieldInterface {
        Object getAField()
    }

    class InjectionDeclaringClass implements AFieldInterface {
        @Inject
        private String aField

        def getAField() {
            return aField
        }
    }

    def setup() {
        TestApplicationContext.reset()
    }

    def "test the if statement"() {
        given:

        def injectedValue = "DogCow"
        def injectionInstantiation = new InjectionDeclaringClass()

        TestApplicationContext.register(List.class, injectionInstantiation)
        // notice above that I'm registering the injectionInstantiation object as a List class.
        // injectionInstantiation is of class InjectionDeclaringClass, and InjectionDeclaringClass doesn't implement List (it only implements AFieldInterface).
        TestApplicationContext.register(String.class, injectedValue)

        TestApplicationContext.injectRegisteredImplementations()

        when:
        def aFieldValue = injectionInstantiation.getAField()

        then:
        injectedValue == aFieldValue
    }
}
```

Run that unit test and the test fails with a thrown `NullPointerException` from the `TestApplicationContext.injectRegisteredImplementations()` line if the if statement is removed.  If the if statement was to remain, an exception wouldn't be thrown, and instead the test fails from the `injectedValue == aFieldValue` assertion failing.

So, basically to run into this corner case, you need to incorrectly register some implementation as some completely unrelated class.  E.g. the `injectionInstantiation` as `List` instead of `InjectionDeclaringClass` or `AFieldInterface`.

I'm comfortable with allowing the removal of the if statement because any unit tests will fail regardless and you have to really mess up registering of implementations.

luis-pabon-tf: Thanks for hunting down how to trigger that statement. We'd been having a lot of trouble doing so. I am adding your test case example with just a few tweaks into the file.

PR #1648, opened 2024-12-09
TI Script Updates

saquino0827:
# Description

- Updates the README for `script/README`
- Updates the set up scripts to choose a `gradle` vs `docker` setup for the reportstream setup script

[#1658](https://github.com/CDCgov/trusted-intermediary/issues/1658)


scripts/README.md
```diff
@@ -1,8 +1,34 @@
...
+    - [Requirements](#requirements-5)
+    - [Usage](#usage-5)
+  - [lib/common.sh](#libcommonsh)
+  - [lib/submission-utils.sh](#libsubmission-utilssh)
+- [Resources](#resources)
 
 ## Setup
-
-Follow the instructions below to load the environments variables required for these scripts
+<details open>
```
saquino0827, line 29: This **detail** tag is a collapsable section. I wanted to see how others feel about this so I am including it as a draft.

basiliskus: Not sure about this. I think the collapsible section feature could be useful for longer readmes where we have sections that we want to hide/collapse by default, but I don't see why we'd need it here. Specially if it's expanded by default

scripts/README.md
```diff
@@ -18,8 +44,9 @@ Follow the instructions below to load the environments variables required for th
    ```
    **Note**: you may also want to add it to your shell's startup file so you don't need to run it for every terminal session.
 4. Run your script
+</details>
 
-## Available Scripts
+## Available_Scripts
```
basiliskus, line 49: Was the underscore added by mistake?

scripts/README.md
```diff
@@ -1,8 +1,34 @@
-# Scripts
+- [Setup](#setup)
```
basiliskus, line 1: Can we update the TOC automatically when changes to the readme are made?

scripts/setup/setup-reportstream.sh
```diff
@@ -40,9 +40,15 @@ yq eval '.[0].receivers[] |= (
     }
 )' -i "settings/STLTs/Flexion/flexion.yml"
 
+LOCAL_DOCKER_IMAGE_NAME=$(docker ps --filter "name=trusted-intermediary-router-1" | grep trusted-intermediary-router-1)
 echo "Updating local URL and host in transport settings..."
-sed -i '' "s|__TI_API_URL__|${TI_LCL_API_URL}|g" "settings/STLTs/Flexion/flexion.yml"
-sed -i '' "s|__TI_API_HOST__|$(extract_host_from_url "${TI_LCL_API_URL}")|g" "settings/STLTs/Flexion/flexion.yml"
+if [[ ! -z $LOCAL_DOCKER_IMAGE_NAME ]]; then
```
basiliskus, line 45: Instead of adding an if statement here and adding a new env variable, I think we can update the existing one `TI_LCL_API_URL` to the docker local host name. You could add something like this to the top of the script
```
LOCAL_DOCKER_IMAGE_NAME=$(docker ps --filter "name=trusted-intermediary-router-1" | grep trusted-intermediary-router-1)
if [[ ! -z $LOCAL_DOCKER_IMAGE_NAME ]]; then
  export TI_LCL_API_URL="http://host.docker.internal:8080"
fi
```

scripts/README.md
saquino0827: I added it because the README was having issues finding the `Available Scripts` reference in the ToC. I'll play around with it more to see if I can get it working with a space

scripts/README.md
saquino0827: I've been using a VS Code plugin called `Markdown All In One` to generate it. I'll see if there are other options for generating as well.

scripts/README.md
saquino0827: That's fair. I could probably make the `scripts` section collapsable but I think with the new generated ToC, it might not be needed.

scripts/setup/setup-reportstream.sh
basiliskus: In that case, could we still set a variable at the top with the value of the TI endpoint depending if we're using docker or not? That would simplify and help avoid code duplication. Something like:
```
LOCAL_DOCKER_IMAGE_NAME=$(docker ps --filter "name=trusted-intermediary-router-1" | grep trusted-intermediary-router-1)
if [[ ! -z $LOCAL_DOCKER_IMAGE_NAME ]]; then
  ti_api_url=TI_LCL_API_URL
else
  ti_api_url=TI_DOCKER_LCL_API_URL_RS_CONFIG
fi
```

scripts/README.md
saquino0827: Intellij's Markdown plugin also has a generate Table of Contents feature. In Intellij, you open the markdown file -> command + n -> Generate Table of Contents

scripts/README.md
saquino0827: I was able to get the generate to work with spacing

scripts/setup/setup-reportstream.sh
```diff
@@ -40,9 +40,15 @@
     }
 )' -i "settings/STLTs/Flexion/flexion.yml"
 
+LOCAL_DOCKER_IMAGE_NAME=$(docker ps --filter "name=trusted-intermediary-router-1" | grep trusted-intermediary-router-1)
```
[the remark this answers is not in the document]
saquino0827: This is taken care of already

basiliskus:
@saquino0827 I just tried this PR and it's not working for me. The if statement is true, and it adds the docker host name, but that doesn't work in my case. I'm wondering if we should just add a note in the docs for people to update their hostname to whatever works in their environment instead of trying to set it automatically. We could still have two separate api url env variables, but instead of having one specific for docker, we could have one specific to the rs setup script

basiliskus:
I'd propose having this in the env.template file
```
TI_LCL_API_URL="http://localhost:8080"
TI_LCL_API_URL_RS_CONFIG="http://host.docker.internal:8080"
```
And directly using TI_LCL_API_URL_RS_CONFIG in scripts/setup/setup-reportstream.sh without the if statement

saquino0827:
> I just tried this PR and it's not working for me. The if statement is true, and it adds the docker host name, but that doesn't work in my case. I'm wondering if we should just add a note in the docs for people to update their hostname to whatever works in their environment instead of trying to set it automatically. We could still have two separate api url env variables, but instead of having one specific for docker, we could have one specific to the rs setup script

Yeah, I think I agree with this and I'll update the README.

basiliskus:
> I gave it a bit more thought with others and I might try one last attempt to getting the scripts working with Docker vs Gradle. We theorized that host.docker.internal settings works when both TI and RS are running in docker. I'll make some changes and do some testing

Ok. In case that doesn't work, I propose completely removing the if statement and just adding a note to the readme. Setting values to the person's specific environment in the .env file is the purpose of that file anyway
https://github.com/CDCgov/trusted-intermediary/pull/1648#issuecomment-2537215176

scripts/setup/setup-reportstream.sh
```diff
@@ -16,6 +16,16 @@ echo "Resetting the database and loading the baseline settings..."
 ./gradlew reloadTables
 ./gradlew reloadSettings
 
+# Update RS Configuration for the TI API based on docker or gradle
+# PS: Logic to determine to use gradle vs docker needs updating
+# The logic needs to check if BOTH TI and RS are running in docker, otherwise it should use localhost:8080
```
jcrichlake, line 21: Could we put an echo or two here notifying people of the behavior in case it's not known or that someone forgets?

saquino0827:
@basiliskus can you pull the latest changes and test to see if your environment gets set up correctly?
My machine is having issues running gradle in RS. I'm currently working to figure this out but I figure someone else with a gradle setup could test as well.

scripts/setup/setup-reportstream.sh
```diff
@@ -16,6 +16,18 @@ echo "Resetting the database and loading the baseline settings..."
 ./gradlew reloadTables
 ./gradlew reloadSettings
 
+# Update RS Configuration for the TI API based on docker or gradle
+# If either TI or RS are not running in docker, then use a gradle hosted URL
+LOCAL_DOCKER_IMAGE_NAME=$(docker ps --filter "name=trusted-intermediary-router-1" | grep trusted-intermediary-router-1)
```
basiliskus, line 21: Naming convention for local variables is to use lowercase. Could you also add `ti` to this variable name? So the variables names would be `local_ti_docker_image_name` and `local_rs_docker_image_name`

basiliskus (commented):
I tried it and it works for me! I just left one comment regarding the variable names

saquino0827:
> I'd propose having this in the env.template file
> 
> ```
> TI_LCL_API_URL="http://localhost:8080"
> TI_LCL_API_URL_RS_CONFIG="http://host.docker.internal:8080"
> ```
> 
> And directly using TI_LCL_API_URL_RS_CONFIG in scripts/setup/setup-reportstream.sh without the if statement

I went with your suggestion for naming

basiliskus (approved):

Issue #1424, opened 2024-10-14
Integration Tests for NTE Segment Mapping


tjohnson7021:
# DevEx/OpEx

Add an integration test in RS and RS-E2E to make sure the NTE segment is mapped correctly. The test has to be FHIR to HL7. 

## Tasks

- [x] ~Add a test for RS~ Integration test for OML already exists. 
- [x] Add a test for RS-E2E
  - [x] Add an assertion definition
    - [x] ~Possibly need to reconfig how assertion works to allow multiple segments~ We only need to test that the first segment exists. 

## Additional Context
- https://flexion.slack.com/archives/C055XTF22B0/p1710274890926099
- See bug report for more context on what to test: https://github.com/CDCgov/prime-reportstream/issues/13626

Add any other context or screenshots about the work here.

pluckyswan:
After some research with Gilmore and Sylvie, it was discovered that there is an existing integration test for OMLs within RS already. Therefore, only the TI RS-E2E was needed.

PR #1649, opened 2024-12-09
NTE Integration Test

pluckyswan:
# Description

Added a new assertion to test that NTE segment exists after transformation from FHIR to HL7. This was tested by triggering the Automated Staging integration tests. It successfully ran with the right conditions, and successfully broke when the conditions were purposely fiddled with.  

## Issue

https://github.com/CDCgov/trusted-intermediary/issues/1424

## Checklist


rs-e2e/src/main/resources/assertion_definitions.json
```diff
@@ -111,6 +111,16 @@
...
         "OBR-4.1 = '54089-8'"
       ]
+    },
+    {
+      "name": "NTE segment mapping",
+      "conditions": [
+        "MSH-10 = '011'"
+      ],
+      "rules": [
+        "NTE-3 = 'note 1'",
```
GilmoreA6, line 121: I understand why we are using this rule but I don't love it since it only works because our HL7 parser only returns the first segment of each type and there is no indexing.

GilmoreA6: Consider using NTE-3 = input.NTE-3 which would work even if we add indexing

GilmoreA6 (changes requested):

rs-e2e/src/main/resources/assertion_definitions.json
basiliskus: Great idea

pluckyswan: Good point. Changed it

pluckyswan: I reran the automation test as well and it works.

GilmoreA6 (approved):

Issue #1532, opened 2024-11-01
Look into completely removing the Hapi HL7 library dependency from the `rs-e2e` module


basiliskus:
# DevEx/OpEx

We've already written our own parser for HL7 and we've had some issues parsing with the Hapi HL7 library, which is not actively maintained, so it'd be better for maintainability to remove the dependency with that library 

## Tasks

- [x] Refactoring parser
- [x] Expanding the functionality of the parser to include what we're currently using the HAPI HL7 library for
- [x] Remove dependency to HAPI HL7 library

## Additional Context

https://flexion.slack.com/archives/C055XTF22B0/p1727717357268009

luis-pabon-tf:
On an initial check, I'm seeing that the implementation of the HL7 parser is still using classes from the `ca.uhn.hl7v2` library. This might cause issues removing the whole thing, but we might still be able to remove part of it.

basiliskus:
I can help with this one. I was planning to work on it anyway. The idea is to remove all the ca.uhn.hl7v2 dependencies and rely solely in the parser we have. We may need to add functionality the parser doesn't currently have to be able to completely remove the dependency

PR #1654, opened 2024-12-12
Refactor rs-e2e project to remove Hapi HL7 library

basiliskus:
# Description

- Removed the `ca.uhn.hapi` dependencies
- Added `HL7Message`, `HL7Segment`, `HL7Encoding` and `HL7Path` classes to replace Hapi's `Message` class
- Improved and refactored HL7 parser and expression evaluator
- Improved handling of edge cases
- Reorganized file structure
- Added full test coverage

## Issue

#1532 
## Checklist


jbiskie:
Tested and looks good.

Sonar flagged a circular dependency between HL7Message and HL7Parser, and there are actually a lot of them. Should we consider tagging this a a separate task to clean up?

basiliskus:
> Sonar flagged a circular dependency between HL7Message and HL7Parser, and there are actually a lot of them. Should we consider tagging this a a separate task to clean up?

I'll take a look. If it's straightforward to fix, we can include it in this PR

jorg3lopez:
Test task fails when I run `./gradlew :rs-e2e:test --info`

image

jorg3lopez:
> Test task fails when I run `./gradlew :rs-e2e:test --info`
> 
> image

I think we need to see the mocking for the rulesloader or rule. This is what is probably causing the null pointer:

image

image

basiliskus:
@jorg3lopez I'm not able to replicate your error. We didn't make any changes to the assertion engine or its test in this PR. It's probably related to your local configuration. Can you try running the other the rule engine tests (ValidationRuleEngineTest and TransformationRuleEngineTest) and see if you get the same error?

jorg3lopez:
> @jorg3lopez I'm not able to replicate your error. We didn't make any changes to the assertion engine or its test in this PR. It's probably related to your local configuration. Can you try running the other the rule engine tests (ValidationRuleEngineTest and TransformationRuleEngineTest) and see if you get the same error?

Per our conversation, this is a pre-existing bug that we will need to look into. This bug is present in the main repo, so ignore it for this PR.

rs-e2e/src/main/java/gov/hhs/cdc/trustedintermediary/rse2e/hl7/HL7Encoding.java
```diff
@@ -0,0 +1,78 @@
...
+
+import java.util.Map;
+
+/** The HL7Encoding class represents the encoding characters used in an HL7 message. */
+public record HL7Encoding(Map<String, Character> characters) {
+    public static final String DEFAULT_SEGMENT_DELIMITER = "\n";
+    public static final char DEFAULT_FIELD_DELIMITER = '|';
+    public static final char DEFAULT_COMPONENT_DELIMITER = '^';
+    public static final char DEFAULT_REPETITION_DELIMITER = '~';
+    public static final char DEFAULT_ESCAPE_CHARACTER = '\\';
```
GilmoreA6, line 11: Is there a reason the default escape character is not '\'. I'm assuming we need to escape the escape character for some reason?

basiliskus: Right, we need to escape the escape character. That's because `\` is also an escape character in Java

rs-e2e/src/main/java/gov/hhs/cdc/trustedintermediary/rse2e/hl7/HL7Path.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public static HL7Path parse(String path) {
+        Matcher matcher = HL7_FIELD_NAME_PATTERN.matcher(path);
+        if (!matcher.matches()) {
+            throw new HL7ParserException("Invalid HL7 path format: " + path);
+        }
+
+        String segmentName = matcher.group(1);
+        int[] indices =
+                Arrays.stream(matcher.group(2).split("\\.")).mapToInt(Integer::parseInt).toArray();
```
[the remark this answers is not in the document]
GilmoreA6: I assume we don't need to catch this since the HL7_FIELD_NAME_PATTERN regex will already throw an error for invalid path format?

jbiskie (approved):

rs-e2e/src/main/java/gov/hhs/cdc/trustedintermediary/rse2e/hl7/HL7Path.java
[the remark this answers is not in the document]
basiliskus: Right, the regex is already making sure the string are ints separated by dots. Otherwise it will throw an exception when matching

GilmoreA6 (approved):

rs-e2e/src/main/java/gov/hhs/cdc/trustedintermediary/rse2e/hl7/HL7Encoding.java
somesylvie: Will the `\n` also need a second `\` to escape its `\`? It would also be helpful to add a clarifying comment about the `\\`

basiliskus: The `\` in `\n` is the escape character itself so we don't need another one. This is all standard Java syntax so not sure if we need to add further clarification. If you write a Windows path string in Java you would write `C:\\Program Files\\Java`, and if you write a newline in a string you'd write `Hello\nWorld`

somesylvie: Ugh Java is so weird, thanks for the explanation

Issue #1122, opened 2024-05-30
Load Tests Against Azure Environment


JohnNKing:
# DevEx/OpEx

Currently, load tests are only run locally. Obtain more realistic data on TI service performance by running the tests within Azure.

## Tasks

- [ ] Something.

## Additional Context

Add any other context or screenshots about the work here.

jcrichlake:
In order to utilize our existing load tests we need to sign up for the azure locust preview test program.  This will take a few days to be completed. 

Adding the blocked tag to this until we hear back

brick-green:
@jcrichlake Who are we waiting to hear back from?

halprin:
we got past the current issue of the load tests.  The problem was we were missing `https://` (i.e. the protocol) for `LOCUST_HOST`.  The load test successfully hit Internal.

We can probably start adding back the other parts of the load test.


We also need to figure out how to authenticate agains the `token` endpoint.


There seems to be a section for secrets in the load test config, so perhaps we could use it to fill in a JWT or perhaps the private key that is used to sign JWTs.

Updated this ticket with additional tasks and context.

halprin:
I got the original load test file working in Azure (with some modifications) which includes authenticating against the `token` endpoint and hitting the other endpoints.  I got the secrets working with the JWT.

We've only ever done short load tests, so I will experiment with longer load tests and make sure the Azure load test framework continues to work.

halprin:
@somesylvie, is the task `Look into why failing to link metadata results in 5XXs.` complete since you fixed our database in Internal?  Your load test and my subsequent larger load test seems to have 0 errors.

halprin:
@somesylvie, I ran a larger test after your latest one (description of `Bigger test`).  I ran 2 engines, each with 50 users, for 10 minutes.  It passed with flying colors.

What are your thoughts on marking the task `Optional: look into why our deployed environments are slow for the orders and results endpoint (is it really due to RS staging being slow?).` complete?  Based on your last test and my larger test, the slowness _seems_ to be due to RS.  Thoughts?

somesylvie:
> @somesylvie, is the task `Look into why failing to link metadata results in 5XXs.` complete since you fixed our database in Internal? Your load test and my subsequent larger load test seems to have 0 errors.

Yep, I think this task is completed/resolved

somesylvie:
[...]

PR #1665, opened 2024-12-18
Azure load tests

somesylvie:
# Description

Adding support for load testing in Azure.

## Issue

#1122

## Checklist


etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
```diff
@@ -226,6 +227,12 @@ protected DomainResponse handleMessageRequest(
         boolean markMetadataAsFailed = false;
         String errorMessage = "";
 
+        if ("True".equals(request.getHeaders().get("load-test"))
```
basiliskus, line 230: Would it be better to do something like `Boolean.parseBoolean(request.getHeaders().get("load-test"))`? This handles casing and null safety

halprin: I like that.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
```diff
@@ -173,6 +174,12 @@ DomainResponse handleResults(DomainRequest request) {
     }
 
     DomainResponse handleMetadata(DomainRequest request) {
+        if (Boolean.parseBoolean(request.getHeaders().get("load-test"))
+                && ApplicationContext.isPropertyPresent("REPORT_STREAM_URL_PREFIX")) {
```
saquino0827, line 178: Is Application Context is set when we build the app?

operations/locustfile.py
```diff
@@ -131,22 +142,38 @@ def assert_stats(environment):
 def get_auth_request_body():
     # set up the sample request body for the auth endpoint
     # using a valid test token found in the mock_credentials directory
-    auth_scope = "report-stream"
-    with open("mock_credentials/report-stream-valid-token.jwt") as f:
-        auth_token = f.read()
+
+    # TODO - notes/clarification on 2 different creds, plus expiration date of jwt
+    # TODO - do we want to TF the tests? If yes which envs? In CDC envs, may need to adjust IP allow list on app. Also set as private endpoints in test config?
```
saquino0827, line 147: This is a good question. My assumption is that we don't want load test bombarding CDC environments unless they're okay with that. If they aren't, this could just be an internal running load test. However, I feel if we only run it in internal, it loses value since that would mean we're only running these until we pass the application off to the CDC.

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
halprin: I'm having a hard time understanding your question here.  Can you rephrase it?

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
```diff
@@ -226,6 +233,12 @@ protected DomainResponse handleMessageRequest(
         boolean markMetadataAsFailed = false;
         String errorMessage = "";
 
+        if (Boolean.parseBoolean(request.getHeaders().get("load-test"))
```
somesylvie, line 236: Since any new places that call RS will also need to copy this code, we should document that (probably both in a comment here and the other location we evaluate this, plus in docs)

etor/src/main/java/gov/hhs/cdc/trustedintermediary/etor/EtorDomainRegistration.java
saquino0827: No worries, I think I answered my own silly question. I forgot the that the whole point of env variables was to allow them to be set differently for each environment lol.

basiliskus (approved):

Issue #1600, opened 2024-11-20
Create E2E "Golden Copy" tests


GilmoreA6:
In order to holistically test complete mappings and transformations for test messages from our partners, we need to create the ability to have "golden copy" E2E tests where an input test message received from the lab/hospital is run through the entire message processing workflow and the output is compared to a known "good" copy. 

Requirements:
-Goes through entire flow and any relevant mappings/transforms
-Is part of RS E2E tests
-Tests every validation message from the UCSD implementation as defined in https://docs.google.com/spreadsheets/d/1AnPN2h8Gxf1234wbykmHTP4leqBOq2P4/edit?gid=340931274#gid=340931274

GilmoreA6:
@brick-green @scleary1cs @JohnNKing Ticket that will need refinement for introducing a different way of validating end2end tests for the CLIA validation scenarios

pluckyswan:
@dbgolson Can you verify if [the spreadsheet that is linked within the card](https://docs.google.com/spreadsheets/d/1AnPN2h8Gxf1234wbykmHTP4leqBOq2P4/edit?gid=340931274#gid=340931274) is the current copy? It is in the Archived folder. If so, where can I get the files mentioned within Original CDPH File Name row? Also, what is the Archived File Name step?

dbgolson:
> @dbgolson Can you verify if [the spreadsheet that is linked within the card](https://docs.google.com/spreadsheets/d/1AnPN2h8Gxf1234wbykmHTP4leqBOq2P4/edit?gid=340931274#gid=340931274) is the current copy? It is in the Archived folder. If so, where can I get the files mentioned within Original CDPH File Name row? Also, what is the Archived File Name step?

That spreadsheet is a "living document" as it is updated, I have archived the previous version with a date stamp prefix in the file name and placed it in a subdirectory called "Archived".  The most current working document titled "CA PHL NBS Test Results - Test Configurations Concept Codes and Test Case Scenarios.xlsm" is found here - https://drive.google.com/drive/folders/17_jQZj8KVrvtslnn_bAXmb6lEh-o0eNO?usp=drive_link

pluckyswan:
We currently have the modified workflow for automated staging test back up and running. We need to fix the encoding issue with our input example files, send them through the workflow, and store the expected files in our repo. We may also need to modify RS to have the blob use the external file name.

GilmoreA6:
[...]

PR #1674, opened 2024-12-23
Add Golden Copy Tests

pluckyswan:
# Description

Under construction...

## Issue

https://github.com/CDCgov/trusted-intermediary/issues/1600

## Checklist


rs-e2e/src/main/java/gov/hhs/cdc/trustedintermediary/rse2e/AzureBlobFileFetcher.java
```diff
@@ -47,12 +47,25 @@ public static FileFetcher getInstance() {
         return INSTANCE;
     }
 
+    //  TODO - we need to modify the guts to only grab the golden or automated path
     @Override
     public List<HL7FileStream> fetchFiles() {
+        String files_path = System.getenv("LOCAL_FILE_PATH");
```
basiliskus, line 53: Just a reminder to add instructions to the readme on how to set up the LOCAL_FILE_PATH env var. You could also consider adding it to `generate_env.sh`

basiliskus: I didn't realize this changes are in AzureBlobFileFetcher. This class is meant only for blobs in Azure. To grab local files, the correct one is LocalFileFetcher

rs-e2e/src/test/groovy/gov/hhs/cdc/trustedintermediary/rse2e/GoldenCopyTest.groovy
```diff
@@ -0,0 +1,80 @@
...
+        localFiles = localFileFetcher.fetchFiles()
+    }
+
+    def cleanup() {
+        for (HL7FileStream fileStream : localFiles + azureFiles) {
+            fileStream.inputStream().close()
+        }
+    }
+
+    def "Compare files"() {
```
basiliskus, line 59: Since this file has almost the same setup as AutomatedTest.groovy, could we merge them into one file and have one test for the rs integration test and another test for the golden copy?

.github/workflows/staging-test-run_reusable.yml
```diff
@@ -0,0 +1,49 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@v4
+
+      - name: Run automated tests
+        env:
+          AZURE_STORAGE_CONNECTION_STRING: ${{ secrets.AUTOMATED_TEST_AZURE_STORAGE_CONNECTION_STRING }}
+          LOCAL_FILE_PATH: ${{ inputs.FILE_PATH }}
```
saquino0827, line 38: Perhaps we can  rename this variable. Is this the local file path to our expected output? Maybe `LOCAL_EXPECTED_OUTPUT_FILE_PATH` or something similar/smaller if anyone can think of something better.

rs-e2e/src/main/java/gov/hhs/cdc/trustedintermediary/rse2e/external/azure/AzureBlobFileFetcher.java
```diff
@@ -49,12 +49,25 @@ public static FileFetcher getInstance() {
         return INSTANCE;
     }
 
+    //  TODO - we need to modify the guts to only grab the golden or automated path
     @Override
     public List<HL7FileStream> fetchFiles() {
+        String files_path = System.getenv("LOCAL_FILE_PATH");
```
saquino0827, line 55: I have comment above to address this naming to be more clear.

.github/workflows/staging-test-run_reusable.yml
basiliskus: Good point. It is actually input files. The name could be something like `RSE2E_LOCAL_INPUT_FILE_PATH` or something similar. Adding `RSE2E` prefix would be helpful to narrow the scope given the variable is instantiated in `generate_env.sh`, which is global to the repo

.github/workflows/golden-copy-staging-test-run.yml
```diff
@@ -0,0 +1,16 @@
+name: Golden Copy Staging Test - Run integration tests
+
+on:
+  pull_request:
```
pluckyswan, line 4: Once the PR is approved, I will remove `pull_request`. It is only included because a workflow has to be merged into main to be called/updated.

rs-e2e/src/main/java/gov/hhs/cdc/trustedintermediary/rse2e/external/azure/AzureBlobFileFetcher.java
```diff
@@ -51,10 +51,22 @@ public static FileFetcher getInstance() {
 
     @Override
     public List<HL7FileStream> fetchFiles() {
+        String rse2ELocalInputFilePath = System.getenv("RSE2E_LOCAL_INPUT_FILE_PATH");
```
basiliskus, line 54: I don't think the azure blob fetcher implementation should care about the file fetcher implementation details (like the value for RSE2E_LOCAL_INPUT_FILE_PATH). Also, figuring out if we're using the golden copy or RS integration test with the RSE2E_LOCAL_INPUT_FILE_PATH value won't work because we need to run both but have only one possible value for that envvar. Instead, you could pass a parameter fetchFiles (for both local and azure implementation), which will determine if we're fetching files/blobs for the golden copy test or the other one

rs-e2e/src/main/java/gov/hhs/cdc/trustedintermediary/rse2e/external/localfile/LocalFileFetcher.java
```diff
@@ -29,7 +27,12 @@ public static FileFetcher getInstance() {
 
     @Override
     public List<HL7FileStream> fetchFiles() {
-        try (Stream<Path> stream = Files.walk(Paths.get(FILES_PATH))) {
+        String rse2ELocalInputFilePath = System.getenv("RSE2E_LOCAL_INPUT_FILE_PATH");
+        if (rse2ELocalInputFilePath == null || rse2ELocalInputFilePath.isEmpty()) {
+            throw new IllegalArgumentException(
```
jcrichlake, line 32: You're throwing an unchecked exception here. Meaning if anyone calls this from somewhere else in the code and the exception happens it will cause the application to bomb out.  



Just checking that this is the intended behavior

rs-e2e/src/main/java/gov/hhs/cdc/trustedintermediary/rse2e/external/azure/AzureBlobFileFetcher.java
pluckyswan: Good point. Will implement

rs-e2e/src/main/java/gov/hhs/cdc/trustedintermediary/rse2e/external/localfile/LocalFileFetcher.java
pluckyswan: Per our conversation, I have taken your thoughts into consideration, and due to Basilio's above comment, have pivoted away from the issue. 

Thanks. 
P.S. xoxo

.github/workflows/golden-copy-staging-test-run.yml
pluckyswan: > Once the PR is approved, I will remove `pull_request`. It is only included because a workflow has to be merged into main to be called/updated.

This got buried in the commits but is important to note!

